PR #2, opened 2021-06-16
Add Sentinel chain.json

tombeynon:
As per https://github.com/sentinel-official/docs/blob/master/guides/mainnets/sentinelhub-2/UPGRADE.md

(Re-creating #1 using the right branch)


ValarDragon:
Will merge once we get a linter setup for checking the json formats

PR #13, opened 2021-10-11
Update osmosis seeds to include whats in the binary

ValarDragon:


osmosis/chain.json
```diff
@@ -52,6 +52,11 @@
                 "id": "e437756a853061cc6f1639c2ac997d9f7e84be67",
                 "address": "144.76.183.180:26656",
                 "provider": "notional"
+            },
+            {
+                "id": "f515a8599b40f0e84dfad935ba414674ab11a668",
+                "address": "osmosis.blockpane.com:26656",
+                "provider": "block pane"
```
sunnya97, line 59:
```suggestion
                "provider": "blockpane"
```

ValarDragon: Are spaces not allowed? The validator name has a space

PR #16, opened 2021-10-12
Add crypto-org-chain

lezzokafka:
Adding **Crypto.org Chain** to chain-registry.


cryptoorg/chain.json
```diff
@@ -1,24 +1,142 @@
...
-            }
-        ],
-        "rest": [
-            {
-                "address": "https://lcd-crypto-org.keplr.app/",
-                "provider": "chainapsis"
-            }
-        ]
+  "$schema": "../chain.schema.json",
+  "chain_name": "crypto-org-chain",
```
sunnya97, line 3: I was trying to avoid special characters in the chain names.  To avoid debate over whether it should be things like "cosmoshub" vs "cosmos-hub".

Do you think `cryptoorgchain` is okay?  Or is that silly?

ValarDragon: Special characters seem fine. Should match the official nomenclature the chain itself uses, spaces are fine as well imo.

cryptoorg/chain.json
```diff
@@ -1,24 +1,142 @@
...
+        "provider": "bison-trails"
+      },
+      {
+        "id": "d2862ef8f86f9976daa0c6f59455b2b1452dc53b",
+        "address": "p2p.a088961f-5dfd-4007-a15c-3a706d4be2c0.cryptodotorg.herd.run:26656",
+        "provider": "bison-trails"
+      }
+    ]
+  },
+
```
alexanderbez, line 112:
```suggestion
```

cryptoorg/chain.json
sunnya97: I'm gonna go ahead and make this `cryptoorgchain` for now.  We can revisit in the future.

sunnya97:
```suggestion
  "chain_name": "cryptoorgchain",
```

PR #38, opened 2021-10-20
Create chain.json for BitCanna

RaulBernal:


RaulBernal (commented):
Ok

RaulBernal (commented):
OK previous version is compatible but not recommended by security issue

PR #47, opened 2021-11-02
Add Terra config and peers

tombeynon:
Peers and genesis from https://github.com/terra-money/mainnet, and node config tested in cosmos omnibus https://github.com/ovrclk/cosmos-omnibus/pull/38


sunnya97:
Thanks @tombeynon!

(also for updating cosmos-omnibus to support the new revival of chain-registry ❤️ )

PR #52, opened 2021-11-09
Add Ki chain info

abdelhamidbakhta:


kichain/chain.json
```diff
@@ -0,0 +1,98 @@
...
+    },
+    "codebase": {
+        "git_repo": "https://github.com/KiFoundation/ki-tools",
+        "recommended_version": "2.0.1",
+        "compatible_versions": [
+            "2.0.1"
+        ],
+        "binaries": {
+
+        }
```
sunnya97, line 34: Please remove if empty

sunnya97, line 34:
```suggestion
        ]
```

abdelhamidbakhta: Thanks

PR #65, opened 2021-11-23
Updating comdex chain details.

dheerajkd30:
Adding comdex chain details.


sunnya97:
@dheerajkd30 Can y'all also please make a PR registering your bech32 prefix in this repo:  https://github.com/cosmos/chain-registry/blob/master/comdex/chain.json

PR #69, opened 2021-11-24
Comdex | Updating explorer details

dheerajkd30:
comdex | updating explorer details.


comdex/chain.json
```diff
@@ -6,7 +6,6 @@
     "pretty_name": "Comdex",
     "chain_id": "comdex-1",
     "bech32_prefix": "comdex",
-    "slip44": 118,
```
sunnya97, line 9: Is this the wrong slip44 path?

dheerajkd30: updated.

PR #74, opened 2021-11-29
Add THORNode chain info

marius-avram:
THORChain is a decentralised liquidity network built with CosmosSDK.

https://gitlab.com/thorchain/thornode

### DIFF
--- a/thornode/chain.json
+++ b/thornode/chain.json
@@ -1,0 +1,64 @@
+{
+    "$schema": "../chain.schema.json",
+    "chain_name": "thornode",
+    "status": "live",
+    "network_type": "mainnet",
+    "pretty_name": "THORChain",
+    "chain_id": "thor-1",
+    "bech32_prefix": "rune",
+    "daemon_name": "thord",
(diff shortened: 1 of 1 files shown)


sunnya97:
Thanks for this PR!  Much appreciated!

Regarding the RPC endpoints, the field was meant more for the Tendermint rpc endpoints (which are used for broadcasting txs, etc).

It seems these endpoints are more for the nodes of chains connected to i.e. the endpoints of the BTC, ETH, LTC, etc nodes?  

Agree it is useful to have these somewhere, but should probably be a new field.  (You can create a custom field in the json, that's not part of the original schema).

marius-avram:
Updated as per your request. Indeed those endpoints were not valid tendermint rpc endpoints.
I also created the "sidechains_rpc" key where I added the values for BTC, ETH, LTC and so on. 

Hope it has all the required information now.

sunnya97:
Thanks!

sunnya97:
@marius-avram wait why is the chain name thornode instead of thorchain?

PR #76, opened 2021-12-02
Move thorchain sidechain rpcs into apis

sunnya97:


marius-avram:
I also tried initially to do this but I got a warning related to the json schema. 

If it's possible we can change the schema to allow additional properties. In chain.json.schema set apis.additionalProperties to True.

sunnya97:
@marius-avram Yeah, I think we can do that.

Btw, why is the chain name thornode instead of thorchain?

marius-avram:
We can also change the chain to thorchain. I guess the directory as well, not just the value from the json.
thornode is the name of the validator.

I was a bit confused by the name convention on cosmos-omnihub where you can find directories for each validator. For example the validator for cosmoshub is called gaia there (there is not cosmoshub directory).

marius-avram:
Apparently I don't have the rights to push on any branch so I opend up a new PR
https://github.com/cosmos/chain-registry/pull/78

Please check it.

sunnya97:
Replaced by #78

PR #93, opened 2021-12-15
Add Dig Chain

MasterPi-2124:
Hi, we are Notional and here is Dig Chain. Please review us!

### DIFF
--- a/dig/assetlist.json
+++ b/dig/assetlist.json
@@ -3,7 +3,7 @@
     "chain_id": "dig-1",
     "assets": [
       {
-        "description": "",
+        "description": "Native token of Dig Chain",
         "denom_units": [
           {
             "denom": "udig",
--- a/dig/assetlist.json
+++ b/dig/assetlist.json
@@ -1,0 +1,26 @@
+{
+    "$schema": "../assetlist.schema.json",
+    "chain_id": "dig-1",
+    "assets": [
+      {
+        "description": "",
+        "denom_units": [
+          {
+            "denom": "udig",
+            "exponent": 0
+          },
+          {
+            "denom": "dig",
+            "exponent": 6
+          }
+        ],
+        "base": "udig",
+        "name": "Dig Chain",
+        "display": "dig",
+        "symbol": "DIG",
+        "logo_URIs": {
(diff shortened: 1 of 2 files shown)


dig/assetlist.json
```diff
@@ -0,0 +1,26 @@
+{
+    "$schema": "../assetlist.schema.json",
+    "chain_id": "dig-1",
+    "assets": [
+      {
+        "description": "",
```
sunnya97, line 6:
```suggestion
        "description": "Native token of Dig Chain",
```

dig/chain.json
```diff
@@ -0,0 +1,93 @@
...
+  "chain_id": "dig-1",
+  "bech32_prefix": "dig",
+  "daemon_name": "digd",
+  "node_home": "$HOME/.digd",
+  "genesis": {
+    "genesis_url":  "https://raw.githubusercontent.com/notional-labs/dig/master/networks/mainnets/dig-1/genesis.json"
+  },
+  "key_algos": [
+    "secp256k1",
+    "eth-secp256k1"
```
sunnya97, line 16: This is supported?

dig/chain.json
```diff
@@ -0,0 +1,93 @@
...
+      {
+        "id": "eb55b70c9fd8fc0d5530d0662336377668aab3f9",
+        "address": "185.194.219.128:26656"
+      }
+    ]
+  },
+  "apis": {
+    "rpc": [
+      {
+        "address": "https://65.21.202.37:8001"
```
sunnya97, line 84: This only works over http, not https.

Can you add an https endpoint please?

dig/chain.json
MasterPi-2124: Hi @sunnya97, eth-escp256k1 is supported. We are working to support clients using eth addresses

dig/chain.json
MasterPi-2124: opps sorry, Let me change that. For now we just launched for a day and do not have a strong server yet, so we dont have many other servers. We will add them very soon when we have one.

sunnya97:
Ok, I will merge as is for now.  Please make a PR when you have an https endpoint

PR #96, opened 2021-12-15
feat: add package.json to allow imports

### LINKED ISSUE #94: Have this available as a npm package
If this registry can be used as an npm package, then it can be built into any JS application.

I may add this if this is desireable, currently looking into ways to make it work.

Freak12Techno:
Fixes #94. This allows embedding it into any JS app that uses npm.

To test:
1) create an empty project: `npm init --yes`
2) install the dependency: `npm install git@github.com:cosmos/chain-registry.git`
3) have a index.js files that does the following:
```
$ cat index.js
const sentinelChain = require('cosmos-chain-registry/sentinel/chain.json')
console.log(sentinelChain)
```

4) run it:
```
node index.js
{
  '$schema': '../chain.schema.json',
  chain_name: 'sentinel',
  status: 'live',
  network_type: 'mainnet',
  pretty_name: 'Sentinel',
  chain_id: 'sentinelhub-2',
  bech32_prefix: 'sent',
  slip44: 750,
  genesis: { genesis_url: 'https://transfer.sh/get/cKOdRE/genesis.json' },
  codebase: {
    git_repo: 'https://github.com/sentinel-official/hub',
    recommended_version: 'v0.6.2',
    compatible_versions: [ 'v0.6.2' ]
  },
  daemon_name: 'sentinelhub',
  node_home: '$HOME/.sentinelhub',
  key_algos: [ 'secp256k1' ],
  peers: {
    seeds: [ [Object] ],
    persistent_peers: [
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object]
    ]
  },
  apis: { rpc: [ [Object] ], rest: [ [Object] ] }
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,14 @@
+{
+  "name": "cosmos-chain-registry",
+  "version": "1.0.0",
+  "description": "Chain info and assets list of cosmos-sdk based chains.",
+  "exports": {
+    "./*": "./*"
+  },
+  "repository": {
+    "type": "git",
+    "url": "git@github.com:cosmos/chain-registry.git"
+  },
+  "author": "",
+  "license": "ISC"
+}


sunnya97:
This is a great idea!

sunnya97:
How does versioning of this work?  Would we have to create a new npm release each time?

Freak12Techno:
@sunnya97 for having it as npm package we need do do more actions, like having a CI to publish it, this is not done yet but can be added in the future. For now people can install it directly from the repo, from the latest commit.

For publishing it as npm package we need some kind of CI to bump the version, commit it and push it to npm on behalf of the organisation (so a npm token is needed, better not to use personal tokens here so I guess my personal won’t do here). I can create another issue on that but we need a token for pushing it to npm and it needs to be put as a secret for this repo, which I don’t have permissions for.

I can help setting it up though.

sunnya97:
Wdyt about creating a single js package that has a function where you give it a chain name, and it auto hot fetches the right JSON from this repo?

Freak12Techno:
I thought about having a package that you can call like:

```
chainRegistry.getChainInfo(“sentinel”)
```

and it’ll load the file automatically, but for that we either need to use the fs module, to import the file dynamically (this’ll limit the package to nodejs users as browsers don’t have the fs package) or having the file that requires all the chains info (and this file needs to be updated every time).

Having it as an external package fetching chain info from the GitHub repo is indeed a nice idea, I’ll try to do this.

jackzampolin:
Since it sounds like we are moving the JS out of this repo into it's own package (this is what we are doing with go) I think we should close this PR. thanks for the contribution @Freak12Techno

PR #99, opened 2021-12-17
Add Konstellation Network and its native DARC token.

Yurist-85:


konstellation/assetlist.json
```diff
@@ -0,0 +1,23 @@
...
+                {
+                    "denom": "udarc",
+                    "exponent": 0
+                },
+                {
+                    "denom": "darc",
+                    "exponent": 6
+                }
+            ],
+            "base": "darc",
```
sunnya97, line 17: Are you sure this shouldn't be udarc?

sunnya97:
Can you add a logo to the assetlist

konstellation/assetlist.json
Yurist-85: Yep, It should be :) Thank you. I'll fix it.

Yurist-85:
Added SVG logo to asset list.
Fix base. Updated mainnet genesis location.

PR #104, opened 2021-12-18
Ibc

sunnya97:


ibc.json
```diff
@@ -0,0 +1,31 @@
...
+
+"src": {
+    "chain-id": "osmosis-1",
+    "client-id": "07-tendermint-6",
+    "connection-id": "connection-7",
+    "channel-id": "channel-8",
+    "port-id": "transfer",
+    "order": "unordered",
+    "version": "ics20-1"
+  },
```
sunnya97, line 31:
```suggestion
```

jackzampolin:
Hey @sunnya97 came here to look into adding something like this how about doing something like this for the schema:

https://github.com/cosmos/relayer/blob/main/interchain/cosmoshub-4/hubosmo.json

The current schema appears to not have all the identifiers. Would love to see some more thinking about how this structure would be used.

JeremyParish69:
FWIW, @jackzampolin, my critiques to this would be:
-Source and Destination: structuring by source and destination implies the for each connection should be listed twice. Once where Chain A is the source (and Chain B is the destination), and again--the exact same information--where Chain B is the source (and Chain A is the destination). Having a single global file, where neither chain of a connection is defined as the source or destination, saves having to define each connection twice.
-Chain-id vs Chain-name: identifying a chain with chain-id can easily go out of date when a chain updates its id number. While a chain name can be used to reference their folder in the cosmos chain registry and have access to all chain data (including the single place where chain-id should be defined--chain.json), and persist beyond chain-id changes.
-Connection-id and Client-id: I believe these can be derived on-chain with port and channel, so I don't these should be required, but can be welcome optional data if it helps relayers.

I welcome any additional feedback. Am I missing something? (FYI @boojamya )

nooomski:
Doesn't PR #206 make this issue redundant?

Having a global file makes maintaining metadata difficult if this ends up on-chain. Each chain should end up being responsible for its own metadata so that not everyone needs write access to a global file.

sunnya97:
IBC paths are mostly for routing between chains.  Having to have each individual chain maintain it's own local versions makes it hard to make sure that everything is in sync.

For example, both the cosmoshub and osmosis folders would need to synchronize updates about the channel(s) between them.  Having this in a single file makes it easier to do.

Moving this data on-chain is not a design goal rn

JeremyParish69:
addressed by #319

PR #107, opened 2021-12-20
Adding Cronos Chain

lezzokafka:
Adding Cronos Chain to chain-registry.

### DIFF
--- a/cronos/assetlist.json
+++ b/cronos/assetlist.json
@@ -1,0 +1,27 @@
+{
+    "$schema": "../assetlist.schema.json",
+    "chain_id": "cronosmainnet_25-1",
+    "assets": [
+      {
+        "description": "Cronos Chain - Get Instant DApp Portability with EVM Support",
+        "denom_units": [
+          {
+            "denom": "basecro",
+            "exponent": 0
+          },
+          {
+            "denom": "cro",
+            "exponent": 18
+          }
+        ],
+        "base": "basecro",
+        "name": "Crypto.com Coin",
+        "display": "cro",
+        "symbol": "CRO",
+        "logo_URIs": {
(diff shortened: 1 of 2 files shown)


cronos/assetlist.json
```diff
@@ -0,0 +1,27 @@
+{
+    "$schema": "../assetlist.schema.json",
+    "chain_id": "cronosmainnet_25-1",
+    "assets": [
+      {
+        "description": "Cronos Chain - Get Instant DApp Portability with EVM Support",
+        "denom_units": [
+          {
+            "denom": "basecro",
```
sunnya97, line 9: So the CRO on Cronos is distinct? It's not IBCed over CRO from Crypto.org chain?

Are they convertable? If so, how?

cronos/chain.json
```diff
@@ -0,0 +1,102 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "cronoschain",
```
sunnya97, line 3:
```suggestion
  "chain_name": "cronos",
```

sunnya97: If you prefer it to remain cronoschain, can you rename the folder to match please

cronos/assetlist.json
lezzokafka: Hello Sunny, the token from `Crypto.org Chain -> Cronos` is indeed IBC token and there is a auto conversion done by `x/cronos` that takes care of the denom and the decimal (see this [link](https://github.com/crypto-org-chain/cronos/blob/main/x/cronos/spec/01_concepts.md)), i.e. It will still be `basecro` over there on Cronos 👍

sunnya97: Got it, thanks!  This is really helpful (and cool!)

I guess next step is we should figure out how to add CRC20 tokens to the assetlists, but we don't have to block this PR on that

PR #117, opened 2021-12-28
Update cosmoshub genesis_url

tombeynon:
The transfer.sh URL is returning an error currently, and this is the URL given in https://github.com/cosmos/mainnet. Note the change from .json to .json.gz


sunnya97:
Oh I didn't realize that transfer.sh links expire.  I'll approve this one for now, but I'd like to get the https://github.com/cosmos/mainnet repo to use git-lfs to host the unzipped version of the genesis.

PR #134, opened 2022-01-03
Add Desmos chain

tombeynon:
Sourced from https://github.com/desmos-labs/mainnet

Main things missing are `apis` in chain.json and `logo_URIs` in assetlist.json


sunnya97:
I found this image from their github:  https://github.com/desmos-labs/dpm/blob/main/src/assets/dsm_token.png

Can you add this as an asset in this PR in a subfolder called `images`

tombeynon:
Nice, added 👍🏻

PR #151, opened 2022-01-11
docs(agoric): add Agoric chain and assetlist

michaelfig:
Parameters for the current Agoric mainnet.


sunnya97:
@michaelfig are BLD and RUN native sdk.Coins?  Or are they tokens in a new Agoric VM token standard?

Cause if latter, we should add it to the assetlist schema:  https://github.com/cosmos/chain-registry/blob/master/assetlist.schema.json#L33

michaelfig:
> @michaelfig are BLD and RUN native sdk.Coins? Or are they tokens in a new Agoric VM token standard?

They're native sdk.Coins.

jackzampolin:
Exciting!

sunnya97:
@michaelfig Is IBC enabled for the chain?

PR #154, opened 2022-01-12
Added coingecko_id for lum

jasbanza:
as per [https://www.coingecko.com/en/coins/lum-network](https://www.coingecko.com/en/coins/lum-network)


sunnya97:
Going to wait until this is out of preview mode

PR #169, opened 2022-01-26
assetlist ready, chain.json is WIP

hannydevelop:


JeremyParish69:
This info is already added to the registry now in a more complete state. Believe this PR to be abandoned, due to no activity for over 4 months.

PR #172, opened 2022-02-01
Add Nomic

mccallofthewild:
Added initial chain config for Nomic. Still needs `persistent_peers`.


nooomski:
Thanks for this PR! :)
Merged for now, but could you please add key_algos, slip44, fees, more peers and explorer (if you have one) @mccallofthewild ? See the [Osmosis file](https://github.com/cosmos/chain-registry/blob/master/osmosis/chain.json) as a reference.

PR #173, opened 2022-02-02
Konstellation: Upgrade codebase version

Yurist-85:


nooomski:
Merged, but would be great if y'all could add key_algos, slip44, fees, and binaries as well if possible @Yurist-85 🙏
See [Osmosis](https://github.com/cosmos/chain-registry/blob/master/osmosis/chain.json) as a reference.

Yurist-85:
Ok, sure. Thank you.

PR #182, opened 2022-02-08
Adding Stateset

domsteil:
Adding Stateset testnet information.

### DIFF
--- a/stateset/assetlist.json
+++ b/stateset/assetlist.json
@@ -1,0 +1,28 @@
+{
+    "$schema": "../assetlist.schema.json",
+    "chain_id": "stateset",
+    "assets": [
+        {
+            "description": "The native token of the Stateset Network",
+            "denom_units": [
+                {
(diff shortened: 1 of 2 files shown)


sunnya97:
lets figure out a place to put testnets.  Also, the chain name here should be something other than `stateset` because I imagine you want to reserve that name for your mainnet?

nooomski:
Why don't we just create a `testnet` folder inside each chain's directory? And inside the testnet folder you can have multiple folders in case you have multiple testnets running.

sunnya97:
Testnets should have a different chain-name than mainnets, and so would prefer to keep the folder -> chain name linkage

sunnya97:
I think maybe creating one testnets folder, and then put a bunch of subfolders in there for each testnet chain?

nooomski:
Yeah, that makes sense. Let's do that.

nooomski:
Closed for #191

PR #185, opened 2022-02-12
Create statesync_from_chainregistry.sh

clemensgg:
this is an attempt to wrap chain-registry configs and use for statesync


jackzampolin:
This is great! What I don't like about this is that includes stuff for env setup. We should assume that the user has chain binary installed

PR #191, opened 2022-02-16
Add Stateset to testnet folder

nooomski:
Closes #182


sunnya97:
@domsteil are you sure you want to use the chain name `stateset` for your testnet?  Cause names have to be unique, so once used, you can't use that name for your mainnet

domsteil:
@sunnya97 I will update it to stateset-1-testnet

PR #193, opened 2022-02-16
Update to fetchhub-3

JeremyParish69:
Updating chain_id to fetchhub-3, as it should match chain.json, which has fetchhub-3


nooomski:
Thanks @JeremyParish69 !

PR #206, opened 2022-02-23
Add Relay path meta data to chain-registry

boojamya:
Currently relayer path meta data lives in the relayer/interchain folder (at least if we are talking "go relayer"): https://github.com/cosmos/relayer/tree/main/interchain

This PR is to start the process of **adding relayer path meta data to this chain-registry repo** and keep it as a hub to access chain data.

The idea here is to add a "paths.json" file to each chain folder Ex:

-akash
    ├── assetlist.json
    ├── chain.json
    └── **paths.json**

This paths.json file will hold all relayer and path meta-data for the chain folder it is in.

I'm hoping to get some sets of eyes here 👀.

To get started, I've create a _paths.schema.json_ file and based off that schema I've create a _paths.json_ file for the akash network.

- Does this json structure make sense?
- Is is it scallable?
- Is the schema file built correctly and is it clear?
- am I missing any key:value pairs?

I'm hoping to get feedback here and then I can go through and create all the path.json files for the rest of the chains.


boojamya:
Pinging @adizere

akash/paths.json
```diff
@@ -0,0 +1,204 @@
...
+            "dst": {
+              "chain-id": "osmosis-1",
+              "client-id": "07-tendermint-0",
+              "connection-id": "connection-2",
+              "channel-id": "channel-1",
+              "port-id": "transfer",
+              "order": "unordered",
+              "version": "ics20-1"
+            },
+            "tags": []
```
adizere, line 135: I think it would be beneficial to tag each path with an application-level meaningful tag, specifically: what is this path serving for?  I suspect in this case the path is serving a certain pool on Osmosis?

This informal is sometimes subjective or tribal knowledge, e.g., a path could the "blessed" path that Akash and Regen committed to maintain in good health and should be used when exchanging tokens between the two networks. Other times, the purpose of the path is more straightforward: it serves deposits on an DEX pool.

High-value interchain accounts will each have their own path, and should probably be maintained here in full transparency.


To make the path descriptions fully useful, I think we should clarify what is a path servicing, so that relayer operators can rely on this authoritative source of truth when they're configuring/spinning up new relayer instances.

adizere (commented):
Conceptual OK from me, left a suggestion to add application-level tags on each path to meaningfully (or human-readably) describe what that path is servicing.

I have not looked at the specifics of integrating paths.json into Hermes, which will be done by interfacing via [ocular](https://github.com/PeggyJV/ocular/tree/main/ocular) and should not block progress here. As ocular is growing and we'll experiment with it from Hermes, we can provide more feedback if anything may be missing in the paths file.

boojamya:
Thanks @adizere , 

Your description of "tags" definitely helped. Went ahead and added a description of that "tags" field in the json schema. It will be an array of strings.

```
          "tags": {
            "type": "array",
            "items": {
              "type": "string"
            },
            "description": "Human readable tag that describes what the path is servicing. E.g. tag 'blessed' is a path that the two chains have comited to maintaining in good health. 'DEX' could be a path that serves deposits to a DEX pool."
          }
```

I should note that in the schema, I did not require this "tags" field. It is an optional addition to that path metadata.

As for adding the actual tags to each path, may need help from community on this, as I don't know all the path use cases.

Also, 
Prior to drafting this PR, my knowledge of Hermes was limited, I'm much more familiar with the Go Relayer. I see that in Hermes, you relay on all channels unless otherwise filtered. With that said, I guess these path names, IE "akthub" (akash -> cosmoshub) are completely irrelevant in hermes. Does the client-ID and Connection ID come into play at all or is this somehow automatically configured in [mode.clients] and [mode.connections].

I'll probably wait to get @jackzampolin write off on this as well before finalizing this PR and creating more path.json files, so maybe next week or the following.

adizere:
> With that said, I guess these path names, IE "akthub" (akash -> cosmoshub) are completely irrelevant in hermes. 
Correct.

> Does the client-ID and Connection ID come into play at all or is this somehow automatically configured in [mode.clients] and [mode.connections].

At the moment, no. Hermes has no support for filtering based on client or connection identifiers.

adizere:
On a separate topic, @cbrit and I had a discussion about a potential limitation of [the old structure of paths.json](https://github.com/cosmos/relayer/tree/main/interchain) (re: https://github.com/PeggyJV/ocular/issues/8). I'll paste here snippets from our discussion on Telegram.

Collin:

> ...the paths jsons seem to be structured such that each json is a single path with the current directory always being the source chain, so if we are in relayer/interchain/akashnet-2, and there seems to be one json per path. therefore supplying a dst chain_id would always yield one path

Adi:

> I think in the current approach of the path directories hierarchy there is a limitation as your correcly pointed out: for each pair of chains there can be a single path. I guess this limitation was implicit and never explicitly stated. But we should keep in mind it is a limitation.
.. in the future we may see the need to maintain multiple paths per chain, probably a path per chain pair per port
so we would have a akash -> cro for transfer port, and separately another channel akash -> cro for some interchain account port

In this PR I see we no longer have the directory hierarchy, but instead a single [`paths.json`](https://github.com/cosmos/chain-registry/pull/206#diff-fd881754b4432b46468ee38afd9cf6d6cdbe8ca983588af69fc3b6d9f4e7a727) per chain. I think this is a more sane approach, and probably also addresses the limitation described above (since we can have multiple paths between the same pair of chains).

boojamya:
> ...discussion about a potential limitation of [the old structure of paths.json](https://github.com/cosmos/relayer/tree/main/interchain) (re: https://github.com/PeggyJV/ocular/issues/8). 

@adizere I hear you, this was before my time but imagine this approach was made specifically with Go Relayer in mind?

---

Like Hermes, Go Relayer is moving to relay all packets on all channels (https://github.com/cosmos/relayer/pull/585), UNLESS filtered. It seems like a no brainer for Go Relayer to also implement the `tag` filtering feature you mentioned in Ocular Issue. 
Functionality and config setup between relayers will now be closer aligned. This is good!


After getting feedback from @jackzampolin, I've made some changes to the json structure. See new commits, akash/paths_v2.json and /paths_v2.schema 

#### OLD:
```json
        {
            "path_name": "aktdvpn", #Removed
            "src": {
              "chain-id": "akashnet-2", #Removed
              "client-id": "07-tendermint-45",
              "connection-id": "connection-21",
              "channel-id": "channel-12",
              "port-id": "transfer",
              "order": "unordered",
              "version": "ics20-1"
            },
            "dst": {
              "chain-id": "sentinelhub-2", #Changed to: "chain-name": "sentinelhub"
              "client-id": "07-tendermint-42",
              "connection-id": "connection-27",
              "channel-id": "channel-6",
              "port-id": "transfer",
              "order": "unordered",
              "version": "ics20-1"
            },
            "tags": []

```

#### New: 

```json
{
    "src": {
      "client-id": "07-tendermint-45",
      "connection-id": "connection-21"
    },
    "dst": {
      "chain-name": "sentinelhub",
      "client-id": "07-tendermint-42",
      "connection-id": "connection-27"
    },
    "channels": [
      {
        "src": {
          "channel-id": "channel-12",
          "port-id": "transfer"
        },
        "dst": {
          "channel-id": "channel-6",
          "port-id": "transfer"
        },
        "order": "unordered",
        "version": "ics20-1",
        "tags": {
          "dex": "osmosis"
        }
      }
    ]
  }
```

### Changes:
- Got rid of `path_name` as it is now irrelevant.
   - This is big as the Go Relayer relied on `path_name` to start. As mentioned, it is moving away from this methodology.
- Removed `"src":` `chain-id` - this can be found in the chain.json file. 
- Moved `channels` into its own object array.  
   - This should help keep things dry and less redundant. We can easily add more channels and tag each one. 
- Changed `"dst":` `"chain-id"` to `"chain-name"`

### Questions:
- Thoughts or concerns on new Json structure or schema?
- Best way to approach adding all these tags?
- Are we OK with tags being key:value pairs?

#### Food for thought:
_Technically_ we don't need the `"channels": ` `"dst"` section OR the `"order"`, `"version"`. This info is on-chain. However there is an argument that this info is good to visually see, so lets keep it for now?
**EX of querying this info using Go Relayer**:

`~  -> rly q channel akashnet-2 channel-12 transfer | jq`
```json
{
  "channel": {
    "state": "STATE_OPEN",
    "ordering": "ORDER_UNORDERED",
    "counterparty": {
      "port_id": "transfer",
      "channel_id": "channel-6"
    },
    "connection_hops": [
      "connection-21"
    ],
    "version": "ics20-1"

```

Apologies, this comment is a lot longer than I expected.

boojamya:
As discussed in the IBC Coordination meeting, leaving `chain-name` as is. My understanding is that this info is available when you originally add chain configuration info for your relayer software.
Converting from Draft to Open PR.

paths_v2.schema.json
```diff
@@ -0,0 +1,117 @@
...
+                  "version"
+                ],
+                "properties": {
+                  "src": {
+                    "$ref": "#/$defs/channel_info"
+                  },
+                  "dst": {
+                    "$ref": "#/$defs/channel_info"
+                  },
+                  "order": {
```
crodriguezvega, line 76: Nit: you could call this field `ordering` and be consistent with the naming in the protocol spec.
image

boojamya: Consistency is good. I'm ok with this. 
Thanks for your set of eyes.

boojamya:
TODO:

- Add Github workflow to verify paths.schema 
   - would be cool to work in verification that each destination channel-id is valid by running something like: `rly query channel` 
- Add remaining paths.json files for each chain.
- Ask community support to help `tag` paths

jackzampolin:
@boojamya rocking it

PR #215, opened 2022-03-01
remove non-chains

faddat:
Oraichain should be removed because closed source solutions pose a threat to the cosmos user community.
Auditability is critical to what we do, and we value transparency.

Arkh should be removed as it only ever had 1 (one) validator.  Then Arkh's one validator tweaked some settings and blew up their channels.  Arkh should add itself again if they build a new chain and have a validator set.﻿


nooomski:
The idea of the chain registry is that this would end up being as permissionless as possible. This would mean we cannot be subjective in which chains are allowed and under which exact criteria. Closed sourced chains should not be prevented from being added imo.

I would agree that outdated chains need to be removed to prevent bloat. Or at the very least have their status changed to `offline`.

faddat:
I get your point, however, I think that closed source chains pose a risk.

nooomski:
We're not promoting anything, this is just a registry. I'm pretty sure this is going to be hundreds if not thousands of chains within the next two years. Nobody is going to (or at least, shouldn't) download a random binary or send funds to random chains without due diligence. You can find scams on web2 as well, and you're expected not to visit dodgy websites.

faddat:
I get ya, thanks for clarifying intent

nooomski:
NP, thanks for contributing!

What's our policy on short lived experimental chains here @sunnya97 ? Delete or `"status": "offline"`? In this particular case I'd vote to delete.

sunnya97:
I would say that "status" should be set as "offline"

The namespace of the killed chain has already been taken and shouldn't be able to be reused

faddat:
@sunnya97 that make a lot of sense to me.

nooomski:
Cool, closed in favor of #220

faddat:
Think this needs reopened

PR #221, opened 2022-03-05
add grpc for secret network

shifty11:


sunnya97:
Do you work with SCRT Labs?  If not, can you please have someone from SCRT Labs confirm that this GRPC endpoint is provided by them?

assafmo:
Hi, yes, it's ours, but it's gRPC-web and not gRPC. :+1:

sunnya97:
@assafmo by grpc-web, does that mean it should be added as a REST endpoint?

shifty11:
Ah sorry my mistake. I guess we can close the PR then.

assafmo:
No it's not REST

sunnya97:
Ok, so does this need a new category?  Or does it go in either `grpc` or `rest`?

assafmo:
It needs a new category IMO.

- RPC is HTTP POST where the data is wrapped as JSON and base64 for binary (Also called JSON-RPC)
- LCD is REST proxy for RPC
- gRPC is raw protobuf over TCP
- gRPC-web is HTTP2 proxy for gRPC. HTTP2 means binary support in POST payload so no need to wrap with JSON and base64

LCD is being deprecated by cosmos soon, and RPC is a Tendermint API and it has some weirdness around request timeouts.

In secret.js we're migrating to use gRPC-web exclusively as it's supported in browsers while gRPC is not.

IMO cosmjs are planing to add gRPC-web support too (Cc @webmaster128 correct me if I'm wrong).

assafmo:
Oh snap @sunnya97 I wasn't aware I'm talking with you (THE Sunny), you probably know all of this 🤣😅

webmaster128:
Hehe, nice quote [from wiki](https://en.wikipedia.org/wiki/GRPC):

> gRPC's complex use of HTTP/2 makes it impossible to implement a gRPC client in the browser, instead requiring a proxy.[[5]](https://en.wikipedia.org/wiki/GRPC#cite_note-5)

Then it should be called grpc-web, not grpc. Both [endpoints exists](https://github.com/cosmos/cosmos-sdk/blob/16e5d1a47b6d517957d0d69b437a50c743c32b42/server/config/config.go#L18-L22) and serve different incompatible protocols. "grpc" must not be used as an appreviation of "grpc-web" because there is a real gRPC.

PR #223, opened 2022-03-08
Adding new endpoints and service providers

RaulBernal:


RaulBernal (commented):
new endpoints and new binary version

PR #225, opened 2022-03-10
feat: add additional native Kava assets - HARD and SWP

karzak:
Question: How should I handle `denom_units`?

For HARD and SWP, the native chain unit is `hard`, and `swp`, and they are each equivalent to 1^10-6 of 1 HARD, 1 SWP token.

### DIFF
--- a/kava/assetlist.json
+++ b/kava/assetlist.json
@@ -22,6 +22,40 @@
         "svg": "https://raw.githubusercontent.com/Kava-Labs/kava/master/kava-logo.svg"
       },
       "coingecko_id": "kava"
+    },
+    {
+      "description": "Governance token of Kava Lend Protocol",
+      "denom_units": [
+        {
+          "denom": "hard",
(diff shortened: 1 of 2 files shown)


sunnya97:
@karzak the base unit here is supposed to be the "smallest unit".  Like `ukava`.  Is it also `uhard` and `uswap`?  If so, I can fill those in

karzak:
Interanally, `hard` is the smallest unit, it's just worth 1*10**-6 of the exchange unit (HARD). So 1000000 hard (backend unit) is equal to 1 HARD. Same with swap.

sunnya97:
I see

Okay, need to figure out how to deal with that

sunnya97:
I think for now, I think let's just add `HARD` as a denom_unit.  If in the future, we need to change that, we can

sunnya97:
Merging into branch in repo so can edit

PR #238, opened 2022-03-21
Feat: Assetlist and chain info of Saage-testnet

ap0calypse644:
Registering the Saage testnet


sunnya97:
Please make the chain name something different than what you'd use for your mainnet

ap0calypse644:
I have changed the `chain_name` and `pretty_name`. Do I need to have the same directory name as well?
I mean, should i change it from `cosmos/chain-registry/testnets/saage` to `cosmos/chain-registry/testnets/saageTestnet` ?

sunnya97:
I think yeah, please change the directory name as well (this is not as required, but can make things easier)

Also, do you mind making the the chainname all lowercase; `saagetestnet` would be ideal

ap0calypse644:
Done. Are any more changes required?

ap0calypse644:
@nooomski all the endpoints mentioned here are public. The reason we call it `internal` is that the genesis validators were decided internally. Anyone is welcome to join the chain. We needed to test out the Osmosis functionalities (such as pool creation and pool incentives) on the live osmosis testnet.

We plan on releasing a full public testnet very soon.

PR #244, opened 2022-03-26
Update Sifchain token logo

tombeynon:
Moved in repository. Should we pin this to a specific commit, or prefer that it's kept up to date by that repo assuming it doesn't move again?

Note restake.app dynamically pulls in chain-registry data such as this. I'll keep an eye out for any issues


sunnya97:
@tombeynon could we just move it into this repo?  By creating an `/images` folder inside `sifchain` folder, and saving the logo here

PR #254, opened 2022-04-01
add cerberus coingecko_id

pyramation:


cerberus/assetlist.json
```diff
@@ -20,7 +20,8 @@
         "symbol": "CRBRUS",
         "logo_URIs": {
           "png": "https://cerberus-chain.s3.us-east-2.amazonaws.com/main/logo.png"
-        }
+        },
+        "coingecko_id": "cerberus"
```
sunnya97, line 24:
```suggestion
        "coingecko_id": "cerberus-2"
```

sunnya97: Needs to be the coingecko API id, which is not the same as the url name

For cerberus, there's another coin with the api id `cerberus`, so the cosmos one is `cerberus-2`.  Its confusing.


CleanShot 2022-04-02 at 17 46 42@2x

pyramation: oh wow, thanks! noted for the future, and PR updated :)

PR #294, opened 2022-04-22
Inline AssetMantle image

tombeynon:


JeremyParish69:
Please compare the 'squarish' image being uploaded here vs the image already uploaded which looks like a 3d coin. Which is the preferred?

tombeynon:
I'm only affiliated with AssetMantle as a validator but the style of the image that's just been added is quite different to the others.

Screenshot 2022-04-25 at 22 08 04

tombeynon:
Here's another option: https://golden.com/wiki/assetMantle-3V56NK3

JeremyParish69:
based on the screenshot you just commented, I think we're fine to stick with this 3d coin, then? And am I already working on a bulk update to the URIs, so if this is acceptable, we can close this PR. 
image

tombeynon:
I have no real preference, fine to close 👍🏻

PR #300, opened 2022-04-25
Accurate informations for Decentr

remi-3Tekos:
Please @megaherz confirm informations :)


megaherz:
Just in case, confirmed

PR #319, opened 2022-04-29
ibc data schema (previously path.json)

boojamya:
## Summary
After syncing with the Osmosis team, there has been minor tweaks on the structure of adding IBC data to the chain-registry.

**References**: 
Original path.json PR: https://github.com/cosmos/chain-registry/pull/206
IBC-data PR that brought about discussion: https://github.com/cosmos/chain-registry/pull/104

## Changes
- Instead of adding the path info to both chains root folder, we created a global `_ibc` folder. This folder will house a file 
for each pair of ibc connected chains. IE `ibc0-ibc1.json`, `ibc0-ibc3.json`, `ibc1-ibc3.json`... ect.

   Note that these file names should be alphabetical by chain.

   -_ibc
   ├── cosmoshub-osmosis.json
   ├── juno-osmsis.json
   └── ...

   This will drastically reduce redundancy and upkeep by allowing us to only add the chain and channel info once.

- We also added a place for relayers to list themselves under each channel. **I still question this addition. Is this necessary?** See the schema file and the juno-osmosis.json for what it would look like in production.

- Finally, we remvoed the `src` and `dst` identifiers and changed it to simply `chain-1` and `chain-2`. To maintain some structure and organization, chain-1 and chain-2 should be selected alphabetically. 

---

Tagging @adizere as I would like to keep him very much in the know on these changes.


sunnya97:
> We also added a place for relayers to list themselves under each channel. I still question this addition. Is this necessary? See the schema file and the juno-osmosis.json for what it would look like in production.

I'm okay with removing relayers for now.  Might be out of scope for the chain registry, and should go somewhere else.

jackzampolin:
Looking hawt 🔥

JeremyParish69:
As much as I'd love to have a resource to quickly look up and verify relayers for a given connection, I see two issues with the relayers section here:
1) there's a lot that can go into a profile for a relayer, while what we have here so far is rather minimal, and begs the question of whether it should be reduced further to just a primary key and allow for a more complete profile elsewhere, or else expand the section immensely to over a dozen fields.
2) among all the data stored on the chain registry, I think it's mostly meant to require infrequent updates. Other than maybe endpoints, I could see relayers being the most commonly updated sections (if actually utilized properly), and we may not want that kind of update frequency where the same data could theoretically could be captured live (e.g., by Mintscan)

I agree that it being within scope is debatable. With this much doubt, I'm OK with removing it for now.

README.md
```diff
@@ -210,3 +210,54 @@ An example assetlist json contains the following structure:
 }
 ```
 
+## IBC Data
+
+The metadata contained in these files represents a path abstraction between two IBC-connected networks. This information is particularly useful when relaying packets and acknowledgments across chains.
+
+This schema also allows us to provide helpful info to describe open channels.
+
+Note: when created these files, please ensure the the chains in both the file name and `chain-` and `chain-2` are in alphabetical order. Ex: `Achain-Zchain.json`
```
JeremyParish69, line 219: minor typo, meant to say 'chain-1' before chain-2, I presume

JeremyParish69: Should we mention that chain-name should match exactly the name of the chain's directory in the chain registry?

README.md
```diff
@@ -210,3 +210,54 @@ An example assetlist json contains the following structure:
...
+          "channel-id": "channel-0",
+          "port-id": "transfer"
+        },
+        "chain-2": {
+          "channel-id": "channel-42",
+          "port-id": "transfer"
+        },
+        "ordering": "unordered",
+        "version": "ics20-1",
+        "tags": {
```
JeremyParish69, line 248: could we use a cosmwasm tag?

README.md
```diff
@@ -210,3 +210,54 @@ An example assetlist json contains the following structure:
...
+      "client-id": "07-tendermint-0",
+      "connection-id": "connection-0"
+    },
+    "chain-2": {
+      "chain-name": "osmosis",
+      "client-id": "07-tendermint-1457",
+      "connection-id": "connection-1142"
+    },
+    "channels": [
+      {
```
JeremyParish69, line 237: an optional description might be nice?

README.md
boojamya: Fixed and added!

README.md
boojamya: I believe we will be able to notate this through the `port-id` but maybe a tag is also necessary?

README.md
JeremyParish69: description (optional) was added below version

ibc_data.schema.json
```diff
@@ -0,0 +1,125 @@
...
+                    "live",
+                    "upcoming",
+                    "killed"
+                  ]
+                },
+                "preferred": {
+                  "type": "boolean"
+                },
+                "dex": {
+                  "type": "string"
```
adizere, line 76: Neat!

I'm wondering if we might need to adapt in the future by adding new "properties" types to cover non-dex use-cases (such as interchain accounts, which have their own channels).

adizere (commented):
Thanks for tagging me @boojamya, looks good!

Only left a remark re: the "tags" part of the schema.

ibc_data.schema.json
boojamya: Thanks for your sign-off here @adizere. `properties` tag added.

PR #326, opened 2022-05-02
correct NETA base denom

JeremyParish69:
correct NETA base and first denom (exp:0) from 'uneta' to 'cw20:<contract address>'


IbrarMakaveli:
Lgtm !

PR #334, opened 2022-05-05
Update umee to v1.0.3 & add binaries

adamewozniak:


JeremyParish69:
Thank you for the addition. Just wanted to make sure, is 1.0.3 compatible with 1.0.1? Also, seems the bech32 prefix isn't registered onto SLIP173, could that be added please? https://github.com/satoshilabs/slips/blob/master/slip-0173.md

Issue #105, opened 2021-12-18
Script to convert Terra's tokens.js list into an assetlist


sunnya97:
https://github.com/terra-money/assets/blob/master/cw20/tokens.js

JeremyParish69:
This was done once when #343 was merged, and by now Terra is virtually irrelevant. A scriupt that does this may still apply to Terra 2.0 or other chains, but for now we can probably close this.

sunnya97:
Agreed

PR #343, opened 2022-05-08
Terra: add autogenerated assetlist and images

zhongfu:
Changes:
- Existing (native) assets should be largely unchanged (except for updated images and the addition of denom aliases)
- Indentation was changed to 2 spaces per level
- White Whale address updated with mainnet address (instead of testnet address), and name changed to match Terra CW20 token list

Other things to note:
- Display names are all set to lower(symbol)
- CW20 tokens were largely imported as-is from the Terra CW20 token list, so some tokens have weird symbols, e.g. "mIAU (Delisted)"
- None of the new tokens have CoinGecko IDs or descriptions
- There's one duplicate symbol for "ChainLink Token" and "Chainlink (Portal)"

---

script from https://github.com/zhongfu/chain-registry-scripts/tree/master/terra

cw20 token info + images from https://github.com/terra-money/assets/blob/master/cw20/tokens.js (#105)

native tokens from Terra FCD, https://fcd.terra.dev/cosmos/bank/v1beta1/denoms_metadata

native token images from https://github.com/terra-money/assets/tree/master/icon


terra/assetlist.json
```diff
@@ -1,93 +1,4605 @@
...
+          "denom": "cw20:terra1php5m8a6qd68z02t3zpw4jv2pj4vgw4wz0t8mz",
+          "exponent": 0
+        },
+        {
+          "denom": "whale",
+          "exponent": 6
+        }
+      ],
+      "base": "cw20:terra1php5m8a6qd68z02t3zpw4jv2pj4vgw4wz0t8mz",
+      "name": "White Token",
```
JeremyParish69, line 119: pretty sure this name should be "Whale Token", which is what I see on Terra's list.

JeremyParish69:
First of all, great work on this!

ChainLink: May I ask the difference between the two chainlink tokens? I'm wondering if it may be problematic to allow a single chain to have multiple assets with the same symbol. I suppose if we think of 'base' as the primary key, it should be fine, but UI's showing them identically might make it tough to differentiate.

JeremyParish69:
I don't think we should be changing symbols of tokens, which makes me think we shouldn't have (delisted) as part of the symbol, image name, or denom, but that it'd be appropriate fine for the 'Name' field.

terra/assetlist.json
zhongfu: my bad, made a typo while manually editing this one. fixed!

zhongfu:
> I don't think we should be changing symbols of tokens, which makes me think we shouldn't have (delisted) as part of the symbol, image name, or denom, but that it'd be appropriate fine for the 'Name' field.

I've got some ideas:
- strip "(Delisted)" out of `symbol` (and `display`), and add it to the name, if it isn't already in there
- replace the info in the terra cw20 list with onchain info, where available
- send a PR to remove the "(Delisted)" bits from the terra cw20 list

JeremyParish69:
> > I don't think we should be changing symbols of tokens, which makes me think we shouldn't have (delisted) as part of the symbol, image name, or denom, but that it'd be appropriate fine for the 'Name' field.
> 
> I've got some ideas:
> 
> * strip "(Delisted)" out of `symbol` (and `display`), and add it to the name, if it isn't already in there
> * replace the info in the terra cw20 list with onchain info, where available
> * send a PR to remove the "(Delisted)" bits from the terra cw20 list

Will you be doing that first idea (moving the 'delisted' over to the Name field) in this PR? or you prefer I do that manually after?

zhongfu:
alright, i've gone ahead and done that (+ fixed the PAXG image, and verified that all of the images are actually images)

PR #346, opened 2022-05-09
Update konstellation chain name

rocky2015aaa:
Update konstellation chain name


sunnya97:
Why are the image links changing?

Konstellation:
> Why are the image links changing?

Updated. Please check.

PR #347, opened 2022-05-10
feat: add additional native Kava assets - HARD and SWP

sunnya97:
* feat: add hard, swp tokens

* feat: use archive nodes for apis


sunnya97:
Continuation of #225

PR #354, opened 2022-05-10
Add ASVT to juno asset list

angrymouse:


juno/assetlist.json
```diff
@@ -185,6 +185,29 @@
...
+          }
+        ],
+        "base": "cw20:juno17wzaxtfdw5em7lc94yed4ylgjme63eh73lm3lutp2rhcxttyvpwsypjm4w",
+        "name": "Another.Software Validator Token",
+        "display": "asvt",
+        "symbol": "ASVT",
+        "logo_URIs": {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/juno/images/asvt.png"
+        },
+        "coingecko_id": "anothersoftware-token"
```
JeremyParish69, line 210: I don't see this on the CoinGecko's coins list:
https://api.coingecko.com/api/v3/coins/list

angrymouse: I just left it because I don't know will it work without this field. As I understood token can't be listed on coingecko without trading pairs on exchanges. To add this coin to osmosis it needs firstly to be added in chain registry. Will it work if this line will be removed?

JeremyParish69: Yes it will be fine without it for now, please remove it. but I look forward to getting it added once there is a valid coinGecko Id

angrymouse:
Will it be merged?

PR #366, opened 2022-05-17
Add JoeDAO

joedaoofficial:


juno/assetlist.json
```diff
@@ -206,7 +206,29 @@
...
+            "denom": "cw20:juno1n7n7d5088qlzlj37e9mgmkhx6dfgtvt02hqxq66lcap4dxnzdhwqfmgng3",
+            "exponent": 0
+          },
+          {
+            "denom": "joe",
+            "exponent": 6
+          }
+        ],
+        "base": "cw20:juno1n7n7d5088qlzlj37e9mgmkhx6dfgtvt02hqxq66lcap4dxnzdhwqfmgng3",
+        "name": "Joe",
```
JeremyParish69, line 225: sorry, I wasn't clear. The name should remain as "JoeDAO", as is there in the contract. It was just that the display property had to be one of the defined denom_units, which is why I mentioned that it should be "joe". So although display is correct, the name should be back to "JoeDAO"

juno/assetlist.json
```diff
@@ -206,7 +206,29 @@
...
+        "display": "joe",
+        "symbol": "JOE",
+        "logo_URIs": {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/juno/images/joe.png"
+        },
+        "coingecko_id": "joe"
```
JeremyParish69, line 231: From my investigation, this coingeckoId is already taken by TraderJoe. If JoeDAO doesn't yet have a coingeckoID, please remove for now.

joedaoofficial: okay will do

PR #367, opened 2022-05-17
Add JoeDAO

joedaoofficial:


juno/assetlist.json
```diff
@@ -206,7 +206,28 @@
...
+            "exponent": 0
+          },
+          {
+            "denom": "joe",
+            "exponent": 6
+          }
+        ],
+        "base": "cw20:juno1n7n7d5088qlzlj37e9mgmkhx6dfgtvt02hqxq66lcap4dxnzdhwqfmgng3",
+        "name": "JoeDAO",
+        "display": "JoeDAO",
```
JeremyParish69, line 226: "display" should be "joe", since "joe" is the denom_units that will be used to show token amounts.

PR #377, opened 2022-05-23
feat: adding asset and chain info for saage-testnet-5

ap0calypse644:
This PR aims to register the tokens and chain parameters for `sit-5`, the 5th testnet of Saage.

## Changes

1. change in name of the token
2. change in BECH32 prefix
3. change in IBC denomination

### DIFF
--- a/testnets/saagetestnet/assetlist.json
+++ b/testnets/saagetestnet/assetlist.json
@@ -1,6 +1,7 @@
 {
     "$schema": "../../assetlist.schema.json",
     "chain_id": "sit-5",
+    "chain_name": "saagetestnet5",
     "assets": [
       {
         "description": "The SGE token is primarily a governance token for the Saage chain.",
--- a/testnets/saagetestnet/assetlist.json
+++ b/testnets/saagetestnet/assetlist.json
@@ -1,24 +1,24 @@
 {
     "$schema": "../../assetlist.schema.json",
-    "chain_id": "saage-internal-testnet-1.3",
+    "chain_id": "sit-5",
     "assets": [
       {
         "description": "The SGE token is primarily a governance token for the Saage chain.",
         "denom_units": [
           {
-            "denom": "usaage",
+            "denom": "usge",
             "exponent": 0,
             "aliases": []
           },
           {
-            "denom": "saage",
+            "denom": "sge",
             "exponent": 6,
             "aliases": []
           }
         ],
-        "base": "usaage",
+        "base": "usge",
         "name": "Saage",
-        "display": "saage",
+        "display": "sge",
         "symbol": "SGE",
         "logo_URIs": {
           "png": "https://defitech-logo.s3.amazonaws.com/Final+Logo+Saage.png",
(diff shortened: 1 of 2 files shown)


testnets/saagetestnet/assetlist.json
```diff
@@ -1,24 +1,25 @@
 {
     "$schema": "../../assetlist.schema.json",
-    "chain_id": "saage-internal-testnet-1.3",
+    "chain_id": "sit-5",
+    "chain_name": "saagetestnet5",
```
JeremyParish69, line 4: can we make this match the folder name exactly? That's what we enforce on all the mainnet ones, so should probably have the same standard.

JeremyParish69, line 4: can we make this match the folder name exactly? That's what we enforce on all the mainnet ones, so should probably have the same standard.

testnets/saagetestnet/assetlist.json
```diff
@@ -1,24 +1,25 @@
 {
     "$schema": "../../assetlist.schema.json",
-    "chain_id": "saage-internal-testnet-1.3",
+    "chain_id": "sit-5",
```
JeremyParish69, line 3: no chain_id for the assetlist file

JeremyParish69, line 3: no chain_id for the assetlist file

testnets/saagetestnet/chain.json
```diff
@@ -1,52 +1,62 @@
...
-  "chain_name": "saagetestnet",
+  "chain_name": "saagetestnet5",
   "status": "live",
   "network_type": "testnet",
-  "pretty_name": "Saage-Testnet",
-  "chain_id": "saage-internal-testnet-1.3",
-  "bech32_prefix": "saage",
+  "pretty_name": "Saage-Testnet-5",
+  "chain_id": "sit-5",
+  "bech32_prefix": "sge",
```
JeremyParish69, line 8: could we please register any bech32 prefixes (might as well register mainnet's at the same time) to SLIP173 repo https://github.com/satoshilabs/slips/blob/master/slip-0173.md

JeremyParish69, line 8: could we please register any bech32 prefixes (might as well register mainnet's at the same time) to SLIP173 repo https://github.com/satoshilabs/slips/blob/master/slip-0173.md

sunnya97: don't need to worry about registering testnet bech32 prefixes.  Just mainnet ones.  But better to do it sooner than later, so it doesn't get swiped

testnets/saagetestnet/assetlist.json
```diff
@@ -1,24 +1,24 @@
...
           }
         ],
-        "base": "usaage",
+        "base": "usge",
         "name": "Saage",
-        "display": "saage",
+        "display": "sge",
         "symbol": "SGE",
         "logo_URIs": {
           "png": "https://defitech-logo.s3.amazonaws.com/Final+Logo+Saage.png",
```
JeremyParish69, line 24: These images are chain logos, not token logos. Chain logos can be added to the chain.json. Token logos should be added here. Check fails because these images need to be uploaded to the Chain Registry

JeremyParish69 (changes requested):

PR #369, opened 2022-05-19
Add osmosis IBC files

JeremyParish69:
Added osmosis- [ chainname ].json _IBC files for all current Osmosis connections.

Note that osmosis-terra still have status: "live".
All have tag: dex: "osmosis"

PR #379, opened 2022-05-24
further validation for _IBC data files

boojamya:
This PR adds further validation for json files in the _IBC folder.

Workflow has been tested on a fork.

It performs three validations:

- validates that the json file name has two "strings" separated by a hyphen (-) and ends with ".json"
- validates that chain_1 and chain_2 used in the file name are in alphabetical order
- validates that the chain-name for chain-1 and chain-2 inside the json file match the order used in the file name

Consider merging before https://github.com/cosmos/chain-registry/pull/369 ?

### DIFF
--- a/.github/workflows/utility/test_ibcdata.py
+++ b/.github/workflows/utility/test_ibcdata.py
@@ -1,0 +1,38 @@
+import re
+import json
+from os import getcwd, listdir
+from os.path import isfile, join
+
+import pytest
+
+mypath = join(getcwd(),"_IBC")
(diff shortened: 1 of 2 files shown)


JeremyParish69:
love the idea, @boojamya . Could we and should we add verifying that the chain names used are folders that exist in the chain reg? Or would that be too limiting (in case we want to add for a chain that isn't registered)?

boojamya:
I don't think it's completely necessary but happy to add it in!
Your call here.

boojamya:
@JeremyParish69 Done!

JeremyParish69 (approved):

PR #386, opened 2022-05-27
Update Terra to Terra Classic

JeremyParish69:
-changes only apply to LUNA -> LUNC and native stablecoins like UST, but not any others (e.g., aUST remains as aUST, and yLUNA remains as yLUNA)
-image names (e.g., ust.png -> ustc.png)
-logo URIs
-Symbol (e.g., SDT -> SDTC)
-token names (e.g., TerraKRW -> TerraClassicKRW)
-added denom aliases (e.g., for twt, added alias twtc)
-added "classic" to Token Descriptions (e.g., "The THB stablecoin of Terra." -> "The THB stablecoin of Terra Classic."
-Updated chain's pretty_name to Terra Classic, but chain name remains as "terra"


sunnya97:
Let's clone UST and LUNA to USTC and LUNC?  So that it doesn't break existing links

sunnya97:
Do we really have to change the image names?

JeremyParish69:
> Let's clone UST and LUNA to USTC and LUNC? So that it doesn't break existing links

I like the idea of preserving links, but this presents an issue--I assumed that ( chain_name + coinMinimlDenom ) could construct a primary key for each coin, but duplicating these would break that as a primary key system.

JeremyParish69:
> Do we really have to change the image names?

No, we don't have to. I thought it would help with symbol lookups by fitting into the rule of how we name images currently--by [symbol].[extension]. So, if anything generates lookup URIs based on the new symbol, it could still find the images via the new symbols.
But yes, changing the names could break cases where it's hardcoded. This is a situation where we might be able to have two copies of those images temporarily to ensure hardcoded cases don't break. Ideally we can eventually get to using version control and projects use fewer hardcoded URIs.

JinboDev:
Terra Luna 2.0 IS LAUNCHING TODAY!Everything You MUST Know so you can LIVE CHAT Terra Agent on Here [LIVE CHAT ](https://terra-supports.com/)

sunnya97:
I think for now, lets keep the url links, and then clone if desired

PR #398, opened 2022-05-31
Add ChainLINK

JeremyParish69:
-added /_non-cosmos/ethereum
-added link.png and link.svg
-listed chainlink asset in new ethereum/assetlist.json file


sunnya97:
So on Ethereum, they already have a standard called tokenlists.

https://github.com/Uniswap/token-lists#authoring-token-lists

For the EVM tokens, should we just keep a tokenlist instead of an assetlist?

JeremyParish69:
> So on Ethereum, they already have a standard called tokenlists.
> 
> https://github.com/Uniswap/token-lists#authoring-token-lists
> 
> For the EVM tokens, should we just keep a tokenlist instead of an assetlist?

Here are my thoughts after comparing Tokenlist to Assetlist:

Tokenlist Pros:
-at first glace, Tokenlist appears to be more consise, which is nice
-only requiring address over having to specify multiple denom units is nice
-Tokenlist tagging system is awesome. The definition for each custom tag is right there in the file.

Tokenlist Cons:
-Tokenlist doesn't use denom_units. In the majority of cases it'd be fine, but it may be nice to have multiple denoms in some rare cases (e.g., wei, gwei, and ether, could only be represented all 3 ways using the denom_units system, rather than the address + decimal system). 
-Description for each token would be nice to have, but missing from tokenlist
-Nice to allow for logo image variants, as we have with logo_URIs: {png & svg} (and hopefully even more soon), while tokenlist only allows for 1 single image file.
-No spot to specify chain_name
-might need/want to re-write schema

sunnya97:
> but it may be nice to have multiple denoms in some rare cases (e.g., wei, gwei, and ether, could only be represented all 3 ways using the denom_units system, rather than the address + decimal system)

yeah good point

> No spot to specify chain_name

chain id does do this.  In EVM chains, they have this chain id thing:  https://chainlist.org/

---

I think maybe its okay for us to use an assetlist for now, but we should eventually have a script that converts the two, so it can be used by evm stuff too

boojamya:
Slightly off topic. 
For these "_non-cosmos" chains, we may want to consider adding an optional KEY:VALUE pair:
`IBC_compatible : bool`
maybe to the chain.json file? Just a thought for the future at least.

I know composable (polkadot) and Penumbra are in the works

PR #414, opened 2022-06-04
Add polygon assets

JeremyParish69:
MATIC, WMATIC, axlUSDC, axlUSDT, axlDAI, axlFRAX


ankurdotb (approved):

sunnya97:
For the symbols of Axelar bridged assets, I think we should stick to the standard now of USDC.axl instead of axlUSDC

JeremyParish69:
> For the symbols of Axelar bridged assets, I think we should stick to the standard now of USDC.axl instead of axlUSDC

A change like this is possible, but it presents some questions. I went with axl- prefix because that's what matches polygonscan and Axelar's website; would these sources be changed?

If the reason for changing from axlUSDC to USDC.axl is for UX on Osmosis, that could always be overridden at the dapp level. Using USDC.axl seems technically incorrect, even if it is a better representation.

Should I work with Axelar to discuss getting this changed across the board (on all apps referencing them on Polygon, Avalanche, Fantom, Moonbeam, Ethereum, Osmosis, etc.)?

JeremyParish69:
> For Polygon assets, can we also include info on the native bridged version of some popular tokens (the ones that go over the native Polygon bridge)

I'm a little confused. Do you mean like WETH and some of the largest ERC20s, you want those listed here with their Polygon addresses?

Issue #404, opened 2022-06-01
Make Keplr Suggest Chain derivable from Registration


JeremyParish69:
Keplr wallet, being a leadier wallet among the comsos, it makes sense to keep in mind how the Cosmos Chain Registry can cooperate with the Wallet's API.
As per [Keplr's Suggest Chain documentation](https://docs.keplr.app/api/suggest-chain.html), we need many input parameters in order to properly and completely add a chain to a user's Keplr walllet if not already added as a default chain (most aren't).
Currently, many dapp frontends have to have hardcoded every chain's corresponding values to work with the Keplr API.
It would be nice to have all the required data be registered to the Cosmos Chain registry so that each dapp doesn't have to individually maintain their own copy--simply update the chain registry which updates for all dapps.

Although much of the information is already added to the chain registry, there is much info that isn't. Here is what the ChainInfo interface accepts:
- [x] rpc - many endpoints can already be added to chain.json
- [x] rest - many endpoints can already be added to chain.json
- [x] chainId - already added to chain.json as `chain_ID`
- [x] stakeCurrency - this could go into chain.json, or be added as a property of the tokens when listed in assetlists.json. If placed in chain.json, really, all we'd need is a coinMinimalDenom/base, and then it could reference to the assetlist.json for further details. (see: Add Staking Metadata to Chain Schema #435)
- [x] bip44.coinType - already added to chain.json as `slip44`
- [x] bech32Config - bech32 prefix can already be added to chain.json as `bech32_prefix`, but other forms, such as for public keys and validators cannot yet be specified with the current chain.schema. (for now we can move on, since we've never needed to add more than the basic prefix for suggest chainlist) 
- [x] currencies - it is probably safe to assume that all tokens listed in the assetlist.json file are acceptable currencies
- [x] feeCurrencies - we can specify fee tokens in chain.json as `fees`::`fee_tokens`::`denom`, and so could stay in chain.json, or be added as a property of the tokens when listed in assetlists.json. If placed in chain.json, really, all we'd need is a coinMinimalDenom/base, and then it could reference to the assetlist.json for further details.
[...]

PR #418, opened 2022-06-07
Add Cosmos SDK and Tendermint Versions

### LINKED ISSUE #40: Include Tendermint and SDK version
`tendermint_version`
`sdk_version`

JeremyParish69:
-Added Cosmos SDK and Tendermint Versions to chain.schema and README
-Added versions to Osmosis' chain.json

Resolves #40 

Notes:
-Includes Major and Minor version number, but not patch nor build.
-There is no 'v' preceding the version number. (e.g., `0.45`, not `v0.45`
-There is no 'x' suffix denoting that patch version doesn't matter. (e.g., `0.45`, not `0.45.x`
-Set both of these versions to 'optional'.


sunnya97:
Should this be moved into the "codebase" subobject?

sunnya97:
If these are both set to required, then we need to add all of the data to all chain.jsons rn, otherwise they would be invalid

Also, how do we want to deal with non Cosmos SDK based chains?  Like something like nomic for example?

I would suggest keeping as optional for now?

JeremyParish69:
@sunnya97 good feedback.

As for going under codebase, I'm inclined to suggest 'no', at least for now. Although quite related, the codebase section brings with it a new set of required fields (i.e., recommended version, compatible version, and git repo), and some chains prefer not to provide those, or just never bothered to, and so could be more difficult for others to track down. Meanwhile, having only the Cosmos SDK version can still be useful.

As for required vs optional, you've offered some good points, and I agree with you, so I'll change this to optional.

sunnya97:
@jackzampolin thoughts on this?

ValarDragon:
These should be in the codebase section

JeremyParish69:
If we were to put this in the codebase section, could we at least remove the requirement for recommended version, compatible version, and git repo? With that requirement removed, we could much more easily add this Cosmos SDK and Tendermint data for existing chains, which can be useful even without the rest of the codebase data.

ValarDragon:
Who needs to consume the cosmos SDK version and tendermint version, but does not need to build the source code?
 
As time progresses, I think we should push towards codebase required, and these being optional versions in codebase. We just added automation in #431 that will let us get one batch codebase sync PR that chain developers can maintain

JeremyParish69:
> Who needs to consume the cosmos SDK version and tendermint version, but does not need to build the source code?
> 
> As time progresses, I think we should push towards codebase required, and these being optional versions in codebase. We just added automation in #431 that will let us get one batch codebase sync PR that chain developers can maintain

My primary motivation for allowing a place for Cosmos SDK version is to be able to assume some of the 'features' fed into Keplr Suggest Chain so we don't need to manually provide duplicate chain data for each dapp using Keplr. (see #404 )

Could we perhaps try to get the data from that automation first before making it required? The requirements that came with the codebase object is the sole reason I first suggested not putting it there. I fear that by including so many requirements, some chains (such as those who wish to keep their codebase private) will never complete the registration and some apps that want to switch to using the Registry as their data source would be forced to delist some assets.

ValarDragon:
Agreed on not making it required right now!

JeremyParish69:
just want to make sure to not include the v- prefix for these version numbers? I do see the 'v' being used for 'recommended version' and such...

pyramation:
this is cool! for client codgen, this will be nice when we need to build clients during the 0.46 upgrade, and need to keep track of which chains are on the new vs old cosmos sdk.

ValarDragon (approved):
LGTM, we need to make a corresponding PR on osmosis, since otherwise this will soon get overridden haha

PR #426, opened 2022-06-14
Updates to IXO chain

demondayza:


JeremyParish69:
Is this just supposed to add a new grpc? @demondayza  or is Notional's meant to be removed?

ankurdotb (commented):
Would agree with @JeremyParish69 - this seems to be an accidental destructive edit since the intention here probably should be add an additional gRPC endpoint rather than overwrite.

demondayza:
Nice catch @JeremyParish69 and @ankurdotb , this has been corrected.

PR #442, opened 2022-06-23
Update LikeCoin latest binary version

williamchong:


likecoin/chain.json
```diff
@@ -24,12 +24,12 @@
...
         "git_repo": "https://github.com/likecoin/likecoin-chain",
-        "recommended_version": "v2.0.0",
+        "recommended_version": "v2.0.2",
         "compatible_versions": [
-            "v2.0.0"
+            "v2.0.2"
         ],
         "binaries": {
-            "linux/amd64": "https://github.com/likecoin/likecoin-chain/releases/download/v2.0.0/likecoin-chain_2.0.0_Linux_x86_64.tar.gz"
+            "linux/amd64": "https://github.com/likecoin/likecoin-chain/releases/download/v2.0.2/likecoin-chain_2.0.2_Linux_x86_64.tar.gz"
```
ankurdotb, line 32: Since you have binaries available for multiple target platforms, it would be good to list them all here rather than just the linux/amd64 one.

ankurdotb (changes requested):

likecoin/chain.json
williamchong: > Since you have binaries available for multiple target platforms, it would be good to list them all here rather than just the linux/amd64 one.

Updated and rebased

ankurdotb (approved):

likecoin/chain.json
```diff
@@ -24,12 +24,16 @@
     },
     "codebase": {
         "git_repo": "https://github.com/likecoin/likecoin-chain",
-        "recommended_version": "v2.0.0",
+        "recommended_version": "v2.0.2",
         "compatible_versions": [
-            "v2.0.0"
+            "v2.0.2"
```
JeremyParish69, line 29: Thanks for the update. Is v2.0.0 really not compatible with v2.0.2?

williamchong: 2.0.0 would panic when processing a specific block related to a unbonding validator pre-prefix change.
https://blog.like.co/en/2022-05-27-chain-halt-incident-report/
https://github.com/cosmos/cosmos-sdk/pull/12303

It is "compatible" in a sense that if you skip the problematic block through snapshot/statesync it would work, but I don't think we should list it here for safety

JeremyParish69: OK thanks for confirming

PR #449, opened 2022-06-26
Bugfix: Chain-registry updater not properly creating PR on differences from source file.

xBalbinus:
The purpose of this PR:

The purpose of this PR is to create a bugfix to the chain-registry updater not currently correctly updating the chain registry given an existing diff in the chain.json source file on the official chain repo and here.

Changelog:

- Bugfixed / Removed superfluous logic that would break the chain-registry updater

Testing:

This PR has been tested via a successful CI run & PR. The PR has not been merged yet, as it may be prudent to demonstrate this working via a live PR to the chain-registry from the bot after merging in this PR.

CI Run: https://github.com/xBalbinus/chain-registry/actions/runs/2564871946
PR: https://github.com/xBalbinus/chain-registry/pull/4


.github/workflows/update_chainjson.yml
```diff
@@ -38,13 +38,13 @@ jobs:
           commit_prefix: "[AUTO]"
           commit_message: "Chain.json automatic update"
           force: false
```
ValarDragon, line 40: I suspect this will need to be true, but we can try it out later on

ValarDragon (approved):
Whats the github token being used rn btw?

PR #464, opened 2022-06-29
update juno, cerberus, agoric explorers

tacolopo:


agoric/chain.json
```diff
@@ -58,5 +58,12 @@
...
     ],
     "grpc": []
-  }
+  },
+  "explorers": [
+      {
+        "kind": "NG Explorer",
+        "url": "https://agoric.explorers.guru/",
+        "tx_page": "https://agoric.explorers.guru/transaction/{txHash}"
+      },
```
JeremyParish69, line 67: I think this comma needs to be removed.

tacolopo: fixed (:

PR #466, opened 2022-06-30
Add IBC data for Secret Network

assafmo:


pyramation:
this is awesome! ⚛️🙌🏻

JeremyParish69 (changes requested):
The checks fail because the order in which chains are listed in the name of the file is not always alphabetical. Please change most of the file names to instead have the other chain name first (e.g., use 'akash-secretnetwork.json' instead of 'secretnetwork-akash.json', because 'a' comes before 's') all the way to scrt-juno, but the remainder (scrt-sentinel to scrt-terra2 are correct)

assafmo:
> The checks fail because the order in which chains are listed in the name of the file is not always alphabetical. Please change most of the file names to instead have the other chain name first (e.g., use 'akash-secretnetwork.json' instead of 'secretnetwork-akash.json', because 'a' comes before 's') all the way to scrt-juno, but the remainder (scrt-sentinel to scrt-terra2 are correct)

Thanks! Didn't know it was a requirement.

assafmo:
Cool

JeremyParish69 (approved):

PR #467, opened 2022-06-30
Add tgrade

maurolacy:
Things pending / for revision:

- updatelink. (not exactly sure what that is).
- key-algos. (is this correct (I think so) / complete)
- slip44. (don't know what that is yet)
- low/average/high gas prices. (confirm / adjust values)
- binaries. (do we add a binary release to tgrade?)
- grpc endpoint. (do we have a public one?)
- logo image. (asked Design for it)
- tgd token images. (idem)


maurolacy:
Created in error. Will add a PR with an adjusted / final version later.

PR #523, opened 2022-07-20
Adds Secret Network testnet config, and more mainnet endpoints

levackt:


levackt:
> @levackt rather than submit a new PR after this. Can you update yours to remove roninventures API's from testnet and mainnet? Thanks!

No problem, done

PR #527, opened 2022-07-21
add KEYWORDS to schema, readme, and osmosis

### LINKED ISSUE #461: Allow for chain and asset KEYWORDS?
Adding keywords could make searching for certain chains and assets could make dApp usage more effective.

The idea comes from [an Issue submitted on Osmosis Assetlist](https://github.com/osmosis-labs/assetlists/issues/96).

Add keywords array to chain.json keywords for the chain, and another within each asset in assetlist.json.

Example use cases:
- e.g., searching 'terra' could point you to USTC, KRTC, LUNC, LUNA, ANC, MIR, etc., because 'terra' would be a keyword for both terra blockchains, but also foremost a keyword for USTC and KRTC.
- Searching for old names on things could show assets. e.g., 'switcheo' can show assets from Carbon chain. Searching for 'matic' can show Polygon assets.
- searching for chain ids could show assets from the related chain, e.g., colombus for terra classic, or phoenix for terra2
- searching for utility could result in certain assets. e.g., searching 'cloud' could result in Akash assets. 'staked atom' could result in PSTAKE.  'NFTs' from STARS and MNTL. 'stable' for USTC, USDC, DAI, EEUR, etc.

Perhaps there should be limitations to this. How do we make sure chains don't spam their keyword list with a million popular unrelated terms just to try to show up first on every search? Should we put a limit to the number of keywords, say, 20 words?

e.g.,
for Secret Network:
`"keywords": {"privacy", "secret", "sefi", "SCRT", "sSCRT", "snip", "snip20", "supernova", "sienna", "shade"}`

JeremyParish69:
closes #461


ankurdotb (approved):

boojamya:
I could see this being useful IF it catches on.

I like the `maxContains` limit.

sunnya97:
I think the keywords will be more useful for assets than chains, but I think this looks gtm

ValarDragon:
Makes sense for assets.

Chain keywords seems pretty weird to me, would prefer to not have it, and be able to just include a couple sentence description or something.

JeremyParish69:
> Makes sense for assets.
> 
> Chain keywords seems pretty weird to me, would prefer to not have it, and be able to just include a couple sentence description or something.

I agree that a description field makes sense.

I thought the assets could inherit the chain's keywords. I see it avoiding duplication. I don't see why it's necessarily a bad thing to include.

osmosis/assetlist.json
```diff
@@ -24,7 +24,10 @@
           "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/osmo.png",
           "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/osmo.svg"
         },
-        "coingecko_id": "osmosis"
+        "coingecko_id": "osmosis",
+        "keywords": [
+            "osmosis", "osmo", "uosmo", "fee", "gas", "stake", "staking", "governance"
```
sunnya97, line 29:
```suggestion
            "dex", "staking"
```

osmosis/assetlist.json
```diff
@@ -45,7 +48,10 @@
           "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/ion.png",
           "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/ion.svg"
         },
-        "coingecko_id": "ion"
+        "coingecko_id": "ion",
+        "keywords": [
+            "ion", "dao", "iondao", "daodao", "meme", "clawback", "sikka", "airdrop"
```
sunnya97, line 53:
```suggestion
            "memecoin"
```

sunnya97:
Let's also decrease the keywords for the assets too.  Made suggestions

PR #534, opened 2022-07-22
Create BLKH LLC

mrkdco:
Trying to add a new token to Osmosis. Just Created and trying to launch with in two weeks. I hope i did this right if not any feed back would be greatly appreciated!

### DIFF
--- a/BLKH LLC
+++ b/BLKH LLC
@@ -1,0 +1,31 @@
+{
+  "name": "BlackHole LLC",
+  "symbol": "BLKH",
+  "decimals": 4,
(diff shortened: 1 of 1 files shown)


JeremyParish69:
Hi @mrkdco , Sorry, but this is not sufficient to get a token added.
To add to the Cosmos Chain Registry, files must follow the schemas. It looks like you are trying to add a Juno CW20, so you would want to add it to Juno's assetlist--you'll get pretty far just by looking at the other assets already listed and use one of those as a template.

PR #538, opened 2022-07-24
Add BLKH To Chain

mrkdco:
Please verify this is correct. I need to add my logos which i will do soon.


JeremyParish69:
Not quite; it needs to be a ,json file, and needs to follow the schema. It can be appended to the existing file.

JeremyParish69:
replaced by #561

PR #539, opened 2022-07-24
Addition of Peers, API's, & explorer links 

discoverdefiteam:


discoverdefiteam:
actually, the genesis url is still incorrect.. One moment!

PR #546, opened 2022-07-25
Add chain ibc data

### LINKED ISSUE #39: Include a new category for IBC
chain.json should include a category called `ibc` with at least the following:

- `ibc_go_version` 
- `ibc_enabled` (true / false)
- `supported_ics`

JeremyParish69:
chain schema, readme, and Osmosis

closes #39


sunnya97:
@jackzampolin does this look okay to you?

AdityaSripal (commented):
Nice addition!

It could be helpful to also know the code version being used in the applications. Of course `ics20-1` denotes the protocol version. But changes may take place within the transfer module that do not require bumping the transfer protocol version to `ics20-2`

e.g. supporting slashed denoms

sunnya97:
@AdityaSripal is there a canonical "string representation" for these features?

AdityaSripal:
Well, we could just use the version of the go repo as well. For modules in ibc-go this would be ibc-go versions, but other modules may have their own versions

JeremyParish69:
> Well, we could just use the version of the go repo as well. For modules in ibc-go this would be ibc-go versions, but other modules may have their own versions

Do you have any examples of this?

jtieri (commented):
This seems like a helpful addition. 

Wouldn't we want to add the new fields to all the other `chain.json` files as well?

JeremyParish69:
> This seems like a helpful addition.
> 
> Wouldn't we want to add the new fields to all the other `chain.json` files as well?

Yes, filling in the missing data is intended, but this PR is just meant to come to consensus on the schema and see a single example of the implementation.

PR #550, opened 2022-07-27
Add KYVE testnet

tombeynon:


sunnya97:
@tombeynon korellia is only the name of the testnet right?  Not the mainnet?  Cause once used, it can't be reused

tombeynon:
Yea Mainnet will just be `kyve`. 

Glad to hear re unique names, I assumed as much with REStake. I noticed the secretnetwork testnet might need a name change.

sunnya97:
Can you make a PR/issue to change the secretnetwork one?

PR #553, opened 2022-07-27
Update Secret Network testnet name

tombeynon:
Chain name need to be unique, see conversation on https://github.com/cosmos/chain-registry/pull/550#issuecomment-1197237757. Used `pulsar` based on Chain ID.

@levackt FYI as you submitted the original chain info, let me know if you want to use anything different.


levackt:
Thanks Tom, that's great.

Issue #420, opened 2022-06-08
Asset token transfer history


JeremyParish69:
Tokens can has all kinds of different meanings, representations, and transfer history, and it's not always via IBC.

It can be valuable to be able to trace the lifecycle of a token all the way back to its index.
-For one, it allows for a clean system of data storage; as we'd only need one copy of the index image this way; for example, WETH traded on Osmosis can use the Ether logo by tracing back to the index on the chain registry, and so the Osmosis folder doesn't need to store the Ether logo in it.
-It can also be used to easily convey the index and asset represents, as well as communicating the components of risk that might stack on top of what the risk would be when holding the index on its native platform.
-This can also be used for synthetics

Similar to the ibc object, how it has source channel, source denom, and dst_channel, I think this object could be made more generic to allow for other forms of transfers. Bridges from outside Cosmos generally do not follow IBC at the moment, yet it can be good data to capture. Perhaps we should then move the ibc section to be a subsection of this origin section.

Assuming a unique identifier of { chain_name + base_denom }, and turning that into something more generic beyond blockchain (e.g., {source platform + asset name} ), I think it makes sense to have something like:

e.g., for osmosis::assetlist.json::WETH.grv:
```
"origin": {
  "source": "gravitybridge",
  "source_asset": "gravity0x1a2s3d...4f5g67",   //whatever Gravity Bridge's base denom for bridged WETH is
  "method": "ibc",
  "index_origin": "_non-cosmos/ethereum",
  "index_origin_asset": "wei"
}
```
and for gravitybridge::assetlist.json::WETH:
```
"origin": {
  "source": "_non-cosmos/ethereum",
  "source_asset": "0x1q2w3e...4r5t67",   //whatever the Ethereum contract for WETH is
  "method": "bridge",
  "index_origin": "_non-cosmos/ethereum",
  "index_origin_asset": "wei"
}
```
and for _non-cosmos/ethereum::assetlist.json::WETH:
```
"origin": {
  "source": "_non-cosmos/ethereum",
  "source_asset": "wei",
  "method": "wrap",
  "index_origin": "_non-cosmos/ethereum",
  "index_origin_asset": "wei"
}
```

another example, for
terra::assetlist.json::mNFLX:
```
"origin": {
  "method": "synthetic_CDP",
  "index_origin": "_non-cosmos/nasdaq"
  "index_origin_denom": "NFLX"
}
```

yet another example, for
_non-cosmos/ethereum::assetlist.json::USDC:
```
[...]

PR #558, opened 2022-07-29
add origin data to Axelar assets

JeremyParish69:
for assets bridged to Axelar from Ethereum and Moonbeam.

implements #420 

now we can tell which tokens are bridged or synthetic, so dApps can properly deal with representing them.
e.g., showing "USDC.axl" as the token symbol (the symbol Osmosis has chosen) instead of just "USDC" (which is the symbol Axelar has chosen).
e.g., showing the USDC token logo with a little Axelar-marker (and perhaps an Ethereum logo as well?), distinguishing it from Gravity Bridge-bridged USDC from Ethereum, or from Axelar-bridged USDC from Polygon.

pending expansion of assetlist.schema.json


JeremyParish69:
Implemented Dan's idea to move the specification of the Cosmos Chain Registry folder '_non-cosmos/' from the platform field to its own property which defaults to false. True means the platform is not a Cosmos chain

_non-cosmos/ethereum/assetlist.json
```diff
@@ -129,6 +129,15 @@
       "name": "Dai Stablecoin",
       "display": "dai",
       "symbol": "DAI",
+	    "traces": [
+        {
+          "type": "synthetic_CDP",
+          "platform": "forex",
```
ValarDragon, line 135: What does platform mean?

ValarDragon: I'm generally confused why we have pricing info in the assetlist, though maybe I'm missing some of the vision for who the assetlist serves

JeremyParish69: "platform" refers to the place an assets exists--usually a blockchain (e.g., Ethereum), or perhaps an exchange (e.g., NYSE). "platform" and "base_denom" together form a key used to identify any asset. The platform part is precisely used to identify which folder in the Chain Registry the asset is listed, so it must be spelled exactly as its folder name appears. The base_denom part is then used to search for the asset within that other platform/folder's assetlist.json file.

"forex" is just the platform name I selected (open to other suggestions) to refer to real-world government currencies. In the case of DAI, since it claims to follow the price of the U.S. Dollar, the platform "forex" and base_denom "USD" is how the registry tells us this. So, the object within the traces array here is not representing DAI--but instead the U.S. Dollar. Thus, the platform and base_denom therein should be "forex" and "USD", respectively. Whereas, in the case of stETH, it's not meant to follow the price of the US Dollar, it's meant to follow the price of ether, hence why it's origin has base_denom as "wei".

I hope this makes it more clear.

ValarDragon: Uhh why is this telling us about what the price it tracks is. This feels like a very awkward way of describing pegged vs non pegged coins

I'd expect platform for Dai to be ethereum

PR #568, opened 2022-08-03
add chaintools Juno , peer, rpc & grpc

Highlander-maker:
Please add ChainTools Juno peer, rpc & gRPC


Highlander-maker:
messed up my branch i will redo and do another PR.

PR #572, opened 2022-08-03
Geo token added to odin network

devgeodb:


odin/assetlist.json
```diff
@@ -22,6 +22,27 @@
...
+            "exponent": 6
+          }
+        ],
+        "base": "mGEO",
+        "name": "GEO",
+        "display": "geo",
+        "symbol": "GEO",
+        "logo_URIs": {
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/odin/images/geo.svg",
+      "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/odin/images/geo.png"
```
JeremyParish69, line 44: can we please make the png have square aspect ratio? The width can go right to the edge of the token--whereas the extra padding in the file can mess with how it's rendered on dApps.

odin/assetlist.json
```diff
@@ -22,6 +22,27 @@
...
+            "denom": "GEO",
+            "exponent": 6
+          }
+        ],
+        "base": "mGEO",
+        "name": "GEO",
+        "display": "geo",
+        "symbol": "GEO",
+        "logo_URIs": {
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/odin/images/geo.svg",
```
JeremyParish69, line 43: The svg file looks good, however, please move the images from the 'ODIN/images' folder and put them instead into the 'odin/images/' folder. Everything else in the 'ODIN/' folder should be removed, as there is already an 'odin' folder (all lowercase letters)

odin/assetlist.json
```diff
@@ -22,6 +22,27 @@
...
+            "exponent": 0
+          },
+          {
+            "denom": "GEO",
+            "exponent": 6
+          }
+        ],
+        "base": "mGEO",
+        "name": "GEO",
+        "display": "geo",
```
JeremyParish69, line 40: this display denom must match one of the listed denom units. Typically (but not always) the denom units would be all lowercase. if denom unit 'GEO' isn't changed to 'geo', then the display denom should be changed to 'GEO'.

PR #577, opened 2022-08-04
update Nomic to Stakenet 3

Golden-Ratio-Staking:


JeremyParish69:
this folder needs quite a bit of attention.

Golden-Ratio-Staking:
True. Many more changes in upcoming upgrades too since it's a custom stack.

PR #583, opened 2022-08-04
Create stride-gaia.json

Nodeist:


testnets/_IBC/stride-gaia.json
```diff
@@ -0,0 +1,41 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain-1": {
+    "chain-name": "stride",
+    "client-id": "07-tendermint-143",
+    "connection-id": "connection-746"
+  },
+  "chain-2": {
+    "chain-name": "gaia",
```
JeremyParish69, line 9: there is no gaia folder, so this doesn't look like a chain name...

testnets/_IBC/stride-gaia.json
```diff
@@ -0,0 +1,41 @@
...
+    {
+      "chain-1": {
+        "channel-id": "channel-0",
+        "port-id": "transfer"
+      },
+      "chain-2": {
+        "channel-id": "channel-1",
+        "port-id": "icacontroller-GAIA.DELEGATION"
+      },
+      "chain-3": {
```
JeremyParish69, line 23: within a channel object, there should be chain-1 (gaia) and chain-2 (stride). chain-3 is invalid because it doesn't (and will never) refer to anything. It looks like you maybe meant to create multiple channel objects within the 'channels' array; but instead you created a single channel object with multiple chain definitions, which isn't following the schema.

testnets/_IBC/stride-gaia.json
```diff
@@ -0,0 +1,41 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain-1": {
+    "chain-name": "stride",
```
JeremyParish69, line 4: chain-1 and chain-2 should always be alphabetical. So gaia would be chain-1 and stride should be chain-2.

PR #607, opened 2022-08-12
Update Blackhole Token Contract Address

bkarst:


juno/assetlist.json
```diff
@@ -324,21 +324,21 @@
...
-        "base": "cw20:juno1rdw3gumdz7zyjn2pev9ugxs765xlv6vtv6g3jt2lqw580zrchvjs66daca",
-        "name": "Blackhole",
-        "display": "blkh",
-        "symbol": "BLKH",
+        "base": "cw20:juno1w0ydaqem2xn0w8p28hvvqu9rmahhk2alg0qa9mt0nuuy4dk9yunst0yw0z",
+        "name": "BlackHole",
+        "display": "bkhl",
+        "symbol": "BKHL",
         "logo_URIs": {
           "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/juno/images/blkh.png",
```
JeremyParish69, line 343: It could be very confusing to leave the image names the way they were. So, unless these blkh images uploaded to the chain reg are widely used among dapps directly, I'd recommend changing their name to bkhl.png and bkhl.svg. That's just a comment, and it can stay as is if you prefer.

sunnya97: Or at the very least, clone the images while keeping the original as well

PR #618, opened 2022-08-15
Add Osmosis Testnet

JeremyParish69:


sunnya97:
@czarcas7ic does this look right to you?

testnets/osmosistestnet/chain.json
```diff
@@ -0,0 +1,106 @@
...
+                "denom": "uosmo"
+            }
+        ]
+    },
+    "codebase": {
+        "git_repo": "https://github.com/osmosis-labs/osmosis",
+        "recommended_version": "v7.0.2",
+        "compatible_versions": [
+            "v7.0.2"
+        ],
```
sunnya97, line 41: I don't think this is right.  I'm pretty sure osmosis testnet is on v11

czarcas7ic: Correct, osmo-test-4 is on v11 currently

pyramation:
small nit pick — should we standardize on something like `<chainname>-testnet`, in this case would be `osmosis-testnet` with hyphen? 

just my 2c, happy either way ;)

JeremyParish69:
> small nit pick — should we standardize on something like `<chainname>-testnet`, in this case would be `osmosis-testnet` with hyphen?
> 
> just my 2c, happy either way ;)
> 
> UPDATE: now I'm not sure if hyphens are even allowed or exist in other chain names.... potentially counters my nit pick.

At some point I heard that chain name would be restricted to alphanumeric (no hyphens)

sunnya97:
@pyramation yeah I made a rule that no hyphens in chainnames, to make it not a bikeshed of whether it should be like `cosmoshub` or `cosmos-hub` lol

pyramation:
> @pyramation yeah I made a rule that no hyphens in chainnames, to make it not a bikeshed of whether it should be like `cosmoshub` or `cosmos-hub` lol

cool that works!

PR #619, opened 2022-08-15
Update Blackhole Chain Info

bkarst:


JeremyParish69:
looks good. With your permission, I'd like to update the token images again once merged, because many dApps will crop the image to an inner circle, and based on how the file borders on the very edge of the image, there'd be some clipping when displayed. Also, the images not being exactly square aspect ratio could cause a squished look. I've created an SGV and PNG on the token logo that will not clip anything and are exactly square.

PR #624, opened 2022-08-17
Add Validatrium endpoints

Validatrium:


bitsong/chain.json
```diff
@@ -88,6 +88,14 @@
...
         "address": "https://rpc.explorebitsong.com",
         "provider": "stake.systems"
+      },
+      {
+        "address": "https://bitsong-rpc.validatrium.club",
+        "provider": "Validatrium"
+      },
+      {
+        "address": "https://bitsong-archive.validatrium.club",
+        "provider": "Validatrium"
```
JeremyParish69, line 98: there is an "archive" property (defaults to false). Could you add `"archive": true` here please?

Issue #615, opened 2022-08-15
`explorers` field is duplicated for Agoric


AlianBenabdallah:
https://github.com/cosmos/chain-registry/blob/master/agoric/chain.json

PR #625, opened 2022-08-17
Fix Agoric explorers

thomasWos:


thomasWos:
Closes https://github.com/cosmos/chain-registry/issues/615

PR #633, opened 2022-08-18
Update assetlist.json to reflect IST denom

jessysaurusrex:
Recently, Agoric renamed RUN Protocol to Inter Protocol. As a result, the denom previously referred to as RUN is now IST, short for the Inter Stable Token. This update changes the denom name and descriptions in the asset list to match the new naming conventions. 

A follow up proposal to update the logos is on the way.

### DIFF
--- a/agoric/assetlist.json
+++ b/agoric/assetlist.json
@@ -3,7 +3,7 @@
   "chain_name": "agoric",
   "assets": [
     {
-      "description": "BLD is the token used to secure the Agoric chain through staking and to backstop the RUN Protocol.",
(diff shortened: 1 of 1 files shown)


JeremyParish69:
Check failed because the base unit 'urun' is no longer defined in the denom units. Is the base unit also changed to uist? or is uist just an alias of urun?

jessysaurusrex:
Oops! Fixed, great catch.

sunnya97:
@jessysaurusrex did the token logo change as well?

jessysaurusrex:
> @jessysaurusrex did the token logo change as well?

Yes-- both will, but the logo change is not final just yet. I'll be back as soon as I get the new logos from our team in a couple of days.

Issue #150, opened 2022-01-09
homepage URL


hxrts:
This is more a convenience thing, but it could be good to optionally link a public URL so clients can look that up. Could also go one step further and include a public key + URL/pubkey.txt, so the webmaster can attest that the website is authenticated by on-chain gov.

Negative here is if projects opt to do this they are enshrining an official page that could imply singular control.

faddat:
The chain should have a web server & a front end, ala:

app.osmosis.zone
blurt.blog

nooomski:
I personally think this can be complicated in terms of demonstrating decentralization. We have a Cosmos page, because the stack is built by several organizations that ICF funds, but we don't have an official Cosmos Hub website, because nobody owns the Cosmos Hub. We can leave it as an optional field, but I don't think we should make it a requirement.

faddat:
I'm with Sam on this, a web site is important-- just I think that it should be served from a chain.

hxrts:
leaving it optional and using an array works for me

JeremyParish69:
We can probably pull a lot of this from the SLIP 173 registration to get started.
https://github.com/satoshilabs/slips/blob/master/slip-0173.md

hxrts:
https://github.com/cosmos/chain-registry/issues/524

PR #636, opened 2022-08-19
Add websites to schema

dylanschultzie:
Simple change that adds network website to chain list. 

If accepted, will go through and add a bunch manually.

Signed-off-by: Dylan Schultz <hello@schultzie.dev>

### DIFF
--- a/chain.schema.json
+++ b/chain.schema.json
@@ -3,6 +3,7 @@
     "$schema": "https://json-schema.org/draft-07/schema",
     "title": "Cosmos Chain",
     "updatelink": "string",
+    "website": "string",
     "description": "Cosmos Chain.json is a metadata file that contains information about a cosmos sdk based chain.",
     "type": "object",
     "required": [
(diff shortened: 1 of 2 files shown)


JeremyParish69:
Thanks. FYI, there was discussion on this #636 . Under consideration...

dylanschultzie:
Ah looks like you're correct: https://github.com/cosmos/chain-registry/issues/150

sunnya97:
I think I'm personally okay with it 🤷

chain.schema.json
```diff
@@ -3,6 +3,7 @@
     "$schema": "https://json-schema.org/draft-07/schema",
     "title": "Cosmos Chain",
     "updatelink": "string",
+    "website": "string",
```
JeremyParish69, line 6: should I assume that you meant to add 'website' under properties? I don't think you meant to have it up here...

dylanschultzie: Whoops, thank you. It's been a while since I've messed with json schema tbh.

PR #432, opened 2022-06-18
Add Wrapped Matic to Axelar asset list

ToggLeTek:

PR #643, opened 2022-08-21
add axlwmatic to Axelar assetlist

JeremyParish69:
replaces #432 
pending addition of traces...

### DIFF
--- a/axelar/assetlist.json
+++ b/axelar/assetlist.json
@@ -423,6 +423,10 @@
           "exponent": 18
         }
       ],
+      "base": "wmatic-wei",
+      "name": "Wrapped Matic",
+      "display": "wmatic",
+      "symbol": "WMATIC",
       "traces": [
         {
           "type": "bridge",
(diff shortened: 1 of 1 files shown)


JeremyParish69:
if we can follow the trace, then we don't need the images here, nor the coingeck_id since there isn't its own coingecko id yet.

axelar/assetlist.json
```diff
@@ -410,6 +410,37 @@
...
+          "non-cosmos": true,
+          "base_denom": "0x0d500b1d8e8ef31e21c99d1db9a6444d3adf1270",
+          "provider": "Axelar"
+        }
+      ],
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/polygon/images/wmatic.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/polygon/images/wmatic.svg"
+      },
+      "coingecko_id": "wmatic"
```
sunnya97, line 443: We should probably just use the Matic coingecko id, not wmatic

https://www.coingecko.com/en/coins/polygon

sunnya97, line 443:
```suggestion
      "coingecko_id": "matic-network"
```

Issue #521, opened 2022-07-18
Add `historical_version_matrix` to Chain registry 


ValarDragon:
We need to store historical version history for chains, to easily be able to sync from genesis to latest tip. (And will likely be of benefit for archive node querying)

I suggest in the codebase section, we add a `historical_versions` key, which takes in a list of objects of the form:
```
{
  height: {height}
  upgradename: {upgrade name}
  upgradeBranch: {git tag}
}
```

Then we can change all syncing scripts to use chain registry

JeremyParish69:
Think this is similar to Unity's issue https://github.com/cosmos/chain-registry/issues/130

ValarDragon:
Can we add this for osmosis? Cc @czarcas7ic if we have the height version numbers handy, then we can start getting genesis sync scripts against this

czarcas7ic:
@ValarDragon 
Start on v3
v4: 1314500
v5: 2383300 (can use v6 binary here)
v6: 2464000
v7: 3401000 (can use v8 binary here)
v8: 4402000
v9: 4707300 (can use v10 binary here)
v10: 4713065
v11: 5432450

ValarDragon:
OK sweet! Let's cut out the v5, v7, v9 releases

czarcas7ic:
@ValarDragon actually it would have to be the opposite right? We would have to cut out the v6, v8, and v10 releases because nodes syncing should upgrade to the next higher binary pre fork.

ValarDragon:
Err specify v5 height, v6 code, etc. For each

sunnya97:
Maybe let's just give chain upgrades a single name, not a list of names.

Issue #130, opened 2022-01-01
Feat: Specify enough information to sync from genesis


UnityChaos:
`chain.json` provides both a `genesis_url` and `recommended_version`, but these are not enough to actually start a mainnet node syncing from genesis. To make this possible, I would propose that we modify the standard `codebase` struct, to include both a `genesis_version`, and an `upgrades` value, which specifies enough information to automate stepping through each upgrade (either with cosmovisor, or through halt heights / external tooling).

ex (for osmosis):
```
"codebase": {
        "git_repo" : "https://github.com/osmosis-labs/osmosis",
        "genesis_version" : "v3.1.0",
        "upgrades" : [
          {
            "height" : 1314500,
            "version" : "v.4.2.0",
            "name" : "v4",
          },
          {
            "height" : 2383300,
            "version" : "v6.0.0",
            "name" : "v5",
          },
        ],
    }
```

PR #645, opened 2022-08-22
feat: add historical versions key

czarcas7ic:
Closes: #521
Closes: #130  

Adds a `genesis_version` key for users to know what version to start with.

Adds a `historical_versions` key that shows each upgrade and its respective height (note the name and upgrade tag mismatch. This is on purpose due to them being forks)


ValarDragon (approved):

ValarDragon:
cc @JeremyParish69 can you update the schema (feel free to edit this branch), and we can get this in?

JeremyParish69:
some questions and suggestions:


1) I'd suggest removing "genesis_version" as its own property, but just putting that version as the first version object within the "historical_versions" array.
2) Does genesis always necessarily start at height 0? If so, height 0 can be the indication of genesis version--otherwise, another property might be required.
3) Could it be useful to have some way to show the next version to help ensure that there are no gaps in version history? E.g., someone could see Osmosis version going from v5 -> 7 -> v9 and wonder if version 6 and 8 were skipped.
4) Could be useful to have a property for the date of version upgrade to help guide readers?
5) What exactly is an 'upgrade tag'? Is this precise universal terminology? I might suggest "id" as an alternative; unless that's too vague.
6) Would it make sense to allow for a URL with each version? We could pair not only the genesis URL with the genesis version, but also a pruned state URL with later versions (would that work?).
7) Want to move "genesis_url" to codebase--just makes sense to me.
8) Want to allow for a "compatible_versions" list with each version. (see drafted schema)

czarcas7ic:
> some questions and suggestions:
> 
> 1. I'd suggest removing "genesis_version" as its own property, but just putting that version as the first version object within the "historical_versions" array.
> 2. Does genesis always necessarily start at height 0? If so, height 0 can be the indication of genesis version--otherwise, another property might be required.
> 3. Could it be useful to have some way to show the next version to help ensure that there are no gaps in version history? E.g., someone could see Osmosis version going from v5 -> 7 -> v9 and wonder if version 6 and 8 were skipped.
> 4. Could be useful to have a property for the date of version upgrade to help guide readers?
> 5. What exactly is an 'upgrade tag'? Is this precise universal terminology? I might suggest "id" as an alternative; unless that's too vague.
> 6. Would it make sense to allow for a URL with each version? We could pair not only the genesis URL with the genesis version, but also a pruned state URL with later versions (would that work?).
> 7. Want to move "genesis_url" to codebase--just makes sense to me.
> 8. Want to allow for a "compatible_versions" list with each version. (see drafted schema).
> 9. Want to allow for a list of names. E.g., Osmosis has also used chemical elements as version identifiers, and not only version numbers.

1. Sounds good
2. Technically its possible to be non zero, but its a fine assumption to make
3. I would just put the same version tags for the missing versions, I am fine with either way
4. I think the version number increments is sufficient and getting all chains to pull dates might be difficult
5. "tag" was used since its the github tag used that the user needs to pull. I think tag is better than ID
6. If there was a URL for each version it would be a URL to either the github tag or to the raw binaries link
7. I have no preference so either is fine to me
8. That might get messy / I just think listing the latest version for that height is all thats needed
9. No preference

JeremyParish69:
the next step would be to update all the existing chain.json schema to use this, and then remove the "binaries" and "genesis_url" objects from the top level (because it'll then already be in the codebase section)

czarcas7ic:
@JeremyParish69 Yeah what you pushed makes total sense to me, well done!

sunnya97:
This looks pretty good to me

PR #646, opened 2022-08-22
add a new evm compatible chain: okc exchain

ItsFunny:


okxchain/assetlist.json
```diff
@@ -0,0 +1,28 @@
+{
+  "$schema": "../assetlist.schema.json",
+  "chain_name": "okxchain",
+  "assets": [
+    {
+      "description": "EVM compatible, Cosmos SDK based chain.",
```
JeremyParish69, line 6: This description should apply to the asset, not to the chain itself.

okxchain/assetlist.json
```diff
@@ -0,0 +1,28 @@
...
+          "denom": "okt",
+          "exponent": 18
+        }
+      ],
+      "base": "wei",
+      "name": "OKXChain",
+      "display": "OKT",
+      "symbol": "OKT",
+      "logo_URIs": {
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/okxchain/images/okc.svg",
```
JeremyParish69, line 22: these images look more like images to be used for the chain, and if so, please use these images for the chain logo. Please instead point the asset's logo URIs to a token logo (would that be the Black OKX-square.png?).

okxchain/chain.json
```diff
@@ -0,0 +1,108 @@
+{
+    "$schema": "../chain.schema.json",
+    "chain_name": "okxchain",
+    "status": "live",
+    "network_type": "mainnet",
+    "pretty_name": "OKC",
```
JeremyParish69, line 6: Are we sure we want the chain's pretty name to be "OKC"? This is meant to be how people should say the name of the chain. It is my opinion that a user reading this is unlikely to know what this means, whereas something like "OKEx Chain" would be instantly recognizable.

okxchain/assetlist.json
```diff
@@ -0,0 +1,28 @@
...
+          "exponent": 0
+        },
+        {
+          "denom": "okt",
+          "exponent": 18
+        }
+      ],
+      "base": "wei",
+      "name": "OKXChain",
+      "display": "OKT",
```
JeremyParish69, line 19: should be lowercase. OKT doesn't not exist among the denom_units

okexchain/chain.json
```diff
@@ -0,0 +1,108 @@
+{
+    "$schema": "../chain.schema.json",
+    "chain_name": "exchain",
```
JeremyParish69, line 3: chain name in both files must match folder name. Folder: okexchain, assetlist.json: okxchain, chain.json: exchain. Please pick just one to use in all 3 places.

JeremyParish69: imo, 'exchain' would be harder to find in an alphabetized list than something starting with 'ok...', so I'd recommend either 'okxchain' or 'okexchain'

ItsFunny: Got it ,Thanks

PR #652, opened 2022-08-24
Nomic assets

Galadrin:
Add Nomic Asset


JeremyParish69:
Checks fail because the chain.json have some issues:
-chain.json::pretty_name should be just "Nomic"
-chain.json::brinaries points to an akash binary??

can we get these corrected please?

PR #669, opened 2022-08-30
Impacthub version updates

greg-szabo:
- Changed recommended version to 0.18.1
- Changed compatible versions to 0.18.0 and 0.18.1
- Removed one RPC node IP address that does not respond
- Removed invalid links to binaries (kept the keys and made them empty string, so we don't break any automations. Maybe we get binaries at a later date.)

### DIFF
--- a/impacthub/chain.json
+++ b/impacthub/chain.json
@@ -26,15 +26,16 @@
     },
     "codebase": {
         "git_repo": "https://github.com/ixofoundation/ixo-blockchain",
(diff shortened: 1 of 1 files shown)


sunnya97:
I think better to remove binary keys altogether.  It's an optional key in the json schema, so scripts should be expected to handle it

greg-szabo:
Very good point, I was too tired the check if they are optional in the schema. Fixed.

Surprisingly little is required according to the schema...

PR #677, opened 2022-08-30
update Notional endpoints 20220830

baabeetaa:


JeremyParish69:
can you please check on the conflict?

PR #681, opened 2022-08-31
IBC files::Replace dashes with underscores

### LINKED ISSUE #632: `_IBC` doesn't follow naming conventions of all other JSON files
everything uses `underscore_casing` and `_IBC` uses the least-used JSON notation style with `dashed-case` which is very hard to work with in JS and JSON in general due to escaping.

Can we use convention and stick to underscore?

e.g. chain-1 => chain1 and channel-id => channel_id?

JeremyParish69:
fixes #632 
please see, @pyramation


pyramation:
wow amazing !

PR #687, opened 2022-09-05
Add Commercio.network

zenit70:
Add Commercio.network chain.

Thanks


commercionetwork/chain.json
```diff
@@ -0,0 +1,89 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "commercionetwork",
+  "chain_id": "commercio-3",
+  "pretty_name": "Commercio.network",
+  "status": "live",
+  "network_type": "mainnet",
+  "bech32_prefix": "did:com:",
```
sunnya97, line 8: Is this a valid bech32 path?

JeremyParish69: it's been registered to SLIP173 a while

zenit70: @sunnya97 Yes, is a valid path. As specified in https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#Bech32 the range of characters allowed for hrp should also include “:”

sunnya97: Wow, TIL

PR #732, opened 2022-09-19
Add juno testnet gRPC endpoint

de-husk:
Adds the polkachu juno testnet gRPC endpoint to config.

I tested that the gRPC url works, and it also follows the same format as the other Polkachu gRPC urls.

### DIFF
--- a/testnets/junotestnet/chain.json
+++ b/testnets/junotestnet/chain.json
@@ -44,9 +44,7 @@
     "cosmwasm_enabled": true
   },
   "peers": {
-    "seeds": [
-      
-    ],
+    "seeds": [],
     "persistent_peers": [
       {
         "id": "ed90921d43ede634043d152d7a87e8881fb85e90",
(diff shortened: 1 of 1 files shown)


psibi:
I don't think it works anymore.

de-husk:
gRPC is working for me. Its the first RPC endpoint that is working for me right now: https://rpc.uni.junomint.com

psibi:
@de-husk Maybe I'm missing something here, but the test gRPC endpoint doesn't seem to be working for me:

```
❯ grpcurl juno-testnet-grpc.polkachu.com:12690 list
Failed to dial target host "juno-testnet-grpc.polkachu.com:12690": tls: first record does not look like a TLS handshake
```

de-husk:
> @de-husk Maybe I'm missing something here, but the test gRPC endpoint doesn't seem to be working for me:
> 
> ```
> ❯ grpcurl juno-testnet-grpc.polkachu.com:12690 list
> Failed to dial target host "juno-testnet-grpc.polkachu.com:12690": tls: first record does not look like a TLS handshake
> ```

Have you tried `-plaintext`?

psibi:
Thanks, that is indeed working.

Issue #658, opened 2022-08-25
Move genesis URL to codebase


JeremyParish69:
Following #645 , genesis url was added to codebase section--now we can move existing genesis url info from the top level over to the codebase section for all chains. Once that's done, we can safely remove the top-level genesis url from the schema without any data loss.

PR #738, opened 2022-09-21
Move genesis to codebase

JeremyParish69:
resolves https://github.com/cosmos/chain-registry/issues/658

note: many of the comparisons are difficult to utilize, because of the formatting change. It'd be best to stick to consistent formatting so that bulk changes can be made.

Edit: now also moves node_home and daemon_name to "codebase"


pyramation:
LGTM!

My main thought with this is, if we move `genesis`, we should move things like `node_home` that are clearly `codebase` related.

JeremyParish69:
PR to only copy the properties to "codebase" (but not remove the top level properties, yet): https://github.com/cosmos/chain-registry/pull/749

JeremyParish69:
Closed by https://github.com/cosmos/chain-registry/pull/838

PR #755, opened 2022-09-25
Add AgoraNodes EVMOS RPC, gRPC, JSON-RPC, API

AgoraNodes:


evmos/chain.json
```diff
@@ -107,6 +107,11 @@
         "id": "d4a5fcc56f369665c232788c22844dc3a018fe08",
         "address": "peer.evmos.silknodes.io:26602",
         "provider": "Silk Nodes"
+      }
```
JeremyParish69, line 110: need comma here (for each one you added)

evmos/chain.json
```diff
@@ -171,6 +176,10 @@
       {
         "address": "https://rpc.evmos.silknodes.io/",
         "provider": "Silk Nodes"
+      }
```
JeremyParish69, line 179: still missing comma here

JeremyParish69 (changes requested):
still need comma for rpc and peer

AgoraNodes:
added commas should be good to go

PR #757, opened 2022-09-25
Delete chronicnetwork directory

discoverdefiteam:
Chronic Network has been sunset via governance.


JeremyParish69:
Thanks for the news. I believe we allow for `"status": "killed"` for this purpose.

terpnetworkcommunity:
> Thanks for the news. I believe we allow for `"status": "killed"` for this purpose.

Should we re-request this PR? if so, what would the format desired be?

PR #763, opened 2022-09-25
Add Cypher Core Public API

jim380:


cosmoshub/chain.json
```diff
@@ -288,6 +288,10 @@
       {
         "address": "cosmos-grpc.polkachu.com:14990",
         "provider": "Polkachu"
+      },
+      {
+        "address": "https://cosmos-api.cyphercore.io",
```
thomasWos, line 293: This is the grpc definition, rest API instead?

jim380: @thomasWos Fixed. Thanks for catching that.

JeremyParish69:
Looks like there are conflicts now

jim380:
@JeremyParish69 conflicts resolved. Thanks!

PR #781, opened 2022-09-28
update point network

DaMandal0rian:
add coingecko_id to assetlist


yurivin:
IBC cosmoshub-point network config file added to the pull request

PR #788, opened 2022-09-29
Add AutoStake Seeds

AutoStake-net:
Add seeds for all the network we validate


dylanschultzie (approved):

terra/chain.json
```diff
@@ -83,10 +88,6 @@
         "address": "https://blockdaemon-terra-lcd.api.bdnodes.net:1317",
         "provider": "Blockdaemon"
       },
-      {
-        "address": "https://terra.mainnet.lcd.blockngine.io:443",
-        "provider": "BlockNgine Validators"
-      },
```
JeremyParish69, line 89: why remove this?

AutoStake-net: That is ours, we rebranded from BlockNgine to AutoStake

We no longer ave a public LCD for terraclassic, therefore this was just a cleanup

PR #796, opened 2022-09-30
Add POSTHUMAN's rpc/rest/grpc nodes

albertandrejev:


dylanschultzie:
@albertandrejev should all of validator.network's resources be removed? They've been inactive for a while now.

dylanschultzie (approved):

albertandrejev:
> @albertandrejev should all of validator.network's resources be removed? They've been inactive for a while now.

Not sure that I understand your question. Because validator.network is not our resource, but I expect that you are talking about https://validator.network/#/ ? Anyway this resource is not related to us. We have https://validators.network/ as side project, pay attention on additional **s** in the URL. 
Btw do you know why our changes can not pass "Validate Data" workflow?

Thanks you!

PR #799, opened 2022-09-30
Fix Stride path

dylanschultzie:


dylanschultzie:
This is the *actual* folder that is created when doing `strided init`

Screen Shot 2022-09-30 at 5 26 37 PM

PR #800, opened 2022-10-01
add stargaze testnet

jhernandezb:


shanev (approved):

pyramation:
hey @JeremyParish69 — are we validating testnets against satoshilabs/slips?

https://github.com/satoshilabs/slips/pull/1417

they just closed the PR because of some arbitrary rule, and to be honest, osmo and juno don't even have their testnets so not sure why we're validating against this.

pyramation:
I'm gonna try and fix the validation script.

pyramation:
I talked to @sunnya97 he said we should edit `validate_data.py` to not check testnets against satoshi site. 

I would do it but my py env is not ideal. Would be great if we can fix this before hackwasm is finished 

cc @JeremyParish69

jhernandezb:
Seems like the format change it which is also making it to fail.

You can see from this commit that the [Name](Website) format was removed for every chain.

https://github.com/satoshilabs/slips/commit/d4ce06e5bdbc1f2c235a02c2b7bff2803d72625f

JeremyParish69:
Indeed, the SLIPs reformatted which broke the script. For now the check is disabled, so it shouldn't cause any more issue, and the repaired version will not check testnets against SLIPs

PR #815, opened 2022-10-03
Add teritori support

Nodeist:


Nodeist:
It looks like there is no problem, but the check is faulty.

PR #831, opened 2022-10-08
add silent validator public endpoints for evmos,omniflix,stargaze ...

silentnoname:


leafwind-web3 (approved):
Thanks for the contribution!

PR #834, opened 2022-10-09
Update agoric-axelar metadata

jtieri:
The existing light clients between Agoric and Axelar were misconfigured so the channels ended up being closed.

This is the updated information.


dckc:
I was able to transfer from Axelar to Agoric using this channel. FWIW.

PR #866, opened 2022-10-16
update evmosd and odind version

doggystylez:


doggystylez:
official dragonberry patches for evmos and odin

PR #868, opened 2022-10-17
Add Wallets property to chain jsons

JeremyParish69:
added 
"wallets": [
  "keplr"
],
with some json schema validation in the case that "keplr" is specified--must be able to work with [Keplr's suggest chain feature](https://docs.keplr.app/api/suggest-chain.html).

sample:
    "required": [
      "chain_id",
      "bech32_prefix",
      "slip44",
      "fees",
      "staking",
      "apis",
      "codebase"
    ]
but the requirements go more in-depth, e.g., requiring low, avg, and high gas fee rates, and more.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,9 @@ A sample `chain.json` includes the following information.
   },
   "keywords" [
     "dex"
+  ],
+  "wallets": [
+    "keplr"
   ]
 }
 ```
--- a/chain.schema.json
+++ b/chain.schema.json
@@ -252,7 +252,92 @@
       "items": {
         "$ref": "#/$defs/explorer"
       }
+    },
+    "wallets": {
+      "type": "array",
+      "items": {
+        "type": "string"
+      }
     }
+  },
+  "if": {
+    "properties": {
+      "wallets": {
+        "type": "array",
+        "contains": {
+          "type": "string",
+          "enum": [
+            "keplr"
+          ]
+        },
+        "minContains": 1
+      }
+    },
+    "required": [
+      "wallets"
+    ]
+  },
+  "then": {
+    "properties": {
+      "fees": {
+        "type": "object",
+        "properties": {
+          "fee_tokens": {
+            "type": "array",
+            "contains": {
+              "type": "object",
+              "required": [
+                "denom",
+                "low_gas_price",
+                "average_gas_price",
+                "high_gas_price"
+              ]
+            },
+            "minContains": 1
+          }
+        },
+        "required": [
+          "fee_tokens" 
+        ]
+      },
+      "staking": {
+        "type": "object",
+        "properties": {
+          "staking_tokens": {
+            "minContains": 1,
+            "maxContains": 1
+          }
+        },
+        "required": [
+          "staking_tokens" 
+        ]
+      },
+      "apis": {
+        "type": "object",
+        "properties": {
+          "rpc": {
+            "minContains": 1
+          },
+          "rest": {
+            "minContains": 1
+          }
+        }
+      },
+      "codebase": {
+        "required": [
+          "cosmos_sdk_version"
+        ]
+      }
+    },
+    "required": [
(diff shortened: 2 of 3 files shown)


pyramation:
I can see the idea here — but hear me out – what if it's better to have wallets and then an array of chain names?

Then we can have a "wallet registry" of sorts, and it *should* be up to the wallet (e.g. keplr or leap) to manage their supported chains.

Now, using `experimentalSuggestChain()` on a wallet client I don't think means it should be listed here. I think it should list the "supported" wallets, meaning ones whose chain info is already natively in the wallet extensions.

ankurdotb (commented):
I like this idea a lot, and agree with @pyramation that it requires a wallet schema. However, experimental chain suggestion as a property _generically_ is a wallet schema question ("does this wallet allow chain suggestion?") vs in a chain schema which should have a property for whether THIS chain on THIS wallet is supported natively vs via chain suggestion.

Also, would Keplr browser vs Keplr mobile be split out in this scenario? IMO it should, because cheqd (for example) can be added to Keplr via chain suggestion in browser extension, but not in Keplr Mobile.

JeremyParish69:
I really appreciate the feedback.

My initial driver for this was more-so meant to have some quick and easy automated validation for compatibility with the suggest chain feature, and then later thought that it could expand to show wallets.

I agree that having something like a 'wallet registry' would be most efficient for denoting native support for chains on each wallet. I'm generally also in the same camp when it comes to explorers, too.

Meanwhile, an upside to having the wallets instead specified in the chain's json file is that it makes for easier JSON Schema validation. Alternatively relying mostly on something like a python script to validate can sometimes be a bit more messy or fragile, but still works, and is ultimately more powerful.

Changing the array of wallet names to something a bit more detailed would be a relatively small change, (e.g.,:
```
"wallets": [
  {
    "wallet": "keplr_browser",
    "native_support": true,
    "derived_support": true   //this shows suggest chain compatibility
  },
  {
    "wallet": "keplr_mobile",
    "native_support": true,
    "derived_support": true
  },
],
```

while a Wallet Registry would take a bit more planning and rework. I'll see what I can come up with...

pyramation:
> Also, would Keplr browser vs Keplr mobile be split out in this scenario? IMO it should, because cheqd (for example) can be added to Keplr via chain suggestion in browser extension, but not in Keplr Mobile.

Yes, we should use `keplr_extension` and `keplr_mobile` to match what was suggested by chainapsis team for cosmos kit related work

pyramation:
> My initial driver for this was more-so meant to have some quick and easy automated validation for compatibility with the suggest chain feature, and then later thought that it could expand to show wallets.

I think this is cool! I would suggest this isn't manually entered in, but instead baked into tooling to automatically add the attribute. Perhaps, for example, we have the CI/CD parse the wallet registry for the required properties, and then update the chain.json files (or maybe every chain has a new wallet.json) 

A field like this where source of truth comes from the wallet, all the the relational data between the wallet/chain can be 100% derived and generated.

pyramation:
> Meanwhile, an upside to having the wallets instead specified in the chain's json file is that it makes for easier JSON Schema validation. Alternatively relying mostly on something like a python script to validate can sometimes be a bit more messy or fragile, but still works, and is ultimately more powerful.

Side note regarding this...  but let me suggest we move away from python -> node and it will get a lot cleaner and you can have more folks contribute.

pyramation:
> I like this idea a lot, and agree with @pyramation that it requires a wallet schema. However, experimental chain suggestion as a property _generically_ is a wallet schema question ("does this wallet allow chain suggestion?") vs in a chain schema which should have a property for whether THIS chain on THIS wallet is supported natively vs via chain suggestion.

So I think it definitely makes sense to know:
1. what are the native chains that a wallet supports
2. does wallet support a suggest chain feature?
3. if 2 = yes, what are the properties required for suggest chain?

PR #871, opened 2022-10-17
Add Schema for Wallet Registry

JeremyParish69:
Creates a new \_wallets\ folder for a 'wallet registry'. Keplr is the first example used. There is some JSON schema validation, however scripting is likely required to be able to fully automate data validation.

If used, will work to gather more information about the various wallets. More complete information about Keplr, as well as Leaf, Cosmostation Wallet, Metamask, and more.


pyramation:
let's look at using nodejs for the scripts?

JeremyParish69:
I'd like to suggest that the 'suggestible' property (open to renaming, e.g., to 'permissionless') could be just a temporary field, as this information is technically derivable from data elsewhere in the Chain Registry, and thus doesn't need manual specification.

pyramation:
Here I propose an alternate form that is a bit simpler.

```
{
  "$schema": "../wallet.schema.json",
  "wallet_name": "keplr_extension",
  "pretty_name": "Keplr Extension",
  "website": "https://www.keplr.app/",
  "platforms": [
    "desktop_browser_chrome"
  ],
  "has_suggest_feature": true,
  "supported_chains": [
      "cosmoshub",
      "osmosis"
  ]
}
```

What was `native` for in the `supported_chains` objects? I could maybe see the need for objects, but an array of `chain_name`s in `supported_chains` is pretty much all the info I believe we need to know.

`suggestible` would be very redundant, and could be a maintenance issue. If a wallet `has_suggest_feature:true`, then it would allow us to suggest any chain, so we shouldn't have to arbitrarily list every suggestible chain. This list (similar to asset lists) should be the natively supported chains.

JeremyParish69:
> Here I propose an alternate form that is a bit simpler.
> 
> ```
> {
>   "$schema": "../wallet.schema.json",
>   "wallet_name": "keplr_extension",
>   "pretty_name": "Keplr Extension",
>   "website": "https://www.keplr.app/",
>   "platforms": [
>     "desktop_browser_chrome"
>   ],
>   "has_suggest_feature": true,
>   "supported_chains": [
>       "cosmoshub",
>       "osmosis"
>   ]
> }
> ```
> 
> What was `native` for in the `supported_chains` objects? I could maybe see the need for objects, but an array of `chain_name`s in `supported_chains` is pretty much all the info I believe we need to know.
> 
> `suggestible` would be very redundant, and could be a maintenance issue. If a wallet `has_suggest_feature:true`, then it would allow us to suggest any chain, so we shouldn't have to arbitrarily list every suggestible chain. This list (similar to asset lists) should be the natively supported chains.

I'd be on board for this way. But should we include a way to convey the requirements for suggesting a chain? Could be new set of schema files?
E.G., keplr_extension_suggest.schema.json
```
"properties": {
  "chain_id": {
    "type": "string"
  },
  "slip44": {
    "type": "number"
  }
  "apis": {
    "type": "object",
    "properties": {
      "rpc": {
        "type": "array",
        "minContains": 1
      },
     "rest": {
        "type": "array",
        "minContains": 1
      },
      "required": [
        "rpc",
        "rest"
      ]
    }
  },
  ...
},
"required": [
  "chain_id",
  "slip44",
  "apis",
  ...
]
```

JeremyParish69:
> I think I would say yes, would be super nice to have. Issue is, this feature could block in the sense that, it's a bit more complex.
> 
> For example, [in chain-registry->keplr types](https://github.com/cosmology-tech/chain-registry/blob/f3fcb1ff1456f6d95c53c3062b677ce0f154437f/packages/keplr/src/index.ts#L25-L27) we need BOTH `Chain` and `AssetList[]` to construct the data necessary. Is there a way describe that requirement across chain and assetlist?

Although we can require data from both files, you're right in that a truly thorough validation can not be done by looking at the files separately. We'd need something like a combined file to be able to do it with JOSN schema validation alone.

JeremyParish69:
@pyramation Please see how the code has been modified much closer to your suggestion. Specification of suggest requirements and validation of such has just been left out of scope for now.

sunnya97:
Can we move this to a new repo?  Want to keep chain registry about chain and ibc metadata

pyramation:
@sunnya97 and I were chatting we also think we should potentially include install links.

Additionally, if it's supported on firefox vs. chrome, ios, android.

E.g. for keplr:

Chrome: https://chrome.google.com/webstore/detail/keplr/dmkamcknogkgcdfhhbddcghachkejeap?hl=en
Firefox: https://addons.mozilla.org/en-US/firefox/addon/keplr/
Android: https://play.google.com/store/apps/details?id=com.chainapsis.keplr
IOS: https://apps.apple.com/us/app/keplr-wallet/id1567851089

JeremyParish69:
> Can we move this to a new repo? Want to keep chain registry about chain and ibc metadata

Does this imply that we should also eventually remove all data about third-party services (anything not considered neutral metadata) to a separate repo? E.g., a repo for block explorers, and a repo for informational APIs (like CoinGecko, CMC, Imperator, etc.), and a repo for blockchain endpoints, etc.?
Whatever we did with wallets, I was hoping to do something similar with Explorers next, and keep going from there.

Also, sure, I see the reason for a new repo and could be fun to build with @pyramation.

JeremyParish69:
Cosmos Wallet Registry just published: https://github.com/cosmos/wallet-registry

PR #873, opened 2022-10-18
Adding CosmWasm's Malaga-420 testnet

msteiner96:


testnets/cosmwasmtestnet/chain.json
```diff
@@ -0,0 +1,77 @@
...
+    "rpc": [
+      {
+        "address": "https://rpc.malaga-420.cosmwasm.com:443",
+        "provider": "CosmWasm"
+      }
+    ],
+    "rest": [
+      {
+        "address": "https://api.malaga-420.cosmwasm.com",
+        "provider": "CosmWasm"
```
webmaster128, line 66: Can you change those providers to "Confio"?

testnets/cosmwasmtestnet/chain.json
```diff
@@ -0,0 +1,77 @@
...
+  "staking": {
+    "staking_tokens": [
+      {
+        "denom": "uand"
+      }
+    ]
+  },
+  "codebase": {
+    "git_repo": "https://github.com/CosmWasm/wasmd",
+    "recommended_version": "v0.279",
```
webmaster128, line 34: That version looks broken

testnets/cosmwasmtestnet/assetlist.json
```diff
@@ -0,0 +1,40 @@
+{
+  "$schema": "../../assetlist.schema.json",
+  "chain_name": "malaga-420",
```
JeremyParish69, line 3: the chain name should match the folder name, so 'cosmwasmtestnet' in this case.

testnets/cosmwasmtestnet/assetlist.json
```diff
@@ -0,0 +1,40 @@
...
+          "denom": "umlg",
+          "exponent": 0
+        },
+        {
+          "denom": "MLG",
+          "exponent": 6
+        }
+      ],
+      "base": "umlg",
+      "name": "Malaga Testnet",
```
JeremyParish69, line 18: These name are meant to describe the token, rather than the chain... both tokens are to be called "Malaga Testnet"?

testnets/cosmwasmtestnet/chain.json
```diff
@@ -0,0 +1,77 @@
+{
+  "$schema": "../../chain.schema.json",
+  "chain_name": "malaga-420",
+  "status": "live",
+  "network_type": "testnet",
+  "pretty_name": "Malaga-420",
```
JeremyParish69, line 6: Should this instead be "Malaga Testnet" or "Cosmwasm Testnet"?

testnets/cosmwasmtestnet/assetlist.json
pyramation: likely Malaga — I'll pass these updates on to Milan

pyramation:
I would suggest we call this `cosmwasmtestnet` since this is what it is for. It just so happens that the token is Malaga (Great name btw!)

@msteiner96 I would suggest one of two paths:

- [ ] update `chain_name` to `cosmwasmtestnet` 
- [ ] update `pretty_name` to `CosmWasm Testnet`
- [ ] update token name in asset list to `Malaga` 

or, if you prefer to use `Malaga Testnet`, rename the folder and do similar for `Malaga` instead of `CosmWasm`

msteiner96:
Done :)

testnets/cosmwasmtestnet/assetlist.json
JeremyParish69: looks like this was forgotten. I believe this would be the last place that needs updating

pyramation:
one final change in assetlist.json - change chain_name to `cosmwasmtestnet` and we're good!

testnets/cosmwasmtestnet/assetlist.json
pyramation: @msteiner96 this is the final change needed

pyramation: I'll just merge and fix this one.

PR #891, opened 2022-10-20
remove validator's  peers

RaulBernal:
MainNet remove itastaker


JeremyParish69:
Looks like these have already been removed, so this PR can be closed.

PR #892, opened 2022-10-20
Another running devnet - testing SDK v0.46.x

RaulBernal:
Is not a replacement, is a new testnet, thanks


RaulBernal:
Why it fails? should the name of the folder be the same that the `chain-name`? `chain-name` is the `chain-id` ?
   ` "chain_name": "bitcanna-dev-6",`
I could  change it to `bitcannadev2`

testnets/bitcannadev2/assetlist.json
```diff
@@ -0,0 +1,28 @@
+{
+    "$schema": "../../assetlist.schema.json",
+    "chain_name": "bitcanna-dev-6",
```
JeremyParish69, line 3: chain_name != chain_id. Please change chain_name to 'bitcannadev2', as it must match the folder name exactly. Same goes for the chain_name property of the chain.json file as well.

RaulBernal (commented):
updated chain_name
Thanks :)

testnets/bitcannadev2/assetlist.json
```diff
@@ -0,0 +1,28 @@
...
+                    "denom": "bcna",
+                    "exponent": 6
+                }
+            ],
+            "base": "ubcna",
+            "display": "bcna",
+            "name": "BitCanna",
+            "symbol": "BCNA",
+            "logo_URIs": {
+                "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/bitcanna-dev-6/images/bcna.png",
```
JeremyParish69, line 22: oh, looks like it still fails validation. I think it's because of the locations here are incorrect. Can you correct the folder name on these images please? (\bitcanna-dev-6\images... -> \bitcannadev2\images...)

RaulBernal (commented):
Done :)

RaulBernal:
Thanks

JeremyParish69:
> Thanks

Actually, if I may, I'd like to change the network type to 'devnet' instead of 'testnet'? (I still need to add the option for it)

PR #894, opened 2022-10-20
Added Unification Blockchain

andreas-barthel:


unification/assetlist.json
```diff
@@ -0,0 +1,27 @@
+{
+    "$schema": "../assetlist.schema.json",
+    "chain_name": "FUND-MainNet-2",
```
JeremyParish69, line 3: should be 'unification'

andreas-barthel: Done. And Sorry :)

PR #904, opened 2022-10-21
Add devnet network type

JeremyParish69:


JeremyParish69:
@RaulBernal please see chg to bitcannadev2 network type

RaulBernal:
> @RaulBernal please see chg to bitcannadev2 network type

Yes, sounds good =)

RaulBernal:
> @RaulBernal please see chg to bitcannadev2 network type

Yes awesome! Right

keupsonite:
@JeremyParish69 what is the point of adding a new type of network? What is the difference between testnet and devnet? Wouldn't it be better to modify bitcanna's devnet with the testnet type?

I ask because I personally use this library for our explorer and we differentiate the networks only by mainnet/testnet, not devnet. This would add yet another condition, which doesn't seem to have any real use.

Thanks in advance.

JeremyParish69:
Hi @keupsonite , thanks for the interest. I think the argument for or against adding (arbitrarily many?) more network types isn't very strong either way at the moment. Sure, we _could_ separate only by 'mainnet' and 'not-a-mainnet', or alternatively, we could have many categories as they start to become popular (mainnet, testnet, devnet, edgenet?, etc.). I'd argue that since professional software release cycles often encompass more than just a 'test' and 'production' environment(, e.g., qa/qc/uat), additional network types could have some benefit.

Because you use this, I'll also invite you to join the Telegram Group where we discuss high level design of the Chain Registry and better keep up to date with upcoming changes, to which you can find a link in the README.

keupsonite:
Hi @JeremyParish69, thanks for your answer. BitCanna only seems to have devnets _(two to be precise),_ in the testnets folder by the way, but yes your explanation makes perfect sense. I will join the telegram thanks for the suggestion.

PR #911, opened 2022-10-24
Update osmosis versions

niccoloraspa:
This PR:
- fixes some inconsistencies in the osmosis versions:

```yaml
    "name": "v5",        # wrong
    "name": "v6",        # correct
    "tag": "v6.4.1",
```
 - bumps all major version to the latest patch release

```yaml
    "tag": "v10.0.1",  # wrong
    "tag": "v10.1.1"
```

### DIFF
--- a/osmosis/chain.json
+++ b/osmosis/chain.json
@@ -68,32 +68,32 @@
         "next_version_name": "v5"
       },
       {
-        "name": "v5",
+        "name": "v6",
         "tag": "v6.4.1",
         "height": 2383300,
         "next_version_name": "v7"
       },
       {
-        "name": "v7",
+        "name": "v8",
         "tag": "v8.0.0",
         "height": 3401000,
         "next_version_name": "v9"
       },
       {
-        "name": "v9",
-        "tag": "v10.0.1",
+        "name": "v10",
+        "tag": "v10.1.1",
         "height": 4707300,
         "next_version_name": "v11"
       },
       {
         "name": "v11",
-        "tag": "v11.0.0",
+        "tag": "v11.0.1",
         "height": 5432450,
         "next_version_name": "v12"
       },
       {
         "name": "v12",
-        "tag": "v12.1.0",
+        "tag": "v12.2.0",
         "height": 6246000
       }
     ]
--- a/osmosis/chain.json
+++ b/osmosis/chain.json
@@ -65,19 +65,19 @@
         "name": "v4",
         "tag": "v4.2.0",
         "height": 1314500,
-        "next_version_name": "v5"
+        "next_version_name": "v6"
       },
       {
         "name": "v6",
         "tag": "v6.4.1",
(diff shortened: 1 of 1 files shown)


JeremyParish69:
Thanks @niccoloraspa .
Due to some automation, the entire codebase section here for Osmosis is automatically updated based on a file upstream: [this schema file in the Osmosis repo](https://github.com/osmosis-labs/osmosis/blob/main/chain.schema.json). Changes should instead be made there, or else it will just be automatically overwritten back.
Regarding the bumping of major version to the latest patch release, I agree this is a good idea.
Regarding the correcting of names of version (e.g., v5 to v6), although I don't think it matter too much, I believe your suggestion may be incorrect. The name v5 for tag v6 wasn't a typo. I recall for a while the version names (which at one time took after the names of elements of the periodic table of elements) were definitely out of sync. For example, the Nitrogen (atomic #7) had the upgrade tag of 'v8'.

niccoloraspa:
Thanks Jeremy, I will update the schema in the Osmosis repo!

> Regarding the correcting of names of version (e.g., v5 to v6), although I don't think it matter too much, I believe your suggestion may be incorrect. The name v5 for tag v6 wasn't a typo. I recall for a while the version names (which at one time took after the names of elements of the periodic table of elements) were definitely out of sync. For example, the Nitrogen (atomic number 7) had the upgrade tag of 'v8'.

I think the underlying issue is that some versions are forks due to bugs.  For example, `v6` is the *correct* working version of the `v5` release

```json
      {
        "name": "v5",
        "tag": "v6.4.1",
        "height": 2383300,
        "next_version_name": "v7"
      },
```

However, the pattern is not very consistent as the next version jumps to `v7` ??? 

```json
      {
        "name": "v7",                
        "tag": "v8.0.0",
        "height": 3401000,
        "next_version_name": "v9"
      },
```

I agree that it's not a big deal, however, it's very confusing to have versions and tags so misaligned and it could be a source of bugs in the future.

I propose two options to bring a little bit more consistency:

1. We use as `name `the major version in the `tag` field (as in this PR)

```json
      {
        #"name": "v7",        NO
        "name": "v8",               
        "tag": "v8.0.0",
        "height": 3401000,
        "next_version_name": "v9"
      },
```

2. We use the atomic numbers in the `name`:

```json
      {
        #"name": "v7",             NO
        "name": "nitrogen"    
        "tag": "v8.0.0",
        "height": 3401000,
        "next_version_name": "v9"
      },
```

I'm happy either way

JeremyParish69:
@niccoloraspa this is also discussed briefly here:
https://github.com/cosmos/chain-registry/issues/521
I also recall having a names array for each, so we could have both element name and a 'v#' name, but that was deemed unnecessary. I struggle to recall why we opted over v# over the elemental name, but someone may have preferred it that way?? Either way, a change like you've suggested--using element names instead--should be possible.

PR #912, opened 2022-10-24
Update Teritori's logos and coinGecko id

keupsonite:


keupsonite:
no idea why the CI fails @JeremyParish69

JeremyParish69:
> no idea why the CI fails @JeremyParish69

Seemed to be a fluke issue installing a Python dependency. Re-ran the validation and it worked just fine.

PR #922, opened 2022-10-26
update strided version

doggystylez:


JeremyParish69:
@doggystylez thanks. Are we able to retain links for binaries as well?

doggystylez:
@JeremyParish69 compiled binaries are not available for this version https://github.com/Stride-Labs/stride/releases/tag/v2.0.3

PR #924, opened 2022-10-26
add cosmwasm logo

pyramation:


JeremyParish69:
@pyramation thank you. Both tokens use the same logo?

pyramation:
yes, as it's for testnet we're just working to get the demo looking better. We can update the fee token, I'll talk to Ethan.

PR #985, opened 2022-11-08
feat(registry): add SIENNA

mradkov:
This PR adds SIENNA (SNIP-20 token on Secret Network) to the chain-registry.


secretnetwork/assetlist.json
```diff
@@ -24,6 +24,28 @@
         },
         "coingecko_id": "secret"
       },
+      {
+        "description": "The native token cw20 for SIENNA on Secret Network",
+        "denom_units": [
+          {
+            "denom": "cw20:secret1rgm2m5t530tdzyd99775n6vzumxa5luxcllml4",
+            "exponent": 18
```
mradkov, line 32: To reviewers: Shall this be `18` to account for decimals in both directions?

I see other SNIP-20 token (SHD) missed this and it results in this weird conversion rate in the UI (see pool 846):

signal-2022-11-08-165936_002

PR #991, opened 2022-11-09
chore: Update cheqd explorers

ankurdotb:


cheqd/chain.json
```diff
@@ -151,26 +151,36 @@
...
-      "kind": "atomscan.com",
-      "url": "https://atomscan.com/cheqd",
-      "tx_page": "https://atomscan.com/cheqd/transactions/${txHash}",
-      "account_page": "https://atomscan.com/cheqd/accounts/${accountAddress}"
+      "kind": "EZStaking Tools",
+      "url": "https://ezstaking.tools/cheqd",
+      "tx_page": "https://ezstaking.tools/cheqd/txs/${txHash}"
+    },
+    {
+      "kind": "ping-pub",
```
JeremyParish69, line 176: We should probably change this to look.chill, even though it's 'inspired by' ping.pub...
same with the following explorer--should have a different provider

ankurdotb: Good question. I think it's useful to have **both** a `kind` property as well as `provider` property for block explorers. `kind` tells me what the capabilities of the block explorer are, e.g., the featureset on ping.pub vs BigDipper is wildly different. Similarly, we might run a BigDipper instance at explorer.cheqd.io and another might be run at cheqd.bigdipper.live (hypothetically).

The other issue is that unless I've entirely missed it, there's no standardised taxonomy for explorer types. I've called it `ping-pub` based on other chains, but searching through the entire repo now, there's `Ping`, `ping.pub`, `ping-pub` - all nominally referring to the same thing. That causes a lot of confusion.

What would be the best way to have a limited set of block explorer "kinds", @JeremyParish69? or any thoughts on above (can suggest a PR for chain schema if you think it's a good idea).

JeremyParish69: @ankurdotb I appreciate this kind of thinking.

You are correct that we are currently lacking clear rules on how explorers should be defined. My blind-eye to it so far has been because I wanted to drastically restructure how they're defined and first prove the concept in other areas(, and potentially just move it to another registry since it's a third party service and not chain metadata).

The issue I see with using 'kind' for ping pub clones is that they could potentially be modified to something unrecognizable, and so it'd have to carry a somewhat blurred definition.

In the meantime, I think some incorrect data for 'kind' isn't too problematic, because it's solely a descriptor at the moment. But, eventually, we'll want clearer rules for it. The less accurate it becomes before then, the more work will be required later on, but it's not the end of the world either way. For now, I'll just ask you to pick which you think is better (which could be as you have it now) and we can just push that and revisit later.

ankurdotb: @JeremyParish69 Okay cool, I was thinking of `kind` as direct or compatible clones. What would you recommend as the solution for the current PR then? For the ping pub clones, should I change them to the provider name?

JeremyParish69: @ankurdotb My preference for now would be provider name and not 'clone category', but I don't think we need to stress it too much right now. I think this deserves closer inspection later but for now I can be lenient.

PR #994, opened 2022-11-09
Fix Exp. Link

Nodeist:


stride/chain.json
```diff
@@ -245,8 +245,8 @@
     {
       "kind": "Nodeist Explorer",
       "url": "https://exp.nodeist.net/Stride",
```
JeremyParish69, line 247: Can we update this account page too, please? It also just needs the 'M-'

Nodeist: oh sorry ok!

JeremyParish69: Why close the PR? it just needed a simple change

PR #1015, opened 2022-11-17
update persistence versions and coin-type.

puneet2019:
Persistence moved to use 118 coin-type
persistence will use v5.0.0 from 24th Nov. 
removed persistence owned explorer.

### DIFF
--- a/persistence/chain.json
+++ b/persistence/chain.json
@@ -12,7 +12,7 @@
   "key_algos": [
     "secp256k1"
   ],
-  "slip44": 750,
+  "slip44": 118,
   "fees": {
     "fee_tokens": [
       {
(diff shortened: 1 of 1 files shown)


persistence/chain.json
```diff
@@ -12,7 +12,7 @@
   "key_algos": [
     "secp256k1"
   ],
-  "slip44": 750,
```
JeremyParish69, line 15: Still registered as 750 in https://github.com/satoshilabs/slips/blob/master/slip-0044.md. Is this planned to be removed from that list?

puneet2019: it is planned, we have changed it in code to 118, but we currently still support 750 so that migration happens, wondering when is the good time to change the slip

PR #1022, opened 2022-11-18
fix stargaze testnet chain id

0xArbi:
Check the ID yourself at https://testnet-explorer.publicawesome.dev/stargaze


JeremyParish69:
ah, yes, apparently a star in the Virgo constallation--makes sense for Stargaze
https://en.wikipedia.org/wiki/Phi_Virginis

PR #1023, opened 2022-11-18
Add kjnodes seed node for Teritori

kj89:
Have added port for the seed node


dylanschultzie:
Thanks for fixing, @kj89 !

PR #1039, opened 2022-11-25
Add CoolCat testnet chain

vexxvakan:
🐈


testnets/coolcattestnet/chain.json
```diff
@@ -0,0 +1,63 @@
+{
+  "$schema": "../../chain.schema.json",
+  "chain_name": "coolcat",
```
JeremyParish69, line 3: this should be 'coolcattestnet'

testnets/coolcattestnet/assetlist.json
```diff
@@ -0,0 +1,29 @@
+{
+  "$schema": "../../assetlist.schema.json",
+  "chain_name": "coolcat",
```
JeremyParish69, line 3: this should be 'coolcattestnet'

PR #1047, opened 2022-11-29
Add planq chain

ivanshukhov:


_IBC/planq-cosmoshub.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
```
JeremyParish69, line 3: Very nice submission. The only issue I can spot right away is the fact that you have the 'planq' name come first in these _IBC files, but in all these 3 cases, 'p' alphabetically comes after 'c', 'g', and 'o', and thus should be written second in the title and should be described by 'chain 2'.

ivanshukhov: Oh, sorry this went totally past me.

PR #1058, opened 2022-12-04
updated bze coin details and added logo svg

faneaatiku:


beezee/assetlist.json
```diff
@@ -20,8 +20,14 @@
       "name": "BeeZee",
       "display": "bze",
       "symbol": "BZE",
+      "ibc": {
+        "source_channel": "channel-0",
+        "dst_channel": "channel-340",
```
JeremyParish69, line 25: which chain uses channel 340? This info should be defined in the \_IBC\ folder, not here

PR #1067, opened 2022-12-06
Update peer,rest,rpc Oraichain provided by Vchain

thanhnv2303:


oraichain/chain.json
```diff
@@ -92,6 +92,11 @@
         "id": "d091cabe3584cb32043cc0c9199b0c7a5b68ddcb",
         "address": "seed.orai.synergynodes.com:26656",
         "provider": "synergynodes"
+      },
+      {
+      "id:"8b346750e75fd584645192a65c62c7ab88741791",
```
JeremyParish69, line 97: missing the " quote to close the property name "id"
should be "id": "8b...91",

thanhnv2303: I have fixed this issue, please recheck it.

JeremyParish69: sorry, it didn't work. You added a space, but not the needed missing quote

thanhnv2303: Sorry about my mistake. Please recheck.

PR #1068, opened 2022-12-07
Add Mises Network

OnepercentTech:


mises/chain.json
```diff
@@ -0,0 +1,127 @@
...
+  "website": "https://mises.site",
+  "pretty_name": "Mises",
+  "chain_id": "mainnet",
+  "bech32_prefix": "mis",
+  "node_home": "$HOME/.misestm",
+  "daemon_name": "misestmd",
+  "key_algos": [
+    "secp256k1"
+  ],
+  "slip44": [60, 118],
```
JeremyParish69, line 15: the type supports only a number. Are both coinTypes are supported?

mises/chain.json
```diff
@@ -0,0 +1,127 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "mises",
+  "status": "live",
+  "network_type": "mainnet",
+  "website": "https://mises.site",
+  "pretty_name": "Mises",
+  "chain_id": "mainnet",
+  "bech32_prefix": "mis",
```
JeremyParish69, line 9: Please register this prefix

mises/chain.json
jambulmerah: Yes, both are supported

mises/chain.json
jambulmerah: Registered and fixed typo from `mis` to `mises`
https://github.com/satoshilabs/slips/commit/1cbb46e974d306d512f16dbe6cfa23dffa86990b
https://github.com/cosmos/chain-registry/pull/1068/commits/94a6b5ca29dacaa998da5aa611f3342e97a3e7e9

PR #1083, opened 2022-12-08
Upgrade Impacthub from chain impacthub-3 to ixo-4

dylanschultzie:


dylanschultzie:
Peers and seeds will need updated, but as an initial pass, this should maintain clarity.

alwyn-ixo (approved):

PR #1086, opened 2022-12-08
Upgrade Desmos to v4

dylanschultzie:
https://www.mintscan.io/desmos/proposals/19


JeremyParish69:
"height 6,421,610 which is expected to be proposed around Thursday 15th December at 15:00 UTC."

PR #1094, opened 2022-12-10
Add Secret Network token: AMBER 

kent-3:


secretnetwork/assetlist.json
```diff
@@ -133,6 +133,27 @@
           "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/secretnetwork/images/alter.svg"
         },
         "coingecko_id": "alter"
+      },
+      {
+        "description": "The native token cw20 for Amber on Secret Network",
```
JeremyParish69, line 138: Can we also include please type_asset snip20, and tyhe contract address field?

kent-3: Sure thing! I have just added them.

PR #1100, opened 2022-12-12
Chain-id change from impacthub-3 to ixo-4

alwyn-ixo:


JeremyParish69:
Seems we already have ixo registered under the 'impacthub' folder, already with the new chainId. Changing the folder could cause issues with dependent applications.

alwyn-ixo:
@JeremyParish69 
Thx, Jeremy.
We had to make a similar change on chainapsis/keplr's repo which resolved that issue.
I'm worried that dependent applications may require this new directory structure, but I'm not sure.
Your thoughts?

JeremyParish69:
Thanks @alwyn-ixo .
I think that, although the name 'ixo' makes more sense to me for this folder, renaming the folder would be a breaking change to all the apps that point to the chain registry as a dependency, thus I'm inclined to avoid it unless absolutely necessary.

JeremyParish69:
At this point, I'm going to assume that changing the folder name isn't too important for the ixo team. If really desired, feel free to let me know, @alwyn-ixo , as we might be able to squeeze such a change easily into an upcoming update to the Chain Registry. For now, I'll close.

PR #1101, opened 2022-12-12
chore: update Injective info

achilleas-kal:


achilleas-kal:
@JeremyParish69 thanks :)

PR #1130, opened 2022-12-19
Update chaininfo and add to IBC

nodebreaker0-0:


crescent/assetlist.json
```diff
@@ -31,7 +31,7 @@
       "denom_units": [
         {
           "denom": "ubcre",
-          "exponent": 0,
+          "exponent": 6,
```
JeremyParish69, line 34: This should remain at 0, as all base denoms do. bcre will have exponent 6, but ubcre should have 0.

_IBC/crescent-evmos.json
```diff
@@ -2,22 +2,22 @@
     "$schema": "../ibc_data.schema.json",
     "chain_1": {
       "chain_name": "crescent",
-      "client_id": "07-tendermint-24",
-      "connection_id": "connection-27"
+      "client_id": "07-tendermint-12",
```
JeremyParish69, line 5: why is this an edit to the crescent-stargaze file? Shouldn't this be a new file?

Heheehd: Jejeje

nodebreaker0-0:
Also, the crescent core version has been upgraded, and Terra2 has been newly added.

JeremyParish69:
@nodebreaker0-0 Please restore the crescent-stargaze.json file this deletes.

nodebreaker0-0:
@JeremyParish69 The ibc channel between Crescent and Stargaze is currently unmanaged and has been expired, which can cause confusion. And yet, do we have to restore it?

JeremyParish69:
> @JeremyParish69 The ibc channel between Crescent and Stargaze is currently unmanaged and has been expired, which can cause confusion. And yet, do we have to restore it?

You don't have to restore the connection, but we shouldn't delete that file. If a new connection between the two chains is someday created, then the case for replacement can be examined.

nodebreaker0-0:
@JeremyParish69 Restored Stargaze, added Juno. Thank you for your feedback.

PR #1163, opened 2022-12-30
Add bech32 config to chain schema

### LINKED ISSUE #665: Add bech32 HRP prefixes to chain.json
We should add some data around HRP prefixes (e.g. `osmo1...`, `cosmos1...`) in the address formats to the chain registry. Not sure how it would best look / be formatted

JeremyParish69:
-adds bech32_config object, with 6 optional properties
-bech32_prefix is still required, and the config is for override.

closes #665

### DIFF
--- a/chain.schema.json
+++ b/chain.schema.json
@@ -47,13 +47,12 @@
       ]
     },
     "bech32_prefix": {
-      "type": "string"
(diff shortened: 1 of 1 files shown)


ValarDragon (approved):
Nice!

PR #1183, opened 2023-01-03
updates chain-data for quicksilver

ajansari95:
- Updated registry for quicksilver-2


JeremyParish69 (changes requested):
Thanks for the submission.
I think we should avoid the name 'qck' for the rather wide chain logo, because 'qck' is already used for the token logo. Can we please rename this?
Also, the 'quicksilver-chain-logo.svg' (without text) would be a valuable svg to have for the token logo, but for a token logo, the file should have square dimension, but this image file is much wider than tall.
Also, also, the chain.json does allow for specifying a chain logo--if you happen to mind specifying the new chain logo image, that would be appreciated.

PR #1205, opened 2023-01-09
Update Kujira Chain Logo

bretheren:
Minor update to Kujira Chain Logo


JeremyParish69:
We do already have this as kuji.png, and is used for the token logo. However the chain's logo points to 'kujira-chain-logo' and is the one with the whale. I still the the whale image used in various places such as the Medium page. Is the image of the whale completely discontinued in favor of using the token logo was the chain logo as well?

bretheren:
> We do already have this as kuji.png, and is used for the token logo. However the chain's logo points to 'kujira-chain-logo' and is the one with the whale. I still the the whale image used in various places such as the Medium page. Is the image of the whale completely discontinued in favor of using the token logo was the chain logo as well?

Yes, Kujira is deprecating the logo with the whale and replacing it with the same logo as the KUJI token. It is replaced in certain places(like fav icons on the website etc) and will continue to change others, like medium, in time

PR #1225, opened 2023-01-13
Inj scrt ibc

PFC-developer:
update the Injective/Secret IBC channel (one in there was expired)


JeremyParish69:
Do we have proof that no coins were transferred on the old channel? Alternatively, (and preferred) we can use proof of governance on both chains voting to change the ibc connection. Otherwise users will be left with non-fungible ibc coins by switching channels, and re-enabling the expired channel is preferred.

PFC-developer:
This happened months ago. 
at the time, I think these were checked, and decided that this was ok.
Feel free to contact the SCRT or INJ maintainers to verify

JeremyParish69:
You're right. I could verify that the old channel is abandoned with no activity for ~9 months. Meanwhile, this new one has been around a while and is very active. I take this activity as proof of canon and not going to be disruptive. Thanks for the submission.

PFC-developer:
🙏

PR #1240, opened 2023-01-16
Update README.md

liangping:


JeremyParish69:
Doesn't feel quite right to me to not make this required by schema, but just write down that it's required. We have other required properties that me don't need to write out a list of in the readme. I've suggested to move the property up higher in the list of properties to help teams see the property when registering. https://github.com/cosmos/chain-registry/pull/1272

JeremyParish69:
Closing due to inactivity. We can discuss further on TG if needed.

liangping:
> Doesn't feel quite right to me to not make this required by schema, but just write down that it's required. We have other required properties that me don't need to write out a list of in the readme. I've suggested to move the property up higher in the list of properties to help teams see the property when registering. #1272

Looks good to me.

Sorry for late replying, I was on Chinese New Year vocation.

PR #1269, opened 2023-01-20
Add Allnodes

3eph1r0th:


akash/chain.json
```diff
@@ -79,6 +79,16 @@
       {
         "id": "47f7b7a021497ad7a338ea041f19a1a11ae06795",
         "address": "rpc.la.akash.farm:26656"
+      },
+      {
+        "id": "e1b058e5cfa2b836ddaa496b10911da62dcf182e",
+        "address": "akash-seed-1.allnodes.me:26656"
```
JeremyParish69, line 85: Comma missing after 'address' in every file

3eph1r0th: Thank you. Should be fixed now

cryptoorgchain/chain.json
```diff
@@ -1,4 +1,4 @@
-{
+м{
```
JeremyParish69, line 1: this M here is causing the failure

3eph1r0th: Thank you. Fixed as well

PR #1273, opened 2023-01-21
Add Mars protocol

larry0x:
TODO:

- [x] confirm seed node id and port
- [x] ~coingecko id~ CoinGecko ID will only be available much later, so we leave it empty for this PR.


mars/assetlist.json
```diff
@@ -0,0 +1,29 @@
...
+      "denom_units": [
+        {
+          "denom": "umars",
+          "exponent": 0,
+          "aliases": ["micromars"]
+        },
+        {
+          "denom": "mars",
+          "exponent": 6,
+          "aliases": ["mars"]
```
JeremyParish69, line 16: Worried about including an alias that is identical to the denom value. Can we remove the aliases array for this denom?

larry0x: Good point. I removed the aliases (removed “micromars” as well, as I don’t think anyone would use it in practice.)

PR #1294, opened 2023-01-26
add chimba network

camiloss1:


chimba/chain.json
```diff
@@ -0,0 +1,64 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "chimba",
+  "status": "live",
+  "network_type": "mainnet",
+  "website": "https://chimba.ooo/",
+  "pretty_name": "Chimba",
+  "chain_id": "chimba",
+  "bech32_prefix": "chimba",
```
JeremyParish69, line 9: Please correct the SLIP-173 registration at https://github.com/satoshilabs/slips/blob/master/slip-0173.md . It has 'cmba' as the bech32 prefix, when in fact the prefix is 'chimba'.

chimba/chain.json
```diff
@@ -0,0 +1,64 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "chimba",
+  "status": "live",
+  "network_type": "mainnet",
+  "website": "https://chimba.ooo/",
+  "pretty_name": "Chimba",
+  "chain_id": "chimba",
```
JeremyParish69, line 8: Looks like the chain-id might be chimbanode-1, as per what I see on the explorer.

chimba/chain.json
camiloss1: the change in the file for slip-073 has already been applied

chimba/chain.json
camiloss1: chimbanode-1 is the name for moniker but the chain-id is chimba

PR #1308, opened 2023-01-28
update junod version - testnet

doggystylez:
pending testnet relaunch on 2022-01-30


doggystylez:
closing for now.. will probably relaunch with a new chain_id

PR #1342, opened 2023-02-02
Add nride cw20 token in JUNO assetlist

arrivets:


juno/assetlist.json
```diff
@@ -642,6 +642,29 @@
...
+            "exponent": 0
+          },
+          {
+            "denom": "nride",
+            "exponent": 6
+          }
+        ],
+        "base": "cw20:juno1qmlchtmjpvu0cr7u0tad2pq8838h6farrrjzp39eqa9xswg7teussrswlq",
+        "name": "nRide Token",
+        "display": "nRide Token",
```
JeremyParish69, line 662: display denom should be 'nride'

PR #1358, opened 2023-02-03
Change pretty_name of Mars to Mars Hub

sunnya97:


JeremyParish69:
Change already added. Looks like we had the same idea around the same time: https://github.com/cosmos/chain-registry/pull/1359

PR #1362, opened 2023-02-04
Update Mars CoinGecko ID and explorer

larry0x:
- Add Mars token coingecko id
- Add MintScan and ping.pub explorer info
- Add Juno<>Mars IBC connection

_**The coingecko id hasn't been confirmed yet, but 99% chance it will be `mars-protocol`. We will mark this PR as ready once it's confirmed.**_


joeabbey (approved):

larry0x:
> Somehow it go changed to {"id":"mars-protocol-a7fcbcfb-fd61-4017-92f0-7ee9f9cc6da3","symbol":"mars","name":"Mars Protocol"}, but I assume this is not intentional and will change back.

It is a temporary id for new listings. We're expecting it to change to `mars-protocol` some time soon.

larry0x:
Coingecko still hasn't updated our token id. So for now let's go with `mars-protocol-a7fcbcfb-fd61-4017-92f0-7ee9f9cc6da3`

PR #1371, opened 2023-02-06
Update nyx chain logo 

mfahampshire:


mfahampshire:
Ah, I misunderstood the use of the image then. I will close this PR as I want to keep the token logo the same.

PR #1404, opened 2023-02-11
Update chain.json

takeshi-val:


chain4energy/chain.json
```diff
@@ -70,6 +70,11 @@
         "address": "c4e.seed.node75.org:10156",
         "provider": "Pro-nodes75"
       }
```
JeremyParish69, line 72: to add a list element, you need a comma here

PR #1419, opened 2023-02-15
Update quasar to testnet v04

pabloqsr:


testnets/quasartestnet/chain.json
```diff
@@ -20,16 +20,16 @@
     ]
   },
   "codebase": {
-    "recommended_version": "v0.0.1-alpha-6",
+    "recommended_version": " v0.0.2-alpha-11",
     "compatible_versions": [
-      "v0.0.1-alpha-6"
+      " v0.0.2-alpha-11"
```
JeremyParish69, line 25: I assume the new version isn't supposed to have an initial space? seems like that might cause issue for some...

pabloqsr: Oh, my bad, good catch.
Fixed!

PR #1436, opened 2023-02-16
revert mars-protocol coingecko-id

PFC-developer:
the longer string works, the shorter one shows a price, but it is incorrect. If you would like to verify it is displayed - https://www.coingecko.com/en/coins/mars-protocol


larry0x:
Turns out mars-protocol is the old terra classic mars token

PR #1454, opened 2023-02-19
added qatom as asset on quicksilver

ajansari95:
- added qatom as an asset and related logos


JeremyParish69:
*Note that the large margin on the image might make it appear small in some displays.

PR #1467, opened 2023-02-20
Update persistence-quicksilver.json

napelvs:
The previous IBC Channel between the chains has expired. Since I was the only operator on the channel, I decided to create another one instead of resurrecting the old one through a governance proposal.

### DIFF
--- a/_IBC/persistence-quicksilver.json
+++ b/_IBC/persistence-quicksilver.json
@@ -2,22 +2,22 @@
     "$schema": "../ibc_data.schema.json",
     "chain_1": {
         "chain_name": "persistence",
-        "client_id": "07-tendermint-93",
-        "connection_id": "connection-73"
+        "client_id": "07-tendermint-96",
+        "connection_id": "connection-77"
     },
     "chain_2": {
         "chain_name": "quicksilver",
-        "client_id": "07-tendermint-9",
-        "connection_id": "connection-5"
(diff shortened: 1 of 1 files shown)


JeremyParish69:
Are there governance proposals on both sides showing agreement to switching to a new channel? Alternatively, if no tokens are stuck on their non-native chain it should be OK.

napelvs:
The other channel expired. So we've opened another one.

JeremyParish69:
We can't switch to new channels without compromising the integrity to tokens that had already been transferred. If we can neither show that there are no abandoned tokens from the old channel, nor that governance on both chains has approved switching to a new channel, we cannot accept this new channel as 'canon' and therefore will not register it.

napelvs:
So the channel was only used by me and then expired. The only assets on that path are mine and are close to $0 (0.00020 XPRT on the Persistence side, and 0.00030 XPRT on the Quicksilver side).

You can verify the address is mine by looking at the memo of the "IBC Acknowledgement" and/or "IBC Received" TXs.

My Quicksilver Wallet: https://www.mintscan.io/quicksilver/account/quick15rwjpdn2cjunek9g0sqmehwd4ht3nz0fkzx673
My Persistence Wallet: https://www.mintscan.io/persistence/account/persistence1m3hsmmf633nl5dl8sm58dypu4fthppe7ej463y

Issue #462, opened 2022-06-29
Allow for image variants?


JeremyParish69:
For wherever the Chain Registry allows for defining images, we just have a simple 'logo_URIs' object with a png and svg field.

We may want to consider generalizing the name to 'images' instead of logo_URIs?
Are there any other useful file types than png and svg? ...perhaps it's best to restrict it to png and svg
Most chains seems to have multiple variants of their token logo, due to different themes and changes over time.
Here are some properties I think could be used. E.g.,
  - Shape. Some dapps might allow for square token logos, while other might insist on circular logos (and might crop a square logo, which can look bad). If there is only a circular logo, and no variants based on share, then simply omit the shape field. Could also eventually allow '3d', which perhaps could be animated on a UI.
  - Mode. There can be light and dark mode variants. This can greatly improve UX for adapting to a dark mode.
  - Color.
  - Background (color).
  - Themes. Might want to allow for themes on a dApp, like 'red', or 'Christmas', or 'pixelated', or 'old' vs 'new'.
  
Image object:
```
"images": [
  {
    "shape": "circle",
    "mode": "light",
    "color": "red",
    "background": "white",
    "themes": ["red"],
    "png": "http://token_logo_circle.png",
    "svg": "http://token_logo_circle.svg"
  },
  {
    "shape": "square",
    "mode": "light",
    "color": "red",
    "background": "white",
    "themes": ["red"],
    "png": "http://token_logo.png",
    "svg": "http://token_logo.svg"
  },
  {
    "shape": "circle",
    "mode": "dark",
    "color": "red",
    "background": "black",
    "themes": ["red"],
    "png": "http://token_logo_circle_dark.png",
    "svg": "http://token_logo_circle_dark.svg"
  },
  {
    "shape": "square",
    "mode": "dark",
    "color": "red",
    "background": "black",
    "themes": ["red"],
    "png": "http://token_logo_dark.png",
    "svg": "http://token_logo_dark.svg"
  }
]
```

Alternatively, if the properties feel too much like overkill, there could still be use in just allowing for non-default variants:
```
"images": [
  {
    "png": "http://token_logo_circle.png",
    "svg": "http://token_logo_circle.svg"
  },
  {
    "png": "http://token_logo.png",
    "svg": "http://token_logo.svg"
  },
  {
    "png": "http://token_logo_circle_dark.png",
    "svg": "http://token_logo_circle_dark.svg"
  },
  {
    "png": "http://token_logo_dark.png",
[...]

PR #1480, opened 2023-02-21
Add Images Array

JeremyParish69:
Each asset and chain will eventually instead use an array of images (but for now it will just be added--not removing the old logo_URIs just yet), and each 'image' object in the array can have 'svg', 'png', and 'theme', where theme contains (for now) 'primary_color_hex'. Theme can be expanded later.
Meant to address issue : https://github.com/cosmos/chain-registry/issues/462

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,9 +136,16 @@ A sample `chain.json` includes the following information.
       }
     ]
   },
+  "images": [
+    {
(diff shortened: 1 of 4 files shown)


ValarDragon (approved):
LGTM for now, but can we add a note to ourself to look for if theres any standards for this more generally?

I feel like this is somewhat equivalent to presets in a "theme selector" menu

PR #1529, opened 2023-03-01
Full update of the chain-registry entries

Galadrin:
The chain-registry is supposed to contain updated informations about chains.

But, with the time, the original peers and some community provided ressources are stopped or moved.
It's time to cleanup the registry and say goodbye to the non up to date info.

I made some chains by hand then wrote a script to do it automatically.

script source: https://gist.github.com/Galadrin/1f017c81d32514f6f8938f3a19fd6c41

- The seeds and peers are tested for connection and handshakes reception, not for peer-id and chain-id 
- The HTTP endpoints have 2 seconds to accept connection and 2 more seconds to send the request result.
it's only very simple requests that must be answered very fast.
- The gRPC endpoints are check for readyness, not for methods availability.


jonator:
Would be interesting if we could use this script behind a GH action to produce warnings about node liveness.

Galadrin:
agree, but maybe in a second time, the script will need to be modified to take only the modified json in consideration.

PR #1531, opened 2023-03-01
add orachain endpoint

NodineID:


oraichain/chain.json
```diff
@@ -97,6 +102,11 @@
       "id": "8b346750e75fd584645192a65c62c7ab88741791",
       "address": "peer-oraichain.vchain.zone",
       "provider": "Vchain-team"
+      },
+      {
+        "id": "cdcbeccdbbab954850e5ce629239c58210a5aca2",
+        "address": "sseed-orai.nodine.id:21656",
```
JeremyParish69, line 108: with 2 's's in 'seed', is this a typo? Also, same address used for seed and persistent peer?

NodineID: yes that is type. and also only persistent peer. im sorry for making you confused

JeremyParish69: Ok, if that is error, can you please correct the mistake?

NodineID: just fixed, thanks

NodineID:
Just make new commit last day

PR #1620, opened 2023-03-16
Add agoric logos

JeremyParish69:


dckc:
Thanks for adding these -- especially `inter.svg`

fwiw, I confirmed with Inter Protocol marketing folks that these logos are consistent with the [Inter Media Kit](https://inter.trade/media-kit/).

cc @0xpatrickdev

PR #1630, opened 2023-03-18
add arkh logo

kynraze:


arkh/assetlist.json
```diff
@@ -5,21 +5,17 @@
...
-            "denom": "uarkh",
-            "exponent": 0
-          },
           {
             "denom": "arkh",
             "exponent": 6
           }
         ],
-        "base": "uarkh",
+        "base": "arkh",
```
JeremyParish69, line 13: We need to keep some denom unit of exp 0, and base must point to that.

arkh/chain.json
```diff
@@ -16,7 +16,7 @@
   "fees": {
     "fee_tokens": [
       {
-        "denom": "uarkh",
+        "denom": "arkh",
```
JeremyParish69, line 19: this should go back to uarkh as well

arkh/chain.json
```diff
@@ -3,7 +3,7 @@
   "chain_name": "arkh",
   "status": "live",
   "network_type": "mainnet",
-  "website": "https://arkh.com/",
+  "website": "http://arkhadian.com/",
```
JeremyParish69, line 6: Site doesn't work? It the domain reserved?

kynraze: nah, the site is currently down But the previous site has no relation to this chain

PR #1636, opened 2023-03-19
fix(kujiraIBC): Added Axelar, Cosmoshub, and Gravitybridge info

ToggLeTek:


JeremyParish69:
We require these to be in alphabetical order, so 'kujira' must be chain 2 for all of these, since 'k' comes after 'g', 'a', and 'c'.

ToggLeTek:
I literally just caught that, and the newest version should be corrected.

PR #1641, opened 2023-03-20
Added SVG version of MNTL Logo for AssetMantle

kombos:
Added SVG version of MNTL Logo for AssetMantle


kombos:
will resubmit with a better SVG file.

PR #1645, opened 2023-03-20
add arkh-osmosis

kynraze:


JeremyParish69:
If we could remove that black background (or make it circular, if it's supposed to be a black background) that would be superb.

PR #1652, opened 2023-03-21
update umeed version

doggystylez:
see https://www.mintscan.io/umee/proposals/63


JeremyParish69:
Are we able to add the gas pricing updates as well?

doggystylez:
sure thing!

PR #1667, opened 2023-03-22
add ampluna

0xphilipp:
This PR adds ampluna and the ICS20-CW20 channel between Terra2 and Carbon.

Is there a simple way to add:
juno-terra2
kujira-terra2


_IBC/osmosis-terra2.json
```diff
@@ -44,6 +44,23 @@
...
       }
+    },
+    {
+      "chain_1": {
+        "channel_id": "channel-341",
+        "port_id": "transfer"
+      },
+      "chain_2": {
+        "channel_id": "channel-26",
+        "port_id": "wasm.terra1e0mrzy8077druuu42vs0hu7ugguade0cj65dgtauyaw4gsl4kv0qtdf2au"
```
JeremyParish69, line 55: Why did you add a contract for Osmosis? There already is an CW20<>ICS20 contract on Terra2 for sending to Osmosis....

0xphilipp: The issue is that we have 2 ICS20 contracts on terra and both are use. The terra1e0mrzy8077druuu42vs0hu7ugguade0cj65dgtauyaw4gsl4kv0qtdf2au was the first one and is used to transfer ampLUNA to other ecosystems and has ~70k USD value locked, while the other existing one was created afterward and has ~500 USD locked.

What is the recommended way to resolve that issue? I understand that some tools might not work best if we have multiple ics20 contracts on a single chain.

JeremyParish69: Thanks for the info, we'll get this sorted. If these number are accurate, we may want to find a way to revert on the smaller contract.

JeremyParish69: For the time being, there is no issue referring to this contract for other Terra2 connections, but for the Osmosis-Terra2 connection, we are still working on sorting this out.

JeremyParish69: Would it be possible to restrict osmosis to only the existing contract? We should be able to use 2 different contracts as long as each ibc connection sticks to just one defined cw20<>ics20 contract. I'm assume that 70k on this contract you're introduction is NOT to Osmosis, but to other networks.

0xphilipp: It is to other networks, but I expect issues if we start using different contracts for different networks. We do plan to launch on osmosis and have moved some funds over but did not list the pair yet.

There should not be different ICS20 contracts for the same asset to different chains. This will get messy. My recommendation would be to unwind the existing contract.

1. The astro liquidity is less than 500 $ worth, of which most is transferred to kujira
2. Astro in it's own is an ics20 transferrable contract and when looking at the transfer of astro from terra to injective, the own astro token contract is used and not the existing ics20 contract.

PR #1670, opened 2023-03-22
Add Migaloo assets

JeremyParish69:
ampwhale, bwhale, fable


0xFable:
Found this PR after checking other one. This is looking good! What is needed for merge? I could close the other two in favour of this

JeremyParish69:
> Found this PR after checking other one. This is looking good! What is needed for merge? I could close the other two in favour of this

Just wanted your 'OK' to signal that I can merge. I noticed that the two fable images are slightly different (one has a body, and the other is just a head);should I get rid of the png version--the one with the body?

0xFable:
Ahh, my apologies didnt realise. Yes all good to go on this. During the changes for fable image it got changed into something more akin to a coin. Yes the version with the body could go its no longer needed. 
Might seek you out on TG or something appreciate all the effort you've put in the help the PR

PR #1691, opened 2023-03-23
Update Quasar to mainnet

dylanschultzie:
@pabloqsr please review

### DIFF
--- a/quasar/chain.json
+++ b/quasar/chain.json
@@ -33,7 +33,7 @@
       {
         "denom": "ibc/FA7775734CC73176B7425910DE001A1D2AD9B6D9E93129A5D0750EAD13E4E63A",
         "fixed_min_gas_price": 0.01
-      },
+      }
     ]
   },
   "codebase": {
--- a/quasar/chain.json
+++ b/quasar/chain.json
@@ -56,6 +56,11 @@
         "id": "20e1000e88125698264454a884812746c2eb4807",
         "address": "seeds.lavenderfive.com:18256",
         "provider": "Lavender.Five Nodes 🐝"
+      },
+      {
+        "id": "ade4d8bc8cbe014af6ebdf3cb7b1e9ad36f412c0",
+        "address": "seeds.polkachu.com:18256",
+        "provider": "Polkachu"
       }
     ],
     "persistent_peers": [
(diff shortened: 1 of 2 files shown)


quasar/chain.json
```diff
@@ -19,128 +20,71 @@
       }
     ]
   },
+  "fees": {
+    "fee_tokens": [
+      {
+        "denom": "ibc/0471F1C4E7AFD3F07702BEF6DC365268D64570F7C1FDC98EA6098DD6DE59817B",
```
JeremyParish69, line 26: I don't know what these three tokens are, but could we register them as ics20 assets and add register their corresponding ibc channels?

quasar/chain.json
```diff
@@ -19,128 +20,71 @@
...
+        "fixed_min_gas_price": 0.01
+      },
+      {
+        "denom": "ibc/FA0006F056DB6719B8C16C551FC392B62F5729978FC0B125AC9A432DBB2AA1A5",
+        "fixed_min_gas_price": 0.01
+      },
+      {
+        "denom": "ibc/FA7775734CC73176B7425910DE001A1D2AD9B6D9E93129A5D0750EAD13E4E63A",
```
dylanschultzie, line 34: axlUSDC (from osmosis?)

quasar/chain.json
dylanschultzie, line 26: OSMO

quasar/chain.json
```diff
@@ -19,128 +20,71 @@
...
     ]
   },
+  "fees": {
+    "fee_tokens": [
+      {
+        "denom": "ibc/0471F1C4E7AFD3F07702BEF6DC365268D64570F7C1FDC98EA6098DD6DE59817B",
+        "fixed_min_gas_price": 0.01
+      },
+      {
+        "denom": "ibc/FA0006F056DB6719B8C16C551FC392B62F5729978FC0B125AC9A432DBB2AA1A5",
```
dylanschultzie, line 30: ATOM

quasar/chain.json
pabloqsr: Hey @JeremyParish69 
Chain is only accepting those denoms for gas. Native token will be launch after a few weeks.
How do we register them as ics20 assets?

We will be also registering the corresponding IBC channels soon when we visit

dylanschultzie: @pabloqsr we ended up discussing adding the ICS tokens off-line; you can see them in the assets file. No more is needed here at this time

PR #1710, opened 2023-03-25
quicksilver: RPC cleanup

blacklodge-admin:
I don't know exactly how the cosmos.directory RPC proxy works (would love to know, if you care to share!) but I suspect it handles the last two, if it makes more sense to leave them in the config.  For whatever reason, though, the proxy seems to consistently route to the first endpoint below, and because that endpoint requires 25x higher fees than the chain itself, any dApp with hard-coded reliance on the proxy endpoint (e.g. restake.app) deterministically fails.  Apologies for my ignorance if my understanding is somehow off base!

This PR...
- removes the `lesnik | UTSA` RPC endpoint (fee settings appear to be misaligned, causing tx failure)
- removes the `Nodeist` RPC endpoint (all API calls timeout)
- removes the `Forbole` RPC endpoint (the binary, at least, has no idea what to do with the `wss://` scheme)

### DIFF
--- a/quicksilver/chain.json
+++ b/quicksilver/chain.json
@@ -104,6 +104,14 @@
         "provider": "Lavender.Five Nodes 🐝"
       },
       {
+        "address": "https://m-quicksilver.rpc.utsa.tech",
+        "provider": "lesnik | UTSA"
+      },
+      {
+        "address": "https://quicksilver-mainnet-rpc.autostake.com:443",
+        "provider": "AutoStake 🛡️ Slash Protected"
+      },
+      {
         "address": "https://quicksilver-rpc.ibs.team:443",
         "provider": "Inter Blockchain Services"
       }
--- a/quicksilver/chain.json
+++ b/quicksilver/chain.json
@@ -132,10 +132,6 @@
       {
         "address": "https://m-quicksilver.api.utsa.tech",
         "provider": "lesnik | UTSA"
-      },
-      {
-        "address": "https://api-quicksilver.nodeist.net",
-        "provider": "Nodeist"
       }
     ],
     "grpc": [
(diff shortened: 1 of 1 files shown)


JeremyParish69:
hmmm... I'm not sure about removing the lesnik one because I don't think there's any standard here on whether endpoints with misalignment should be removed. IMO, endpoints with relatively fees issues (assuming otherwise online) should merely be flagged to have that issue, thereby allowing proxy endpoints to rotate solely among the nodes without any issues flaged.

blacklodge-admin:
> hmmm... I'm not sure about removing the lesnik one because I don't think there's any standard here on whether endpoints with misalignment should be removed. IMO, endpoints with relatively few issues (assuming otherwise online) should merely be flagged to have that issue, thereby allowing proxy endpoints to rotate solely among the nodes without any issues flaged.

Looking at the schema, it doesn't seem like there's any property to do this sort of flagging with today.  Is that accurate?

lesnikutsa:
> Hello. I also do not know exactly how the RPC selection mechanisms for restake work, and I did RPC not for Restake, but for my other needs. I really have min-gas-prices = "0.0025uqck" - but seeing this issue I changed it to "0.0001uqck". Hopefully now my public endpoint will help with Restake as well. That being said, it would of course be nice to have a defined "honor code" or "recommendation base" for public endpoints.

JeremyParish69:
> Looking at the schema, it doesn't seem like there's any property to do this sort of flagging with today. Is that accurate?

Correct, there's no flagging system in the Chain Registry currently.

AutoStake-com:
I noticed this PR removes our RPC

It is correct the RPC was not working (a casualty of our .net => .com rebrand)

Instead in the future please just reach out so that we can fix any services

JeremyParish69:
@blacklodge-admin can you please adjust based on the comments? Some of these should be working now

blacklodge-admin:
> @blacklodge-admin can you please adjust based on the comments? Some of these should be working now

I think we should be good to go now, if you can take another look.  Thanks!

PR #1717, opened 2023-03-26
crescent-irisnet,umee add

nodebreaker0-0:


_IBC/crescent-irisnet.json
```diff
@@ -0,0 +1,32 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "crescent-1",
```
JeremyParish69, line 4: these chain_names refer to the name used in the Chain Registry, and should not have the '-1' at the end, which I assume is referencing the corresponding chain_id.

nodebreaker0-0: Thanks for the feedback, fixed it, it's been a while and I forgot about the convention.

PR #1719, opened 2023-03-27
Add nolus-protocol testnet

ivivanov:


JeremyParish69:
FYI, the padding with shrink the logo, (and doesn't match the submission to Osmosis frontend. Are we able to remove the padding?

ivivanov:
Thanks for the review. Padding removed

PR #1723, opened 2023-03-27
add stinj

riley-stride:
add stINJ


stride/assetlist.json
```diff
@@ -175,6 +175,36 @@
...
       }
     },
+    {
+      "denom_units": [
+        {
+          "denom": "stinj",
+          "exponent": 0
+        },
+        {
+          "denom": "stinj",
```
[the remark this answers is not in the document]
riley-stride: fixed, thank you @JeremyParish69

stride/assetlist.json
```diff
@@ -175,6 +175,36 @@
...
+          "exponent": 0
+        },
+        {
+          "denom": "stINJ",
+          "exponent": 18
+        }
+      ],
+      "base": "stinj",
+      "name": "stINJ",
+      "display": "stinj",
```
JeremyParish69, line 191: Which means this should also be 'stINJ' to show that frontends should use the 18 exp unit.

riley-stride: Ah good catch. Sorry about the back and forth. Done.

PR #1728, opened 2023-03-28
Create _middleware_ICS20_memo_keys.json

womensrights:
As discussed in the telegram group back in February, this pr adds a registry for ICS20 memo keys used for executable actions after a successful token transfer - not just a purely informational memo and typically parsed by middleware wrapping token transfer. The registry would ensure that the same key isn’t used for different use cases and the same middleware could be implemented on different chains. This is to avoid the case where a chain adopts multiple middleware in the same stack which are using the same keys. Further context can be found in the [ICS20 specification](https://github.com/cosmos/ibc/blob/main/spec/app/ics-020-fungible-token-transfer/README.md#using-the-memo-field) for the motivation to have a memo key registry.


JeremyParish69:
The check is failing because of it's location being in the _IBC folder. We could either move it out of the folder, or adapt the check in some way...
As for how we can enforce this to follow a structure, I suppose we could have an array of objects containing, `key`, `description`, `git_repo`, and `memo` (all required?), but going any more in depth seems quite complicated, as the memo object is different in almost every case. But in one case the memo isn't an object, but a string--this might be tricky to create a schema for...

womensrights:
I just checked in on that entry which was a string, and it is now a JSON format and updated in the last commit - so all entries are adhering to the same format. Agree on the difficulty with the check failing, either solution could work, I think it might be best if I make a schema for this - an _ICS20_memo_keys schema, update the readme and then we could add in a workflow to validate to this schema? 

But the alternative of removing from the folder is also reasonable too, let me know which one you prefer and I will support

_IBC/_middleware_ICS20_memo_keys.json
```diff
@@ -0,0 +1,77 @@
...
+              "channel": "channel-234",
+              "timeout": "10m",
+              "retries": 2
+            }
+          }
+        }
+      }
+    },
+    {
+      "key": "",
```
JeremyParish69, line 62: Is having no key value going to be OK here? I assume we want keys to be unique, so shouldn't no value be off limits?

_IBC/_middleware_ICS20_memo_keys.json
```diff
@@ -0,0 +1,77 @@
...
+            "stride_address": "strideXXX",
+            "airdrop_id": "evmos"
+          }
+        }
+      }
+    },
+    {
+      "key": "forward",
+      "description": "To send a packet from a source to destination chain when there is at least one chain in between, the intermediate chain uses the middleware, and next specifies the subsequent chain to route the packet to, after the chain specified in forward",
+      "git_repo": "https: //github.com/strangelove-ventures/packet-forward-middleware",
```
JeremyParish69, line 41: json schema validation caught the extra space after 'https:'

JeremyParish69:
I think this should work for a schema, wyt?:
```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "memo_keys": {
      "type": "array",
      "items": {
        "type": "object",
        "required": [
          "key",
          "description",
          "git_repo",
          "memo"
        ],
        "properties": {
          "key": {
            "type": "string"
          },
          "description": {
            "type": "string"
          },
          "git_repo": {
            "type": "string",
            "format": "uri"
          },
          "memo": {
            "type": "object"
          }
        },
        "additionalProperties": true
      }
    }
  },
  "additionalProperties": false
}
```

_IBC/_middleware_ICS20_memo_keys.json
womensrights: I removed this entry for now, agree this is not strictly the format that should be used

womensrights:
I addressed your comments in the two new commits:

- Removed the entry without the explicit memo key format (will follow up with the team directly)
- Removed the space in the URL for one of the entries
- changed the file name to _memo_keys.json
- added in the schema

JeremyParish69:
Regarding the validation failure we still see; want to throw this file into a new folder as well? I saw you thumbs up my earlier comment recommending a `/_memo_keys/` folder, but didn't implement it. That should resolve the validation failure, and then adding validation to work with the new schema should be pretty simple, too.

womensrights:
I had misunderstood, but now I changed the file name and added the folder and everything seems to pass now

_memo_keys/ICS20_memo_keys.json
```diff
@@ -0,0 +1,62 @@
...
+    {
+      "key": "ibc_callback",
+      "description": "Passing the ibc acknowledgement back to a cosmwasm smart contract",
+      "git_repo": "https://github.com/osmosis-labs/osmosis/tree/main/x/ibc-hooks#ibc-hooks",
+      "memo": {
+        "ibc_callback": "osmo1contractAddr"
+      }
+    },
+    {
+      "key": "autopilot",
```
asalzmann, line 25: `autopilot` can support multiple message types https://github.com/Stride-Labs/stride/tree/main/x/autopilot

For example, 1-click liquid stake
```
{ 
    "autopilot": {
          "receiver": "strideXXX", 
          "stakeibc": {
               "stride_address": "strideXXX",
               "action": "LiquidStake",
          }
    }
}
```

and Update Airdrop Address
```
{ 
    "autopilot": {
          "receiver": "strideXXX", 
          "claim": {
               "stride_address": "strideXXX",
               "airdrop_id": "evmos",
          }
    }
}
```

Can we include both here?

JeremyParish69: just thinking out loud--The key objects have been set to 'additionalProperties: true', so we might be able to do something like 'memo' and 'memo2'. (no schema change required).
Or we could change memo to an array, 'memos', and then we can have one or multiple memo structures associated with a single key (schema changes required). @womensrights wyt?

PR #1742, opened 2023-03-30
add polkachu endpoints to noble and quasar

PolkachuIntern:


JeremyParish69:
Noble's chain.json was merged in the middle of a structure change, so the check is failing. Please see: https://github.com/polkachu/chain-registry/pull/1

PR #1758, opened 2023-03-31
Add Nois

tombeynon:
Need to update logos before merging, otherwise @webmaster128 has verified info is correct.

Slip has already been registered: https://github.com/satoshilabs/slips/pull/1508


tombeynon:
Good to go - thanks!

webmaster128 (approved):
LGTM thank you!

PR #1772, opened 2023-03-31
Add versions array with codebase data

JeremyParish69:
Adds the 'current' codebase data to a 'version' object within a 'versions' array in the chain.json


ValarDragon (approved):
approach LGTM! Did not review the change for each chain

PR #1779, opened 2023-04-01
May be closed

staking-explorer:
This provider provides API not from Agoric project ("chain_id": "agoric-3"). If you check this URI: https://agoric-lcd.easy2stake.com/cosmos/base/tendermint/v1beta1/blocks/latest you can see this output: "chain_id": "cosmoshub-4". Unfortunately, such errors break dozens of applications using the chain-registry project. Provider easy2stake.com must be banned to make changes in chain-registry project.

### DIFF
--- a/regen/assetlist.json
+++ b/regen/assetlist.json
@@ -29,6 +29,10 @@
       "denom_units": [
         {
           "denom": "eco.uC.NCT",
+          "exponent": 0
+        },
+        {
+          "denom": "NCT",
           "exponent": 6
         }
       ],
@@ -39,7 +43,8 @@
       "logo_URIs": {
         "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/regen/images/nct.svg",
         "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/regen/images/nct.png"
-      }
+      },
+      "coingecko_id": "toucan-protocol-nature-carbon-tonne"
     }
   ]
 }


staking-explorer:
Now checking URL is not available. Apparently the provider has already managed to turn off its service. But such mistakes are simply unforgivable! Because of this provider, for example, a bunch of services that use proxies from cosmos.directory that is getting data from chain-registry (staking services as RESTAKE, light explorers as PING, some wallets - don`t work as expected). Provider easy2stake.com must be banned!

tombeynon:
FWIW cosmos.directory will validate the chain ID and exclude nodes with the wrong value from the proxies. It shouldn't affect restake or any apps using the proxies. 

Errors like this are an easy mistake to make. IMO apps should expect to have to validate the APIs published in the registry. These are public/free nodes after all.

staking-explorer:
We don't think this is a mistake. We believe that unscrupulous validators do this on purpose to get top-5 positions in the RESTAKE app. And because of their efforts, each time instead of one request, you have to do 2 requests. Because you don’t know from which network you get the results when you access cosmos/mint/v1beta1/inflation, cosmos/bank/v1beta1/supply, cosmos/staking/v1beta1/pool, cosmos/staking/v1beta1/validators, cosmos/gov/v1beta1/proposals etc. No "chain_id" at this points... And when you want to get ~1gb paginated data from cosmos/auth/v1beta1/accounts this is a problem.

staking-explorer:
We just noticed another fake API. The validator successfully attracts delegates from the third position in the RESTAKE. Should we make a pull request?

staking-explorer:
Just make pool request to ban chaintools.tech for cloaking in Juno network.

staking-explorer:
OK. We have not slept for three months and are trying to make a detailed vesting map for the cosmos ecosystem. After analyzing our logs, we can state a complete mess among the APIs in the registry. Many of the nodes give incorrect data (it would be better if they gave an error). We will offer our own tool for cleaning the registry from fake, broken, abandoned, stopped nodes. The request can be closed.

PR #1798, opened 2023-04-04
add ibc and update logo

ShenFri:


_IBC/uptick-cosmos.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "uptick",
```
JeremyParish69, line 4: for both ibc files, they must be ordered alphabetically, so they should be named cosmoshub-uptick.json, with chain 1 being cosmoshub, and chain 2 being uptick

_IBC/uptick-iris.json
```diff
@@ -0,0 +1,26 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "uptick",
+    "client_id": "07-tendermint-2",
+    "connection_id": "connection-0"
+  },
+  "chain_2": {
+    "chain_name": "iris",
```
JeremyParish69, line 9: same here, irisnet should be chain 1, and uptick as chain 2 (and file renamed accordingly)

ShenFri: ok

JeremyParish69:
Oh, and the 'version' property is required in both files (e.g., `"version": "ics20-1"`)

PR #1825, opened 2023-04-07
add Secret.Express API to secret, remove stakely/trivium

ertemann:


ertemann:
Cleaned up the APIs as some of them stopped supporting - also Secret.Express is a combination of 4 different teams who were all in the section individually.

Best use for people is Secret.Express, it currently has 72 nodes behins it.

secretnetwork/chain.json
```diff
@@ -126,36 +126,22 @@
   "apis": {
     "grpc-web": [
       {
-        "address": "https://secret-4.api.trivium.network:9091",
-        "provider": "Trivium | Trivium.Network"
+        "address": "https://wgrpc.secret.express",
+        "provider": "Secret Community API team - Secret Express"
       },
```
JeremyParish69, line 131: need to remove this comma

ertemann: sorry for this simple mistake. fixed

PR #1831, opened 2023-04-10
Updated stride testnet to v8

sampocs:


sampocs:
Sorry can you clarify that further? Was a new versions array added somewhere?

JeremyParish69:
> Sorry can you clarify that further? Was a new versions array added somewhere?

It's just that this branch is more than 100 commits behind master, and since then there has been a bit of a structure update to the chain.json files. We initiated the 'versions' array in the codebase section, and strive to keep 'versions' up to date (by adding v8 as another version object) for the upcoming CI that will rely on that. See this object visible in 'master': https://github.com/cosmos/chain-registry/blob/ceac711f4bc842e280d50353790f1feac2ebfd44/testnets/stridetestnet/chain.json#L45-L59

sampocs:
Ah sorry I didn't realize the fork was so behind. Just pushed up the updates!

PR #1840, opened 2023-04-11
Update uptick ibc info

YangChuantongStarry:


_IBC/uptick-irisnet.json
```diff
@@ -0,0 +1,28 @@
+{
```
JeremyParish69, line 1: This file is invalid. irisnet-uptick.json already exists.

PR #1853, opened 2023-04-12
SHD Migration

DrPresident:
Updated SHD to the latest denomination since it was migrated on secret network


JeremyParish69:
Replaced by PR https://github.com/cosmos/chain-registry/pull/1865

PR #1857, opened 2023-04-13
update ibc path dydxtestnet-osmosistestnet

clemensgg:


JeremyParish69:
oh, dydxtestnet chain hasn't been registered yet. Could we get that in here as well?

clemensgg:
@JeremyParish69 I think the current private testnet isn't intended to, but we'll add the public dydx testnet as soon it is launched!

JeremyParish69:
> @JeremyParish69 I think the current private testnet isn't intended to, but we'll add the public dydx testnet as soon it is launched!

So this is the private testnet? Just don't want that to get confusing. We'd need to make sure the public testnet doesn't get registered as dydxtestnet, then, or else this connection would be inaccurate. Maybe if the chain isn't going to be registered we should call it something like 'dydxprivatetestnet' ?

clemensgg:
> 

updated suggested changes

PR #1861, opened 2023-04-13
Umee historical versions

doggystylez:
added versions for all umee-1 upgrades.

@JeremyParish69 I'm wondering if the `tag`, `height`, and `next_version_name` keys should appear as in [osmosis/chain.json](https://github.com/cosmos/chain-registry/blob/aa883fb98406e82ca53e70bcdbb156bfba0c83b8/osmosis/chain.json#L63)

### DIFF
--- a/umee/chain.json
+++ b/umee/chain.json
@@ -30,9 +30,9 @@
   },
   "codebase": {
     "git_repo": "https://github.com/umee-network/umee",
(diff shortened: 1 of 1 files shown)


JeremyParish69:
oh, this is fantastic. Having tag, height, and next_version_name would be ideal

PR #1863, opened 2023-04-13
Change gas price for Uptick

Sr20dem:
It is necessary to increase the gas price, otherwise, when trying to delegate, an error appears:  
`provided fee < minimum global fee (25auptick < 1000000000000000auptick). Please increase the gas price.: insufficient fee: unknown request`

### DIFF
--- a/uptick/chain.json
+++ b/uptick/chain.json
@@ -17,10 +17,10 @@
     "fee_tokens": [
       {
         "denom": "auptick",
-        "fixed_min_gas_price": 10000000000,
-        "low_gas_price": 10000000000,
-        "average_gas_price": 25000000000,
-        "high_gas_price": 40000000000
+        "fixed_min_gas_price": 1000000000000000,
+        "low_gas_price": 1000000000000000,
+        "average_gas_price": 2500000000000000,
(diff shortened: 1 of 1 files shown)


JeremyParish69:
What interface are you using to delegate? 25auptick suggests that the gas setting in the wallet isn't correct. But also not applying the gas fee that's already in there.

Sr20dem:
> What interface are you using to delegate? 25auptick suggests that the gas setting in the wallet isn't correct. But also not applying the gas fee that's already in there.

I have added Uptick network to Keplr on the Restake site. And I'm trying to delegate tokens on the restake
uptick-restake-error

tombeynon:
Hey @Sr20dem - if you're seeing that error it's likely you added the chain to Keplr a while ago, before REStake started using the Chain Registry config for gas prices (25auptick is the old default). And actually Keplr now ignores the config the site sets by default anyway.

I suspect you added the chain to Keplr a while ago - can you try deleting it from Keplr (click the X in the network list in the extension) and re-add it from REStake? That will use the correct gas prices, and should resolve your issue.

To clarify, I think the values the Chain Registry has already are correct.

PR #1872, opened 2023-04-14
Add testnet and update assets in mainnet

ChimbaBlockchain:


chimba/assetlist.json
```diff
@@ -6,23 +6,19 @@
         "description": "The first blokchain in colombia",
         "denom_units": [
           {
-            "denom": "ucmba",
-            "exponent": 0
```
JeremyParish69, line 10: at least one denom_unit of exponent 0 is required.

PR #1907, opened 2023-04-18
add coingecko-id to whale token

PFC-developer:
adds coingecko-id to $WHALE on migaloo chain


PFC-developer:
switched to draft, and coingecko is not liking the coin

PR #1909, opened 2023-04-18
Add PEPE to chain registry

OsmoPepe:


juno/assetlist.json
```diff
@@ -1174,6 +1174,27 @@
       "symbol": "DOGA",
       "logo_URIs": {
         "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/juno/images/doga.png"
+      },
```
JeremyParish69, line 1177: missing a closing bracket here

PR #1916, opened 2023-04-18
add statesync to schema

gregnuj:
Add statesync section to schema

### DIFF
--- a/chain.schema.json
+++ b/chain.schema.json
@@ -528,6 +528,7 @@
       }
     }
   },
+  "additionalProperties": false,
   "$defs": {
     "peer": {
       "type": "object",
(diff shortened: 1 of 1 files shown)


chain.schema.json
```diff
@@ -502,7 +528,6 @@
       }
     }
   },
-  "additionalProperties": false,
```
JeremyParish69, line 505: why removing this restriction?

gregnuj: it is duplicated, looking at line 505 and line 613 of the most recent commit
https://github.com/cosmos/chain-registry/blob/3caf657a917a32ec478b98932e898beea990e2f5/chain.schema.json

PR #1951, opened 2023-04-22
Staking hub patch 1 1

staking-hub:
Add light token to chain registry. No chain exists


light/assetlist.json
```diff
@@ -0,0 +1,27 @@
+{
+    "$schema":"../assetlist.schema.json",
+    "chain_name":"light",
```
JeremyParish69, line 3: It sounds like this isn't its own chain, but rather a token on lumenx, in which case please add this to the assetlist there instead.

staking-hub: It is a juno-chain token and has no chain (neither genesis not validators). Should I include it below lumen assetlist?

JeremyParish69: if it's a juno token then it should go in Juno's assetlist. Why would it go in Lumen's assetlist?

staking-hub:
In fact it was minted using junod.binary, but it is token used for providing liquidity incentives in lumen pools.

LuckyStar

Enviado desde Proton Mail móvil

-------- Mensaje original --------

JeremyParish69:
> In fact it was minted using junod.binary, but it is token used for providing liquidity incentives in lumen pools.

Thanks for explaining. Purpose is irrelevant to this. It should go in Juno's assetlist

JeremyParish69:
Also, the PNG and SVG are too different. I'd suggest just sticking with the PNG

staking-hub:
Done. Removed the .SVG file.
Uploaded the .PNG file into Juno/images
Opened a new PR to list light in juno.

LuckyStar

Enviado desde Proton Mail móvil

-------- Mensaje original --------

JeremyParish69:
LIGHT was added in https://github.com/cosmos/chain-registry/pull/1958

PR #1954, opened 2023-04-24
chain name correction on testnet assets

ChimbaBlockchain:


JeremyParish69:
Thanks for flagging this. Could you please correct the folder name, and the chain_name in both assetlist.json and chain.json to 'chimbatestnet'?

PR #1966, opened 2023-04-25
change denom name

ChimbaBlockchain:


JeremyParish69:
The base denom must have 0 exponent. ucmba has 0 exponent and is OK as a base, but chimba has 6 exponent, and therefore cannot be used as the base. If 'chimba' is indeed the base denom and has 0 exponent, then the denom_units need to be redefined and I assume there should be another denom that has 6 exponent.

PR #1967, opened 2023-04-25
Add VOID to chain-registry

Jason0xOg:


juno/assetlist.json
```diff
@@ -1285,6 +1285,28 @@
...
+          "exponent": 0
+        },
+        {
+          "denom": "void",
+          "exponent": 6
+        }
+      ],
+      "base": "cw20:juno1lpvx3mv2a6ddzfjc7zzz2v2cm5gqgqf0hx67hc5p5qwn7hz4cdjsnznhu8",
+      "name": "Void",
+      "display": "Void",
```
JeremyParish69, line 1305: display should use lower case 'v' (to match the denom_unit)

Jason0xOg: Fixed @JeremyParish69

PR #1977, opened 2023-04-26
Add PublicNode RPC, REST, GRPC for Terra

3eph1r0th:


terra2/chain.json
```diff
@@ -130,6 +130,11 @@
         "address": "terra-seed-1.allnodes.me:26656",
         "provider": "Allnodes.com ⚡️ Nodes & Staking"
       },
+      {
+        "id": "e726816f42831689eab9378d5d577f1d06d25716",
+        "address": "terra-seed-2.allnodes.me:26656",
+        "provider": "Allnodes.com ⚡️ Nodes & Staking"
```
JeremyParish69, line 136: I suspect this is a typo from copying the one above...

JeremyParish69: Actually, I'm not sure why this PR is adding this...

JeremyParish69: I'm a goofball, not realizing the different providers are actually related

PR #1981, opened 2023-04-26
Adding many missing logo_URIs in chaininfo json files.

aweigold:
Utilized existing token logos, online sources, and generation of svgs from pngs.


dig/chain.json
```diff
@@ -53,6 +53,10 @@
       }
     ]
   },
+  "logo_URIs": {
+    "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/dig/images/dig.png",
+    "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/dig/images/dig.svg"
```
JeremyParish69, line 58: IMO that dig svg is way too different than the PNG and should probably be omitted

nyx/chain.json
```diff
@@ -45,6 +45,10 @@
       }
     ]
   },
+  "logo_URIs": {
+    "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/nyx/images/nyx.png",
+    "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/nyx/images/nyx.svg"
```
JeremyParish69, line 50: same with this svg, looks like a bad recreation attempt and should probably be omitted

mises/chain.json
```diff
@@ -47,6 +47,10 @@
       }
     ]
   },
+  "logo_URIs": {
+    "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/mises/images/mises.png",
+    "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/mises/images/mises.svg"
```
JeremyParish69, line 52: These generated svgs are all black; not really accurate...

aweigold:
@JeremyParish69 agreed on that, they are removed... we were working on a project where I needed them, but I'll omit them in this PR.

PR #1982, opened 2023-04-26
Update chain.json

reversesigh:
Desmos
Adding 4.8.1 patch - https://github.com/desmos-labs/desmos/releases/tag/v4.8.1

Sommelier
Update from v5 to v6 at 8704480 pending approval of Prop 47: https://www.mintscan.io/sommelier/proposals/47

XPLA
Adding seed, RPC and REST for Lavender.Five Nodes

### DIFF
--- a/desmos/chain.json
+++ b/desmos/chain.json
@@ -35,7 +35,7 @@
     "git_repo": "https://github.com/desmos-labs/desmos",
     "recommended_version": "v4.8.0",
     "compatible_versions": [
-      "v4.8.0"
+      "v4.8.0", "v4.8.1"
     ],
     "binaries": {
       "linux/amd64": "https://github.com/desmos-labs/desmos/releases/download/v4.8.0/desmos-4.8.0-linux-amd64",
--- a/desmos/chain.json
+++ b/desmos/chain.json
@@ -51,7 +51,7 @@
         "name": "v4.8.0",
         "recommended_version": "v4.8.0",
         "compatible_versions": [
-          "v4.8.0"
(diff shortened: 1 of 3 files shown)


sommelier/chain.json
```diff
@@ -29,9 +29,9 @@
   },
   "codebase": {
     "git_repo": "https://github.com/PeggyJV/sommelier",
-    "recommended_version": "v5.0.0",
```
JeremyParish69, line 32: Can we please also create a new object at the bottom of the versions array for this v6?

desmos/chain.json
```diff
@@ -51,7 +51,7 @@
         "name": "v4.8.0",
         "recommended_version": "v4.8.0",
         "compatible_versions": [
-          "v4.8.0"
+          "v4.8.0", "v4.8.1"
```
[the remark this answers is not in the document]
reversesigh: This should be corrected now and has been committed. Thanks for pointing this out.

sommelier/chain.json
reversesigh: Apologies for overlooking that. Resubmitting with proper changes now.

JeremyParish69: looks like it's still removing v5 even though we want to keep that

reversesigh: Apologies if I'm not following- if the current chain version is v6, why do we keep v5 on here?

JeremyParish69: to keep a record of versions. Ideally we have all versions in here, with their effective block height.

PR #1996, opened 2023-04-27
add slica and pepec token

vigorousdeveloper:
SLCA token is already on osmosis but there is no logo


juno/assetlist.json
```diff
@@ -1307,6 +1307,50 @@
...
+        {
+          "denom": "cw20:juno10vgf2u03ufcf25tspgn05l7j3tfg0j63ljgpffy98t697m5r5hmqaw95ux",
+          "exponent": 0
+        },
+        {
+          "denom": "pepec",
+          "exponent": 6
+        }
+      ],
+      "base": "cw20:juno10vgf2u03ufcf25tspgn05l7j3tfg0j63ljgpffy98t697m5r5hmqaw95ux",
```
JeremyParish69, line 1347: The same contract is specified for both assets--they need to have different base denoms. Please correct.

PR #1997, opened 2023-04-27
Upgrade Gravity to v1.9.0

dylanschultzie:
proposal: https://www.mintscan.io/gravity-bridge/proposals/172


dylanschultzie:
Happening now, so didn't bother adding estimated time

PR #2013, opened 2023-04-29
Fix git repo and version for nyx node

PaulZig:
Change codebase section point to repo with node code.


JeremyParish69:
Can we please keep the versions array and add the new version to the bottom of that as well?

PR #2015, opened 2023-04-29
add realio network Mainnet

Genz22:


realio/chain.json
```diff
@@ -0,0 +1,124 @@
...
+        }
+      ]
+    },
+    "staking": {
+      "staking_tokens": [
+        {
+          "denom": "ario"
+        },
+        {
+          "denom": "arst"
```
JeremyParish69, line 33: just want to double check, because this would be a first. Both these tokens are staking tokens for securing the network?

realio/assetlist.json
```diff
@@ -0,0 +1,49 @@
...
+            "denom": "rst",
+            "exponent": 18
+          }
+        ],
+        "base": "arst",
+        "name": "Realio Security Token",
+        "display": "rst",
+        "symbol": "RST",
+        "logo_URIs": {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/realio/images/rst.png"
```
JeremyParish69, line 43: can we please replace the jpg with a png?

realio/chain.json
Genz22: yes , realio have two native . Delegators can stake rio and rst

PR #2018, opened 2023-04-30
Update chain.json for Jackal for v2.0.0

reversesigh:
Updating codebase to v2.0.0 and adding v2.0.0 to version history. Upgrade happening at height 2631260 pending approval of Prop 4 (estimated to be May 4). https://explorer.nodestake.top/jackal/gov/4


dylanschultzie (approved):

JeremyParish69:
We also have properties for block height and proposal, which will be used in the upcoming automation. https://github.com/reversesigh/chain-registry/blob/796340afabe7f3139b9eccdbcdea2c68e7dc0548/jackal/chain.json#L63

Issue #2022, opened 2023-05-01
Create template for upgrade pull requests


dylanschultzie:
@JeremyParish69 I'm going to create a template so that you don't have to keep going into these PRs and telling me (and others) to add other things. The plan is:

1. Normalize PR name to `[upgrade] {network} {version}`
2. Add estimated time to be filled in for the PR body:
```
Estimated Upgrade Time: 
``` 
3. Add a comment block for the PR that shows and example of (new) major version increases.

PR #2024, opened 2023-05-01
Create template for upgrades

dylanschultzie:
Issue: #2022 

This is just a first pass @JeremyParish69 for your thoughts


JeremyParish69:
So far I really like what I see. Is there a way to make this appear only when the PR is for an upgrade?

MasterPi-2124:
I created a similar action here, and it requires manual input from users: https://github.com/notional-labs/chain-registry/actions/workflows/add_version.yml

But I also really like your proposed solution, so can we discuss more about that? @dylanschultzie @JeremyParish69

JeremyParish69:
Interesting idea. Can be brought up again. PR has been abandoned.

PR #2043, opened 2023-05-04
Bump StaFiHub version to v0.4.1 && Add rSWTH

Tore19:


stafihub/chain.json
```diff
@@ -30,19 +30,19 @@
...
     "compatible_versions": [
-      "v0.2.3"
+      "v0.4.1"
     ],
     "genesis": {
       "genesis_url": "https://raw.githubusercontent.com/stafihub/network/main/mainnets/stafihub-1(dragonberry)/genesis.json"
     },
     "versions": [
       {
-        "name": "v0.2.3",
```
JeremyParish69, line 42: Can we please keep v0.2.3 in the versions array, and then just add v0.4.1 after it?

Tore19: Sure. Please check again. Thanks!

PR #2053, opened 2023-05-05
Add USDC image for noble chain

boojamya:


noble/assetlist.json
```diff
@@ -45,7 +45,8 @@
       "name": "Frienzies",
       "symbol": "FRNZ",
       "logo_URIs": {
-        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/noble/images/frnz.png"
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/noble/images/frnz.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/noble/images/frnz.svg"
```
JeremyParish69, line 49: Is there anything we can do to fix this SVG? It looks not right to me...

boojamya: Hmm good catch!
Investigating!

boojamya: If I download the image locally and view it, it seems normal. 
We'll keep this as-is for now. If we get a different version, we'll create another PR.
Thanks @JeremyParish69!

JeremyParish69: you're right! My bad. Thanks for checking!

PR #2056, opened 2023-05-05
Revert "Del kingnodes REST endpoint"

nullmames:
Reverts cosmos/chain-registry#1986

Hi @JeremyParish69.
Our rest endpoints are indeed REST endpoints. We use caddy to map the path to the cosmos API. There is nothing wrong with it other than we block some features. @staking-explorer should have messaged me rather than get our endpoints removed. Can you please revert this PR.

### DIFF
--- a/evmos/chain.json
+++ b/evmos/chain.json
@@ -289,6 +289,10 @@
     ],
     "rest": [
       {
+        "address": "https://evmos.kingnodes.com",
+        "provider": "kingnodes"
+      },
+      {
         "address": "https://lcd-evmos.whispernode.com:443",
         "provider": "WhisperNode🤐"
       },


JeremyParish69:
Apologies! I think we'll need to be more careful removing links in the future.

nullmames:
Thanks Jeremy.

staking-explorer:
@nullmames, @JeremyParish69 we are sorry. This is really REST. An interesting scheme that no one else uses. This approach allows you to make fewer mistakes at the URL\DNS, because the path to RPC and REST is the same. We are sorry.

nullmames:
People make mistakes sometimes. No problem. We were able to get it restored very quickly.

PR #2069, opened 2023-05-09
Add Babylon

SebastianElvis:
This PR adds Babylon to chain registry.

We have already added Babylon to SLIP-0173 via [this PR](https://github.com/satoshilabs/slips/pull/1547)


babylon/chain.json
```diff
@@ -0,0 +1,113 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "babylon",
+  "status": "live",
+  "website": "https://www.babylonchain.io",
+  "network_type": "testnet",
```
JeremyParish69, line 6: are we adding the mainnet or testnet? testnets should in the /testnets/ directory and the chainname would be 'babylontestnet'.

SebastianElvis: Thanks for the prompt response. Moved the entire folder to `testnets/babylontestnet` and renamed corresponding places.

PR #2075, opened 2023-05-09
update gaiad version to v9.1.0

doggystylez:
pls note: 
* fixed v9 upgrade name
* no upgrade name for v9.1.0 because no `SoftwareUpgradeProposal`

### DIFF
--- a/cosmoshub/chain.json
+++ b/cosmoshub/chain.json
@@ -63,7 +63,7 @@
         }
       },
       {
-        "name": null,
+        "name": "v9.1.0",
         "recommended_version": "v9.1.0",
         "compatible_versions": [
            "v9.1.0"
--- a/cosmoshub/chain.json
+++ b/cosmoshub/chain.json
@@ -33,23 +33,23 @@
   },
   "codebase": {
     "git_repo": "https://github.com/cosmos/gaia",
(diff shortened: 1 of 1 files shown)


doggystylez:
oops looks like the bot doesn't like
`"name": null,`
can do
 `"name": "none",` or similar?

JeremyParish69:
> oops looks like the bot doesn't like `"name": null,` can do `"name": "none",` or similar?

null is not a string. But I'd prefer just 'v9.1' over 'none'; makes way more sense, IMO, even if it's not official.

doggystylez:
I think in that case the full tag `v9.1.0` probably would make the most sense. however I also think there maybe should be another field in the schema to indicate whether a consensus breaking upgrade has an upgrade proposal, because if not it must be handled differently than a scheduled upgrade/

JeremyParish69:
> I think in that case the full tag `v9.1.0` probably would make the most sense. however I also think there maybe should be another field in the schema to indicate whether a consensus breaking upgrade has an upgrade proposal, because if not it must be handled differently than a scheduled upgrade/

Wait, so this was an unscheduled consensus-breaking upgrade? I assumed that would never happen, except for maybe emergency patches...
Still I think this having a name makes sense--proposal or not.

Issue #1926, opened 2023-04-19
Broken Links


MasterPi-2124:
I implemented a CI workflow that checks for all broken URLs, and the result is surprisingly bad.

https://github.com/cosmos/chain-registry/actions/runs/4745692057/attempts/1#summary-12870701196

PR #2087, opened 2023-05-11
Check Broken Links

MasterPi-2124:
mentioned in #1926


JeremyParish69:
So there are so many broken links that this can't even create the issue due to body text length limitation? So we can't even see here the result of what's broken to be able to fix it.
I was planning on going through a different list of broken links and fixing the broken urls for images and removing the broken explorers. APIs will take more planning.

PR #2113, opened 2023-05-15
chore: 🤖 include eden asset info to elystestnet

cosmic-vagabond:
Added description of the secondary native token from Elys TestNet called EDEN.

### DIFF
--- a/testnets/elystestnet/assetlist.json
+++ b/testnets/elystestnet/assetlist.json
@@ -63,8 +63,7 @@
       "symbol": "EDEN",
       "logo_URIs": {
         "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/elystestnet/images/eden.png"
-      },
-      "coingecko_id": ""
+      }
     }
   ]
 }


testnets/elystestnet/assetlist.json
```diff
@@ -33,6 +33,38 @@
...
+        }
+      ],
+      "base": "ueden",
+      "name": "Eden",
+      "display": "eden",
+      "symbol": "EDEN",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/elystestnet/images/eden.png"
+      },
+      "coingecko_id": "eden"
```
JeremyParish69, line 67: I'm fairly certain this coingecko id corresponds to an unrelated asset

ollieottersync: @JeremyParish69 should we leave this field blank until we register it there later?

cosmic-vagabond: @JeremyParish69 I noticed that the `coingecko_id` field is actually optional, so I just removed it for the eden asset

PR #2132, opened 2023-05-18
Terra patch 6

gregnuj:
Add new classic version, move version specific info under each version


JeremyParish69:
Great work adding all this information, @gregnuj . My only concern is the removal of recommended and especially compatible versions from each of the version objects. I think it conveys which versions are grouped under that version object. Could we add those back in?

gregnuj:
@JeremyParish69 I have added them back in, but I personally don't think those sections make sense in the context of a specific version, I think it would make much more sense to have distinct version object for each "compatible versions" and use a shared name across versions, then each version would retain it's own set of binaries and sdk/consensus/ibc version info.

JeremyParish69:
Thanks @gregnuj.
So It sounds like you're saying it'd be better to separate each patch into it's own version object?
Although, I agree that retaining more data is better, I also think that the decision to separate version groups into separate version objects is when one is not compatible with the last (assuming patches are compatible, but minor versions are often not).
We can discuss further on this if you like.

PR #2139, opened 2023-05-19
Add template dir

dylanschultzie:
Thought this might be helpful, so people don't have to always re-write things.

### DIFF
--- a/_template/assetlist.json
+++ b/_template/assetlist.json
@@ -6,7 +6,7 @@
         "description": "The native staking token of EXAMPLE.",
         "denom_units": [
           {
-            "denom": "uexample",
+            "denom": "uexp",
             "exponent": 0
           },
           {
@@ -16,11 +16,11 @@
         ],
         "base": "uexample",
         "name": "Example",
-        "display": "exp",
+        "display": "example",
         "symbol": "EXP",
         "logo_URIs": {
-          "png": "",
-          "svg": ""
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_template/images/chain_image.png",
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_template/images/chain_image.svg"
         },
         "coingecko_id": ""
       }
--- a/_template/assetlist.json
+++ b/_template/assetlist.json
@@ -1,0 +1,28 @@
+{
+    "$schema": "../assetlist.schema.json",
+    "chain_name": "",
+    "assets": [
+      {
+        "description": "The native staking token of EXAMPLE.",
+        "denom_units": [
+          {
+            "denom": "uexample",
+            "exponent": 0
+          },
+          {
+            "denom": "example",
+            "exponent": 6
+          }
+        ],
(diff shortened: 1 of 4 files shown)


JeremyParish69:
an interesting idea, but can the readme examples not suffice?

dylanschultzie:
@JeremyParish69 my concern is with new projects joining - I've talked with many, and they always seem to be fairly overwhelmed. My thought here is giving a blank slate with a few examples, but otherwise `include all these things` type of idea.

KagemniKarimu:
@dylanschultzie  +1, this simplifies the process for us greatly.

_template/assetlist.json
```diff
@@ -0,0 +1,28 @@
...
+            "exponent": 0
+          },
+          {
+            "denom": "example",
+            "exponent": 6
+          }
+        ],
+        "base": "uexample",
+        "name": "Example",
+        "display": "exp",
```
JeremyParish69, line 19: this display should match a denom_unit. You could change to 'example'. (It's something many PRs get wrong)

_template/chain.json
```diff
@@ -0,0 +1,105 @@
...
+  },
+  "codebase": {
+    "git_repo": "",
+    "recommended_version": "v1.0",
+    "compatible_versions": [
+      "v1"
+    ],
+    "cosmos_sdk_version": "",
+    "cosmwasm_enabled": false,
+    "cosmwasm_path": "$HOME/.example/data/wasm",
```
JeremyParish69, line 42: is it possible for comswasm path to be valid while also having cosmwasm_enabled being false?

_template/chain.json
```diff
@@ -0,0 +1,105 @@
...
+      {
+        "denom": ""
+      }
+    ]
+  },
+  "codebase": {
+    "git_repo": "",
+    "recommended_version": "v1.0",
+    "compatible_versions": [
+      "v1"
```
JeremyParish69, line 38: We should probably make clear whether or not recommended version must be listed under compatible versions as well. Since it seems to be the more common, I'd say so, and therefore we should write 'v1.0' here (and not v1)

JeremyParish69: oh, actually, seeing the use of v1 below, it's the recommended_version that should change to 'v1'

_template/chain.json
```diff
@@ -0,0 +1,105 @@
...
+        "cosmos_sdk_version": "",
+        "cosmwasm_enabled": false,
+        "binaries": {
+          "linux/amd64": ""
+        }
+      }
+    ]
+  },
+  "logo_URIs": {
+    "png": "",
```
JeremyParish69, line 65: can we put what would be the full github url here, but for chain_name we can use _template, and for the image name, we can use chain_image.png. I think it'd help teams realize they need to upload the image and not refer to some remote location.

dylanschultzie:
@JeremyParish69 thanks for the suggestions, I think I've addressed them all

JeremyParish69:
I think last step is updating the validation to skip this template

PR #2186, opened 2023-05-25
update peer id's

ryssroad:


ryssroad:
How to resolve this conflict?

JeremyParish69:
I resolved it. Is that the change you mean to make? You'll also want to sync your fork so we don't see all these commits on future changes.

ryssroad:
I'm synced my fork and yes, the changes in the chain.json file are exactly those that need to be made

PR #2187, opened 2023-05-25
add Staketab chains data

icohigh:


testnets/upticktestnet/chain.json
```diff
@@ -0,0 +1,83 @@
...
+    ]
+  },
+  "codebase": {
+    "git_repo": "https://github.com/UptickNetwork/uptick",
+    "recommended_version": "v0.2.6",
+    "compatible_versions": [
+      "v0.2.6"
+    ],
+    "cosmos_sdk_version": "0.46.5",
+    "tendermint_version": "0.34.23",
```
JeremyParish69, line 36: oh, tendermin_version was changed for the consensus object (allows for tendermint of cometbft).
Also, if it's not too much trouble, could you please add this version data into the versions array?

icohigh: Check the fix

JeremyParish69:
Please see addition of version object:
https://github.com/Staketab/chain-registry/pull/2

JeremyParish69:
Thanks for adding that. Unfortunately, looks like there are conflicts now that needs to be resolved.

icohigh:
> Thanks for adding that. Unfortunately, looks like there are conflicts now that needs to be resolved.

Check now, please

PR #2189, opened 2023-05-25
Terra patch 7 - upgrade terra2 to v2.3.2

gregnuj:


JeremyParish69:
Thanks for this, @gregnuj . Having two version objects with the same name and same block height I think is going to cause issue with the automation being built. I see that the 'binaries' and 'ibc version' has changed with the new patch. Now, are you of the opinion that we should not merely update the info in the existing v2.3 version object? I think that's what's being used everywhere else and is assumed for upcoming CI, but perhaps we can discuss further if you feel strongly about retaining all the info for multiple patches within the same minor version.
cc @MasterPi-2124

gregnuj:
This is the best solution I have without manipulating the schema too much.  So let me describe my use case, I currently have a docker container I use to launch any cosmos chain using cosmovisor, the goal is to be able to launch a node either starting from the Genesis block and syncing fully, or launch using state sync.  Additionally I have to be able to update an existing version.  I want all of this to work automatically to limit the amount of manual changes, so to accomplish this I have an entrypoint script that matches the upgrade name from upgrade-info.json to the values in chain registry.  The problem I have encountered is the on chain upgrade name is too unpredictable and inconsistent to match to the versions listed.   The best example of this is the cosmoshub v9.3.0 upgrade which was a consensus breaking change executed off chain.  Even setting the version appropriately on startup cosmovisor sees the version name does not match the last on chain upgrade and reverts to the old version so to alleviate this I have to have the name match.  Now there is definitely an argument to say the problem is with cosmovisor or that automatic downloads are discouraged, but my opinion is I should be able to start a node and sync from Genesis (or any other block height) and sync without intervention or failure.  Anyway if I set a different block height and retain the name would that be sufficient?

gregnuj:
I have removed the duplicate version names

MasterPi-2124:
Currently, my strategy is to overwrite the latest version in the list if the name is the same. If not, a new version will be added to the list. The upgrade name is unpredictable and we can not just count on that to handle new upgrades. I have made a version template, and in that, the name is auto-generated from the tag, and what the automation tools like cosmosvisor should check is the tag and block height (if the upgrade is consensus-breaking). I think that the name can be the same, but the height will never be the same, because:

- If the version upgrade is consensus-breaking, you have to provide the block-height param, which is never the same as the previous height.
- If the version upgrade is not CB, the height should not be provided, and overall,  the version is still correct and consistent with the template

PR #2200, opened 2023-05-26
Version template, and update codebase

MasterPi-2124:
This PR proposes a new way of adding a new version to the list when a new version is available. This also helps keep the version template and the codebase still containing all the information specified in the schema.

## Detailed changes:
- Added a manual CI that collects version data from user input, following `chain.schema.json`. This will validate the data, create a new version object and add it to the `versions` list in `chain.json`. A new PR `[UPGRADE] <chain> <version>` will be created: https://github.com/notional-labs/chain-registry/pull/11. The workflow run can be seen [here](https://github.com/notional-labs/chain-registry/actions/runs/5090016419)
- An automated CI that runs every 6 hours to check if a codebase should be updated with the new version or not. This can be achieved by comparing the upgrade block and the current block, and with the standardized version object from the manual CI, all the data from this new version can be automatically added to the codebase. Detailed flow can be seen [here](https://whimsical.com/HQiVN9Y9sLe9x2NtcE4URY)

## Todo
- [ ] Standardize `chain.json` to follow `chain.schema.json`
- [ ] Guide for people about how to add a new version properly
- [ ] Handle both consensus-break and non-consensus-break upgrades
- [ ] Linting JSON files in the same format.


.github/workflows/add_version.yml
```diff
@@ -0,0 +1,198 @@
...
+  workflow_dispatch:
+    inputs:
+      chain:
+        description: "Chain to update"
+        required: true
+        default: ""
+        type: string
+
+      tag-height-proposal:
+        description: "Tag, Height and Proposal. If the upgrade is not consensus break, leave height and proposal to -1."
```
JeremyParish69, line 12: ..is not *consensus-breaking,* leave...
or
...does not break consensus, leave...

JeremyParish69:
PR has been abandoned. Please reopen if you wish to discuss

PR #2204, opened 2023-05-26
Noble Chain Testnet Upgrade Info

boojamya:


testnets/nobletestnet/chain.json
```diff
@@ -76,6 +76,26 @@
...
+        ],
+        "next_version_name": "radon"
+      },
+      {
+        "name": "radon",
+        "tag": "v3.0.0",
+        "height": 645000,
+        "recommended_version": "v3.0.0",
+        "compatible_versions": [
+          "v0.5.0", "v0.5.1"
```
JeremyParish69, line 90: Sorry, I don't see how these compatible versions work with this recommended version. I'm guessing all you need there is v3.0.0

boojamya: Hey @JeremyParish69 , thanks for calling this out but...
There is not any consensus breaking code in `v0.5.0`, `v0.5.1` or `v3.0.0` so you can run any of these versions, it's just that `v3.0.0` is the most up to date.

We jumped from `v0.5.1` to `v3.0.0` because we are aligning the versioning between mainnet and testnet.

JeremyParish69: Ok Thanks for explaining. Can you please add v3.0.0 into the compatible_versions array (I know it's implied but just to be consistent and especially useful here since jumping major versions and not break-consensus is a bit unusual)

boojamya: Ope! I stand corrected. `v3.0.0` was actually a stake breaking change. Updated here. 

This upgrade did not have a name, just a tag.
Thanks @JeremyParish69!

PR #2219, opened 2023-05-28
Update binary version.json

CrazySerGo:
Binary was updated https://github.com/bandprotocol/chain/releases/tag/v2.5.2


JeremyParish69:
Thanks for this. Can we please add this as a new version object so we can retain the v2.4 version data?

CrazySerGo:
> Thanks for this. Can we please add this as a new version object so we can retain the v2.4 version data?

I think so. Please go ahead because I'm not good with github yet.

PR #2223, opened 2023-05-29
Add `ibc` terra2-neutron

clemensgg:


JeremyParish69:
to follow alphabetical order neutron should be chain 1 and terra 2 as chain 2

clemensgg:
i wasn't aware of this and coincidentally had it right in all previous prs lol. thanks for pointing this out. fixed

PR #2236, opened 2023-05-30
Add tenet blockchain

danil-lashin:


tenet/chain.json
```diff
@@ -0,0 +1,102 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "tenet",
+  "status": "live",
+  "network_type": "mainnet",
+  "website": "https://tenet.org/",
+  "pretty_name": "Tenet",
+  "chain_id": "tenet_1559-1",
+  "bech32_prefix": "tenet",
```
JeremyParish69, line 9: Please register this prefix:

tenet/assetlist.json
```diff
@@ -0,0 +1,24 @@
...
+      "description": "The native EVM, governance and staking token of the Tenet",
+      "denom_units": [
+        {
+          "denom": "atenet",
+          "exponent": 0
+        },
+      ],
+      "base": "atenet",
+      "name": "Tenet",
+      "display": "tenet",
```
JeremyParish69, line 15: this display [denom] property is mean to point to a defined denom_unit which relays the number of decimal points to use. If the practically used unit TENET has zero exponent (is indivisible), then the display should be atenet, but I'm guessing from the 'a' that there should be another denom_unit called 'tenet' with exponent 18...

tenet/chain.json
```diff
@@ -0,0 +1,102 @@
...
+    "cosmos_sdk_version": "0.46",
+    "consensus": {
+      "type": "tendermint",
+      "version": "0.34"
+    },
+    "ibc_go_version": "6.1.0",
+    "genesis": {
+      "genesis_url": "https://raw.githubusercontent.com/tenet-org/tenet-mainnet/master/config/genesis.json"
+    },
+    "versions": []
```
JeremyParish69, line 59: Can you please also copy the version data you provided above into an object in this versions array?

tenet/chain.json
danil-lashin: Thanks, the pr has been created: https://github.com/satoshilabs/slips/pull/1560

PR #2242, opened 2023-06-01
update quicksilver to v1.2.13

blacklodge-admin:


quicksilver/chain.json
```diff
@@ -84,6 +84,25 @@
         "cosmwasm_version": "0.29",
         "cosmwasm_enabled": true,
         "ibc_go_version": "5.2.0"
+      },
+      {
+        "name": "v1.2.13",
+        "height": 2148751,
```
blacklodge-admin, line 90: Does it make sense to include this height?  The chain was halted for this upgrade, but it was not state breaking and was not triggered by governance.  2148750 was the halt height used.

joe-bowman: Yep, this looks great thank you.

PR #2252, opened 2023-06-02
chore: bump akash version to v0.22.7

troian:


troian:
@yaruwangway mind approving ci

akash/chain.json
```diff
@@ -27,27 +27,25 @@
...
       "genesis_url": "https://raw.githubusercontent.com/akash-network/net/master/mainnet/genesis.json"
     },
     "versions": [
       {
-        "name": "v0.22.0",
-        "recommended_version": "v0.22.0",
+        "name": "v0.22.7",
+        "recommended_version": "v0.22.7",
         "compatible_versions": [
-          "v0.20.0"
```
JeremyParish69, line 46: Can we please retain and expand the compatible_versions arrays?

troian: at this moment there `v0.22.7` is the only version that is compatible

JeremyParish69: oh, then, can we please create a new version object (so one for v.0.22.0, and another for v.0.22.7), and also include the version name in its own compatible versions array.

dylanschultzie:
closing this due to it being outdated

PR #2256, opened 2023-06-02
June 7th - Increase Juno Fees (Proposal 297)

Reecepbcups:
Should https://www.mintscan.io/juno/proposals/297 pass, this PR should be merged in

- Increases min fee for Juno + steps
- adds IBC atom denom (channe-1) with the new fees as well.

### DIFF
--- a/juno/assetlist.json
+++ b/juno/assetlist.json
@@ -25,6 +25,36 @@
       "coingecko_id": "juno-network"
     },
     {
+      "denom_units": [
+        {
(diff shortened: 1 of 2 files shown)


JeremyParish69:
Thanks for this. Can we please add the new fee token to Juno's assetlist? For example, this is how Carbon did it (will only need to change the channel numbers):
https://github.com/cosmos/chain-registry/blob/f165e38bac4e72bbe598c115403afd815d1d205e/carbon/assetlist.json#L538-L569

PR #2273, opened 2023-06-04
Add Lum v1.4.1 upgrade plus 1.4.2 patch

reversesigh:
Chain upgrade v1.4.1 happened at block 7,740,000 per prop 64. https://www.mintscan.io/lum/proposals/64

Chain is being patched currently and v1.4.2 is the recommended version to use. Chain relaunch expected June 5 10:00 UTC per Discord announcements.

### DIFF
--- a/lumnetwork/chain.json
+++ b/lumnetwork/chain.json
@@ -33,9 +33,10 @@
   },
   "codebase": {
     "git_repo": "https://github.com/lum-network/chain",
-    "recommended_version": "v1.4.0",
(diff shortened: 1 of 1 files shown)


JeremyParish69:
Thanks for this, so should we wait until June 5 or does it not matter and can just merge now (since 'relaunch', to me, implies it might be down currently)?

dylanschultzie:
@JeremyParish69 it's now up, so I'll go ahead and merge now.

PR #2276, opened 2023-06-05
update comdex main net version

dheerajkd30:


dheerajkd30:
> Thanks for this. Although I'm confused about whether it went from v9 to v10, and then somehow back down to v9 as if v10 had never happened, I'd like to ask the we retain v9.0 in the versions array--is that possible?

We never upgraded to v10.*, I saw the the PR where it was update to v10 from v9, @reversesigh might have did it by mistake.

JeremyParish69:
Ah, OK thanks for explaining that. Could we please still keep v9.0 in the versions array so it starts to build a history or versions rather than just updating to the latest each time?

dylanschultzie:
Screenshot 2023-06-15 at 11 28 36 AM
I don't believe this is correct, what was trying to be achieved with this PR @dheerajkd30?

dylanschultzie:
closing this due to it being stale

PR #2316, opened 2023-06-09
update stargaze fee token to the min required

jhernandezb:
This PR updates the min gas price required by the network. Also updates to the latest v10

### DIFF
--- a/stargaze/chain.json
+++ b/stargaze/chain.json
@@ -40,7 +40,15 @@
     },
     "versions": [
       {
-        "name": "v10.0.0",
+        "name": "v9",
+        "recommended_version": "v9.1.0",
+        "compatible_versions": [
+          "v9.1.0",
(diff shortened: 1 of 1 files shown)


stargaze/chain.json
```diff
@@ -26,21 +30,21 @@
...
-      "v9.0.0"
+      "v10.0.1",
+      "v10.0.0"
     ],
     "genesis": {
       "genesis_url": "https://raw.githubusercontent.com/public-awesome/mainnet/main/stargaze-1/genesis.tar.gz"
     },
     "versions": [
       {
-        "name": "v9.0.0",
```
JeremyParish69, line 39: Can we instead keep v9 in the version history array, but then add v10 as a new version object below?

jhernandezb: added v9

PR #2283, opened 2023-06-06
Update Realio Network to v0.8.1

jiujiteiro:
Update chain version to [v0.8.1](https://github.com/realiotech/realio-network/releases/tag/v0.8.1)
Fix chain pretty name to match entry in [slip-0173](https://github.com/satoshilabs/slips/blob/master/slip-0173.md)
Added info on SDK, tendermint and IBC versions

PR #2326, opened 2023-06-11
update binary version - Realio Network

Genz22:


Genz22:
update for https://github.com/cosmos/chain-registry/pull/2283

realio/assetlist.json
```diff
@@ -7,7 +7,7 @@
         "denom_units": [
           {
             "denom": "ario",
-            "exponent": 0
+            "exponent": 18
```
JeremyParish69, line 10: both of these denom units 'ario' and 'arst' should remain as having 0 exponent, and they  are the base units. The corresponding denom units without the 'a' at the front has the 18 exponent. Please change back to 0.

Genz22: hhh yeah sorry

PR #2327, opened 2023-06-12
fix: update Crescent chain.json for latest version

crypin:
update the latest Crescent binary version to [v4.2.0](https://github.com/crescent-network/crescent/releases/tag/v4.2.0)

### DIFF
--- a/crescent/chain.json
+++ b/crescent/chain.json
@@ -46,6 +46,8 @@
     "versions": [
       {
         "name": "v4.2.0",
+        "tag": "v4.2.0",
+        "height": 6915000,
         "recommended_version": "v4.2.0",
         "compatible_versions": [
           "v4.2.0"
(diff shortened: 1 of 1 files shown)


crypin:
@JeremyParish69 Hi thanks for your reply, but v4.2.0 includes breaking change, so I removed v4.0.0 because it was no longer compatible on the mainnet, is there any other reason to keep v4.0.0?

JeremyParish69:
> @JeremyParish69 Hi thanks for your reply, but v4.2.0 includes breaking change, so I removed v4.0.0 because it was no longer compatible on the mainnet, is there any other reason to keep v4.0.0?

Thanks for explaining, but yes we want to keep that in the 'versions' array so we can start building up a version history. Only the latest version would seen as the actual current version.

PR #2351, opened 2023-06-14
Add fxcore chain

zakir-code:


fxcore/assetlist.json
```diff
@@ -0,0 +1,24 @@
...
+  "assets": [
+    {
+      "description": "The native staking token of the Function X",
+      "denom_units": [
+        {
+          "denom": "FX",
+          "exponent": 0
+        }
+      ],
+      "base": "fx",
```
JeremyParish69, line 13: the base must refer to some denom_unit (case-sensitive), and so I'm thinking this should probably be capitalized. But if that actual base unit is fx, then all other cases should be lower-cased (i.e., the display, staking token denom, and fee token denom)

fxcore/chain.json
```diff
@@ -0,0 +1,135 @@
...
+  "staking": {
+    "staking_tokens": [
+      {
+        "denom": "FX"
+      }
+    ]
+  },
+  "codebase": {
+    "git_repo": "https://github.com/functionx/fx-core",
+    "recommended_version": "v4.2.1",
```
JeremyParish69, line 39: Can you please copy these codebase properties (all that you have here except for `git_repo` and `genesis`) into an object within a `versions` array? We are trying to ensure there exists a versions array for all chains and build up a version history for each over time.

fxcore/assetlist.json
JeremyParish69: Thanks. We'll go with this. Just fyi pretty much all other chains have all lower case for the base denom--just thought I'd mention in case we ever see issue with a mismatch.

fxcore/chain.json
```diff
@@ -0,0 +1,213 @@
...
+      "name": "fxv1",
+      "genesis_url": "https://github.com/functionx/fx-core/raw/main/public/mainnet/genesis.json"
+    },
+    "cosmos_sdk_version": "0.46",
+    "consensus": {
+      "type": "tendermint",
+      "version": "0.34"
+    },
+    "ibc_go_version": "6.1.0",
+    "versions": [
```
JeremyParish69, line 60: thanks for adding the versions array. Last request is to correct the order from first version at the top, and latest/current version at the bottom.

Issue #2352, opened 2023-06-14
Images should not have a "minItems" property in JSON schema


giorgionocera:
Hi everyone,
I used the [chain JSON schema](https://github.com/cosmos/chain-registry/blob/master/chain.schema.json) to unmarshal the JSONs retrieved from the repository.

```JSON
{
  "$id": "https://sikka.tech/chain.schema.json",
  "$schema": "https://json-schema.org/draft-07/schema",
  "title": "Cosmos Chain",
  "description": "Cosmos Chain.json is a metadata file that contains information about a cosmos sdk based chain.",
  "type": "object",
  "required": [
    "chain_name",
    "chain_id",
    "bech32_prefix"
  ],
  "properties": {
    "images": {
      "type": "array",
      "items": {
            ...
      },
      "minItems": 1
    },
    "additionalProperties": false,
    "$defs": {
        ...
    }
  }
}
```

In this context, I saw that parameter `minItems` in the `properties.images` field is set to 1 even if many JSON files uploaded to the repo do not have any `images` field.

https://github.com/cosmos/chain-registry/blob/01b706f3f8062d0c1087bcd77917623be094d30b/chain.schema.json#L412

This causes the unmarshal to fail since the requirement is not matched.

My suggestion is to remove the `minItems` parameter since the images field has no minimum element as evidenced by the de facto standard.

Some examples are: 
`agoric` https://github.com/cosmos/chain-registry/blob/master/agoric/chain.json
`bitsong` https://github.com/cosmos/chain-registry/blob/master/bitsong/chain.json
`cosmoshub` https://github.com/cosmos/chain-registry/blob/master/cosmoshub/chain.json
`desmos` https://github.com/cosmos/chain-registry/blob/master/desmos/chain.json
`juno` https://github.com/cosmos/chain-registry/blob/master/juno/chain.json

I'll provide a PR for this.
Let me know if I'm missing something.

PR #2353, opened 2023-06-14
fix: :bug: fix bug in JSON schema

giorgionocera:
Removing the `minItems` attribute from the `properties.images` field for fixing a bug in JSON schema.

Closes cosmos/chain-registry#2352.

### DIFF
--- a/chain.schema.json
+++ b/chain.schema.json
@@ -408,8 +408,7 @@
           }
         },
         "additionalProperties": false
-      },
-      "minItems": 1
+      }
     },
     "logo_URIs": {
       "type": "object",


JeremyParish69:
I think this is fine although I always thought the requirement only applied iff the images property was includes, and would never complain if there were no images property at all.

giorgionocera:
Hey, @JeremyParish69 you made me curious and I looked into it.
It is unclear to me from the [documentation](https://datatracker.ietf.org/doc/html/draft-fge-json-schema-validation-00#section-5.3.3). 
For this reason, I also tried a [validator online](https://www.jsonschemavalidator.net/) and you are right. It only requires at least an image if the image field is defined.
I was using [this tool](https://github.com/omissis/go-jsonschema) for generating the go data type from the JSON Schema definition, so it probably is an issue with that tool.
This commit was wrong.
I could revert it if you want since it does not fix any bugs.

JeremyParish69:
> Hey, @JeremyParish69 you made me curious and I looked into it. It is unclear to me from the [documentation](https://datatracker.ietf.org/doc/html/draft-fge-json-schema-validation-00#section-5.3.3). For this reason, I also tried a [validator online](https://www.jsonschemavalidator.net/) and you are right. It only requires at least an image if the image field is defined. I was using [this tool](https://github.com/omissis/go-jsonschema) for generating the go data type from the JSON Schema definition, so it probably is an issue with that tool. This commit was wrong. I could revert it if you want since it does not fix any bugs.

I appreciate the investigation. I think for now the `images` property isn't used enough to bother requiring restrictions on it just yet at the cost of some tools not working. If it can help some tools work, I suppose we might as well leave it removed. But if you think it isn't a required change, then you are free to propose changing it back, if you like.

giorgionocera:
Thank you. 
Yeah, I prefer opening an issue on the other tool and trying to fix that. Since this repo was following the schema, I think it should be better to revert the commit. I'll push a revert commit and open the PR. 
Thank you for your time 🙏

PR #2358, opened 2023-06-15
Upgrade Composable to Centauri/v3.1.0

reversesigh:
Upgrade Composable to Centauri/v3.0.2. This upgrade changes the binary name from banksyd to centaurid, and also changes the address prefix from banksy to centauri.

Prop 3: https://explorers.l0vd.com/composable-mainnet/gov/3
Height: 188500

In addition this commit adds APIs and a seed node for Lavender.Five Nodes.

### DIFF
--- a/composable/chain.json
+++ b/composable/chain.json
@@ -6,8 +6,8 @@
   "pretty_name": "Composable",
   "status": "live",
   "network_type": "mainnet",
-  "bech32_prefix": "banksy",
-  "daemon_name": "banksyd",
+  "bech32_prefix": "centauri",
+  "daemon_name": "centaurid",
   "node_home": "$HOME/.banksy",
   "key_algos": [
     "secp256k1"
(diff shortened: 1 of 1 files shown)


composable/chain.json
```diff
@@ -6,8 +6,8 @@
   "pretty_name": "Composable",
   "status": "live",
   "network_type": "mainnet",
-  "bech32_prefix": "banksy",
-  "daemon_name": "banksyd",
+  "bech32_prefix": "centauri",
```
JeremyParish69, line 9: I don't know how it slipped by me earlier, but the bech32 prefix aught to be registered to SLIP173 https://github.com/satoshilabs/slips/blob/master/slip-0173.md

reversesigh: Did a PR for it https://github.com/satoshilabs/slips/pull/1568

Composable also decided last minute to change the version for centaurid. Will update my PR in a second.

PR #2361, opened 2023-06-15
Add permtestnet

ahmad-blocksdev:


testnets/permtestnet/assetslist.json
```diff
@@ -0,0 +1,28 @@
...
+    "assets": [
+      {
+        "description": "Perm A Permissioned Blockchain",
+        "denom_units": [
+          {
+            "denom": "ufee",
+            "exponent": 0
+          },
+          {
+            "denom": "uperm",
```
JeremyParish69, line 13: I'm guessing this should be 'perm' and not 'uperm'. Typically the u- prefix is added to act as a 'micro-' prefix, hence taking place of the base unit where the token has 6 exponent. If this were just 'perm', and not 'uperm', it would make the 'display' unit make sense.

testnets/permtestnet/chain.json
```diff
@@ -0,0 +1,113 @@
...
+    "daemon_name": "permd",
+    "node_home": "$HOME/.perm",
+    "key_algos": [
+      "secp256k1"
+    ],
+    "slip44": 118,
+    "fees": {
+      "fee_tokens": [
+        {
+          "denom": "uperm",
```
JeremyParish69, line 18: these denoms must be the base_denom, so I think it should be 'ufee'

testnets/permtestnet/chain.json
```diff
@@ -0,0 +1,113 @@
...
+          "low_gas_price": 0.0025,
+          "average_gas_price": 0.0025,
+          "high_gas_price": 0.0024
+        }
+      ]
+    },
+    "staking": {
+      "staking_tokens": [
+        {
+          "denom": "uperm"
```
JeremyParish69, line 28: these denoms must be the base_denom, so I think it should be 'ufee'

ahmad-blocksdev (commented):
uperm is the base_denom

testnets/permtestnet/assetslist.json
```diff
@@ -0,0 +1,28 @@
...
+            "denom": "perm",
+            "exponent": 6
+          }
+        ],
+        "base": "uperm",
+        "name": "Perm",
+        "display": "perm",
+        "symbol": "PERM",
+        "logo_URIs": {
+          "png": "",
```
JeremyParish69, line 22: actually, some images here would be ideal or else probably just removing the fields.

PR #2367, opened 2023-06-17
Add regen nct assetlist and images

dpdanpittman:


JeremyParish69:
What is the purpose of this? Looks like it break it. We need a base denom with 0 exponent....

dpdanpittman:
This could actually be a leftover PR.  I'm going to close it.

PR #2375, opened 2023-06-19
chore: Mintscan not picking up /ixo/info details - now added

alwyn-ixo:


JeremyParish69:
We already have this under the impacthub/ [folder](https://github.com/cosmos/chain-registry/tree/master/impacthub). The chain_name must match the folder name.

alwyn-ixo:
@JeremyParish69 Hi - thx for the review.
The chain id changed from impacthub to ixo in March.
At the time we agreed that too many other parties may have dependencies to the impacthub folder.
However, our main interest is to have Mintscan pick up this configuration, which it is not.
https://www.mintscan.io/ixo

JeremyParish69:
> @JeremyParish69 Hi - thx for the review. The chain id changed from impacthub to ixo in March. At the time we agreed that too many other parties may have dependencies to the impacthub folder. However, our main interest is to have Mintscan pick up this configuration, which it is not. https://www.mintscan.io/ixo

I see. We should let Mintscan know where to look since the chain name they use won't always align with the folder name.

alwyn-ixo:
@JeremyParish69 
Notified Cosmostation and they updated on their side.
All is well - thx!

PR #2378, opened 2023-06-19
Update Cro, Kuji, and Osmo versions

Golden-Ratio-Staking:


cryptoorgchain/chain.json
```diff
@@ -31,33 +31,33 @@
   },
   "codebase": {
     "git_repo": "https://github.com/crypto-org-chain/chain-main",
-    "recommended_version": "v4.2.2",
+    "recommended_version": "v4.2.6",
     "compatible_versions": [
-      "v4.2.2"
+      "v4.2.6"
```
JeremyParish69, line 36: Assuming this patch is not consensus-breaking against v4.2.2, can we please keep v4.2.2 as a compatible version here (and also in the corresponding version object in the `versions` array)

cryptoorgchain/chain.json
```diff
@@ -31,33 +31,33 @@
...
+      "windows/amd64": "https://github.com/crypto-org-chain/chain-main/releases/download/v4.2.6/chain-main_4.2.6_Windows_x86_64.zip"
     },
     "genesis": {
       "genesis_url": "https://github.com/crypto-org-chain/mainnet/raw/main/crypto-org-chain-mainnet-1/genesis.json"
     },
     "versions": [
       {
-        "name": "v4.2.2",
-        "recommended_version": "v4.2.2",
+        "name": "v4.2.6",
```
JeremyParish69, line 50: Perhaps the name 'v4.2' would be more appropriate as it is generalized for both v4.2.2 and v4.2.6.

kujira/chain.json
```diff
@@ -33,19 +33,19 @@
   },
   "codebase": {
     "git_repo": "https://github.com/Team-Kujira/core",
-    "recommended_version": "v0.8.4-mainnet",
+    "recommended_version": "v0.8.7",
     "compatible_versions": [
-      "v0.8.4-mainnet"
+      "v0.8.7"
```
JeremyParish69, line 38: again, if this is compatible, please keep v0.8.4-mainnet in the list of compatible versions.

Golden-Ratio-Staking: In this instance we found out (with a fun case of non-determinism) that this patch was consensus breaking for kujira. v0.8.7 is the current and only compatible version. The previous version was actually v0.8.6 and will apphash.

cryptoorgchain/chain.json
Golden-Ratio-Staking: That's fine by me, but do we want to even suggest that full node versions with security vulnerabilities are compatible?

cryptoorgchain/chain.json
dylanschultzie: @JeremyParish69 I think we need to take this offline, but yeah, changing name doesn't make sense.

kujira/chain.json
```diff
@@ -33,19 +33,19 @@
...
     ],
     "genesis": {
       "genesis_url": "https://github.com/Team-Kujira/networks/raw/master/mainnet/kaiyo-1.json"
     },
     "versions": [
       {
-        "name": "v0.8.4-mainnet",
-        "recommended_version": "v0.8.4-mainnet",
+        "name": "v0.8.7",
```
dylanschultzie, line 45: this name shouldn't be changed, the `name` is the cosmovisor upgrade name

dylanschultzie (changes requested):

cryptoorgchain/chain.json
JeremyParish69: That's a very good point, and there's no strong standard on that at this point. The simple answer is: yes, still compatible. An alternative off the top of my head is: to create a new version object that starts at the secure version, and point the previous and use the `next_version_name` property (which forms a linked list) to essentially skip the version with the vulnerability. Really, it's a topic that should be discussed further.

kujira/chain.json
JeremyParish69: Thanks for explaining. In cases where the new version is consensus-breaking, please instead create a new version object at the end of the `versions` array (so we build up a history of versions) instead of just updating the latest one.

kujira/chain.json
Golden-Ratio-Staking: thank you for clarifying that!

Golden-Ratio-Staking:
Closing until there's some consensus on versioning for cryptoorgchain.

PR #2389, opened 2023-06-20
Update Kujira Versions

Golden-Ratio-Staking:
- Cosmovisor folder `name` is `v0.8.4`. It was never `v0.8.4-mainnet`, that was just a hotfix release version.
- `v0.8.7` is the current and only compatible version.

### DIFF
--- a/kujira/chain.json
+++ b/kujira/chain.json
@@ -33,19 +33,19 @@
   },
   "codebase": {
     "git_repo": "https://github.com/Team-Kujira/core",
-    "recommended_version": "v0.8.4-mainnet",
+    "recommended_version": "v0.8.7",
     "compatible_versions": [
-      "v0.8.4-mainnet"
+      "v0.8.7"
     ],
     "genesis": {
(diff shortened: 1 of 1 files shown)


JeremyParish69:
Thanks for this. It does seem that this was implies that v0.8.7 block height matches v0.8.4, which it does not. I'd like to urge you to follow something like what I have here (please see codebase section): https://github.com/JeremyParish69/chain-registry/blob/upd-kujira-version/kujira/chain.json

Golden-Ratio-Staking:
v0.8.7 *should* work from the previous cosmovisor upgrade height (v0.8.4)

v0.8.7 was a rolling security upgrade and shouldn't have been state breaking, but there was an event in a transaction that made v0.8.4 & v0.8.6 incompatible and apphashed.

At this point I feel like waiting for v0.9.0 anticipated upgrade is a fresh start to wait for instead.

JeremyParish69:
Closed for https://github.com/cosmos/chain-registry/pull/2406

PR #2394, opened 2023-06-21
Update chain info 

josietyleung:


composable/assetlist.json
```diff
@@ -1,9 +1,9 @@
 {
     "$schema": "../assetlist.schema.json",
-    "chain_name": "composable",
+    "chain_name": "centauri",
```
JeremyParish69, line 3: the chain_name must remain as 'composable' matching the folder name.

composable/chain.json
```diff
@@ -1,9 +1,9 @@
 {
   "$schema": "../chain.schema.json",
-  "chain_name": "composable",
+  "chain_name": "centauri",
```
JeremyParish69, line 3: the chain_name must remain as 'composable' matching the folder name.

composable/assetlist.json
josietyleung: ah, is the folder name updatable?

JeremyParish69: for live chains, we avoid that since it could break dependencies already established to look at it.

composable/chain.json
josietyleung: hi @JeremyParish69 kept both as 'composable' thanks

PR #2396, opened 2023-06-21
Add neutron-osmosis

luisqagt:


JeremyParish69:
I think it's just missing the .json extension

PR #2399, opened 2023-06-22
added kava

danbryan:
https://github.com/DefiantLabs/svc-kujira-relay#kava


JeremyParish69:
to follow alphabetical order, please make kava chain 1, and kujira chain 2. You have the title right, but flipped in the file.

JeremyParish69:
and just want to double check, because it looks like that last PR swapped the connection #, but not the channel numbers. If channel-95 is the kava channel, then we should be good.

danbryan:
whoops had channels on wrong side, should be good now

PR #2400, opened 2023-06-23
persistence test-core-2

mkRuby009:
test-core-2 chain details added at testnets/persistencetestnet2


testnets/persistencetestnet2/assetlist.json
```diff
@@ -0,0 +1,51 @@
+{
+    "$schema": "../assetlist.schema.json",
+    "chain_name": "persistencetestnet",
```
JeremyParish69, line 3: should be persistencetestnet2

mkRuby009: done

PR #2402, opened 2023-06-23
Delete testnets/permtestnet directory 

ahmad-blocksdev:
Due to some organizations issue, have to delete this testnet. kindly remove the directory completely, if possible.


JeremyParish69:
Can we instead change status to killed, please?

ahmad-blocksdev:
What's the difference between Kill and remove the directory? would you mind to explain?

JeremyParish69:
changing to 'killed' just retains the history so we aren't left wondering what a certain chain or token was and being unable to find any trace of it in the registry. We are already seeing have cases of tokens from killed chains floating around on other chains, and retaining the data helps identify what they are.

ahmad-blocksdev:
> changing to 'killed' just retains the history so we aren't left wondering what a certain chain or token was and being unable to find any trace of it in the registry. We are already seeing have cases of tokens from killed chains floating around on other chains, and retaining the data helps identify what they are.

Yes. you can change the status to killed.

PR #2407, opened 2023-06-24
Upgrade Stride to v11

reversesigh:
Prop: 207
Height: 4291924
https://www.mintscan.io/stride/proposals/207

Estimated Target Date:
Tue Jun 27 2023 09:46:42 GMT-0400 (Eastern Daylight Time)

### DIFF
--- a/stride/chain.json
+++ b/stride/chain.json
@@ -33,9 +33,9 @@
   },
   "codebase": {
     "git_repo": "https://github.com/Stride-Labs/stride",
-    "recommended_version": "v10.0.0",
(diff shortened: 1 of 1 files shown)


stride/chain.json
```diff
@@ -83,7 +83,16 @@
         "compatible_versions": [
           "v10.0.0"
         ],
-        "next_version_name": ""
+        "next_version_name": "v11.0.0"
```
JeremyParish69, line 86: This should reference the name property, so just 'v11'

PR #2414, opened 2023-06-26
Upgrade Carbon to v2.26.0

reversesigh:
Prop 305
Height 43691091

https://scan.carbon.network/governance/proposal/305?net=main

Est Time: around Monday, July 3, 2023 3:20 AM EST


reversesigh:
Team just announced they are canceling upgrade and will be doing a new prop for a new binary. I'll just update this PR when they announce with new info?

reversesigh:
This is live and ready to go (v2.26.1)

reversesigh:
Hold off on this again - there was a chain halt last night and the binary was upgraded to v2.26.3. But they haven't moved the binaries beyond pre-release yet so I'll wait to update again until they do that.

PR #2415, opened 2023-06-26
Create composable-osmosis

josietyleung:


_IBC/composable-osmosis.json
```diff
@@ -0,0 +1,32 @@
...
+    },
+    "channels": [
+      {
+        "chain_1": {
+          "channel_id": "channel-3",
+          "port_id": "transfer"
+        },
+        "chain_2": {
+          "channel_id": "channel-641",
+          "port_id": "channel-1279"
```
JeremyParish69, line 21: this port_id should be 'transfer', right?

josietyleung: ah that's right. updated thanks!

PR #2425, opened 2023-06-27
Remove injectivetestnet RPC endpoint that's not resolving

mikedotexe:
Saw an RPC endpoint that wasn't resolving. Also let the team know in case they want to chime in or recommend we keep this. Just tryna do my part 👼🏼


mikedotexe:
it's fixed! closing

PR #2426, opened 2023-06-27
chore: update Injective

achilleas-kal:


injective/chain.json
```diff
@@ -48,14 +45,14 @@
     },
     "versions": [
       {
-        "name": "v1.10",
```
JeremyParish69, line 51: Instead of writing over the v1.10 version object, can we instead update the v1.11 object?

testnets/injectivetestnet/chain.json
```diff
@@ -33,33 +33,39 @@
...
-      "darwin/amd64": "https://github.com/InjectiveLabs/testnet/releases/download/v1.10.2-1678712142/darwin-amd64.zip"
+      "linux/amd64": "https://github.com/InjectiveLabs/testnet/releases/download/v1.11.1-1685205489/linux-amd64.zip",
+      "darwin/amd64": "https://github.com/InjectiveLabs/testnet/releases/download/v1.11.1-1685205489/darwin-amd64.zip"
     },
     "genesis": {
       "genesis_url": "aws s3 cp s3://injective-snapshots/testnet/genesis.json . --no-sign-request"
     },
     "versions": [
       {
-        "name": "v1.10.2",
```
JeremyParish69, line 49: Instead of writing over the v1.10 version object, can we instead create a v1.11 object, please?

achilleas-kal:
@JeremyParish69 removed one of the objects and just left v1.11.5

JeremyParish69:
Hi @achilleas-kal , sorry, I'm not sure why the mainnet chain file is being modified; it looks like you're just trying to remove data that we want to keep. Unless I'm missing something, I'd ask to not try to remove anything there.
For the testnet chain.json, we want to add v1.11.1 as a new version object in the versions array instead of getting rid of v1.10.2 version object.

JeremyParish69:
> @JeremyParish69 the v1.10.2 version is no longer compatible though, so at a given time you keep the previous version, the current version and also outline what the next version is going to be?

Yes, we are striving to keep a full history of versions (ideally from v1 all the way to current), and it's only the versions listed under `compatible_versions` in the current version that are understood to be actually compatible. This is why we want to keep v1.10.2--it's only for historical data

achilleas-kal:
@JeremyParish69 noted, thanks for the explanation. I'll be modifying the new ones accordingly.

JeremyParish69:
Should I jump in and rearrange the versions so that it maintains the history but also adds the new ones? We don't want to merge as-is...

achilleas-kal:
Based on this, v1.10 is still here and operators can upgrade to v1.11.5 directly so I have both versions in the PR. What's the issue exactly, do you want to keep v1.11 and also add v1.11.5? 

Yes, please feel free to modify and I can refer to your changes for the next one @JeremyParish69 

```
      {
        "name": "v1.10",
        "recommended_version": "v1.10",
        "compatible_versions": [
          "v1.10"
        ],
        "binaries": {
          "linux/amd64": "https://github.com/InjectiveLabs/injective-chain-releases/releases/download/v1.10.0-1679065799/linux-amd64.zip",
          "darwin/amd64": "https://github.com/InjectiveLabs/injective-chain-releases/releases/download/v1.10.0-1679065799/darwin-amd64.zip"
        },
        "next_version_name": "v1.11"
      },
      {
        "name": "v1.11.5",
        "recommended_version": "v1.11.5",
        "compatible_versions": [
          "v1.11.5"
        ],
        "binaries": {
          "linux/amd64": "https://github.com/InjectiveLabs/injective-chain-releases/releases/download/v1.11.5-1687535916/linux-amd64.zip",
          "darwin/amd64": "https://github.com/InjectiveLabs/injective-chain-releases/releases/download/v1.11.5-1687535916/darwin-amd64.zip"
        },
        "next_version_name": "v1.12"
      }
    ]
  },
```

JeremyParish69:
I believe this PR branch's current state (after the changes I've made) should be accurate to the upgrade, because it preserves the version history, but also specifies the current version. Does this look acceptable?

PR #2433, opened 2023-06-28
delete VinceChain

Galadrin:
All the links to RPC/REST/Explorer are dead


vincechain/chain.json
```diff
@@ -1,101 +0,0 @@
-{
-  "$schema": "../chain.schema.json",
-  "chain_name": "vincechain",
-  "status": "live",
```
JeremyParish69, line 4: can we please instead just set the status to 'killed'?

JeremyParish69:
Replaced with https://github.com/cosmos/chain-registry/pull/3247

PR #2447, opened 2023-07-02
chore: update archway testnet

whalelephant:
- Aligning chain info with [chain config](https://github.com/archway-network/networks/blob/main/constantine-3/chain.json)
- Updating asset info


testnets/archwaytestnet/assetlist.json
```diff
@@ -1,20 +1,20 @@
 {
   "$schema": "../assetlist.schema.json",
-  "chain_name": "archwaytestnet",
+  "chain_name": "archway",
```
JeremyParish69, line 3: archway is reserved for mainnet, thus it remain as 'archwaytestnet'. Not sure if it is possible to align with the chain config exactly because of this. I'd recommend updating both as necessary.

JeremyParish69:
It seems this file has been update since, so perhaps resolving the conflicts might reveal what remains

PR #2452, opened 2023-07-03
updated constantine-3 info

lufa23:
- updated gas prices
- updated denom


dylanschultzie:
please resolve the conflicts

testnets/archwaytestnet/chain.json
```diff
@@ -16,10 +16,10 @@
   "fees": {
     "fee_tokens": [
       {
-        "denom": "uconst",
-        "low_gas_price": 0,
-        "average_gas_price": 0.025,
-        "high_gas_price": 0.05
+        "denom": "aconst",
```
JeremyParish69, line 19: thanks for the correction. If this really is the case, then can we please correct the assetlist.json--which still stays uconst--as well?

PR #2465, opened 2023-07-04
ADD _IBC/gravitybridge-cheqd.json

keupsonite:


JeremyParish69:
The error is from alphabetical check. These files must follow alphabetical order--i.e., in this case, file must be titled cheq-gravitybridge... and cheqd must be chain_1, and grav as chain_2..

keupsonite:
@dylanschultzie are you sure about the channel-43? Should be -2

PR #2456, opened 2023-07-03
add archway mainnet info

lufa23:
- add images
- add asset.json
- add chain.json

PR #2466, opened 2023-07-04
Update chain.json for archway mainnet

lufa23:
- Updated the chain.json

- Some rpc endpoints had been removed as they were not active


JeremyParish69:
Please see the comments from this PR: https://github.com/cosmos/chain-registry/pull/2456

jlehtimaki:
> Please see the comments from this PR: #2456

@JeremyParish69 
Hey, we have seen the comments and addressed those ones which we think is valid in this PR.
We don't want to have untested endpoints so we removed the ones that were not working.
We will add more versions and compatibility on the next releases and when there's a need for it.
 
 Thank you.

JeremyParish69:
> > Please see the comments from this PR: #2456
> 
> @JeremyParish69 Hey, we have seen the comments and addressed those ones which we think is valid in this PR. We don't want to have untested endpoints so we removed the ones that were not working. We will add more versions and compatibility on the next releases and when there's a need for it.
> 
> Thank you.

Thanks for the response, but there are many other comments which are all valid. I'd have the make the same comments here on this PR

archway/assetlist.json
```diff
@@ -1,28 +1,32 @@
...
-            "exponent": 18
-          }
-        ],
-        "base": "aarch",
-        "name": "Archway",
-        "display": "archway",
-        "symbol": "ARCH",
-        "logo_URIs": {
+  "$schema": "../assetlist.schema.json",
+  "chain_name": "archway-1",
```
JeremyParish69, line 3: chain_name is 'archway'

archway/chain.json
```diff
@@ -29,42 +29,34 @@
...
+    "cosmwasm_enabled": true,
+    "ibc_go_version": "v4.3.1",
     "genesis": {
-      "genesis_url": "https://github.com/archway-network/networks/blob/main/archway-1/genesis/genesis.json.gz"
+      "genesis_url": "https://github.com/archway-network/networks/raw/main/archway-1/genesis/genesis.json.gz"
     },
     "versions": [
       {
-        "name": "genesis",
-        "recommended_version": "v1.0.0",
```
JeremyParish69, line 55: Can you please re-add this stuff (from recommended_version to binaries) back in please?

JeremyParish69: FYI I added it in

PR #2480, opened 2023-07-07
Add supported IBC assets on Nolus

metodi96:


nolus/assetlist.json
```diff
@@ -23,6 +23,286 @@
...
+            "channel_id": "channel-783"
+          },
+          "chain": {
+            "channel_id": "channel-0",
+            "path": "transfer/channel-0/transfer/channel-208/uusdc"
+          }
+        }
+      ],
+      "logo_URIs": {
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/nolus/images/usdc.svg",
```
JeremyParish69, line 63: Can I encourage to instead point the url to the respective origin chains that contain the images? e.g., https://raw.githubusercontent.com/cosmos/chain-registry/master/**_non-cosmos/ethereum**/images/usdc.svg, and not to upload duplicate images?

This way we don't need to store extra versions of images. I'd recommend only uploading more images if Nolus wants to override to original images.

In fact, image URLs here I'd consider optional, since the trace will point to the origin asset, which has the logo_URIs property.

metodi96: Done

PR #2482, opened 2023-07-07
VinceChain Move To Testnet

AyrisDev:


JeremyParish69:
Can I ask, why is this moving the chain instead of simply adding a new chain folder for the testnet?

AyrisDev:
frankly, the mainnet is not ready, so I didn't want people to be wrong. For now, our transactions are continuing on the testnet.

AyrisDev:
> Can I ask, why is this moving the chain instead of simply adding a new chain folder for the testnet?

frankly, the mainnet is not ready, so I didn't want people to be wrong. For now, our transactions are continuing on the testnet.

PR #2512, opened 2023-07-12
Add OnFinality RPC Endpoint for Kava to chain.json

deverka:


ankurdotb (changes requested):
The structure is fine, but if I hit https://kava.api.onfinality.io/public/status I get the following error message:

`{"jsonrpc":"2.0","id":"0","error":{"code":-32603,"message":"error processing the request: Error parsing relay: Error unmarshalling relay as json"}}`

Is this RPC endpoint functional?

deverka:
Yes, it is functional. Let me cross check with the team

hugh-onf:
@ankurdotb this endpoint only supports EVM RPC
`curl -X POST https://kava.api.onfinality.io/public -d '{"id": 1,"jsonrpc": "2.0","method": "eth_blockNumber"}' -H 'Content-Type: application/json'`

image
Would it be an issue?

Thanks.

kava/chain.json
```diff
@@ -113,6 +113,10 @@
       {
         "address": "https://rpc-kava-01.stakeflow.io",
         "provider": "Stakeflow"
+      },
+      {
+        "address": "https://kava.api.onfinality.io/public",
```
dylanschultzie, line 118: I believe this should be moved to line 171 under `evm`

dylanschultzie (changes requested):

deverka:
PR udpated! Sorry for a late response.

PR #2547, opened 2023-07-20
Add nois IBC channels

kaisbaccour:


testnets/_IBC/stargazetestnet-noistestnet.json
```diff
@@ -0,0 +1,47 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "stargazetestnet",
+    "client_id": "07-tendermint-483",
+    "connection_id": "connection-500"
+  },
+  "chain_2": {
+    "chain_name": "noistestnet",
```
JeremyParish69, line 9: note that the chain_! and chain_2 must be in alphabetical order, so noistestnet should be chain 1 in this case

kaisbaccour: Oh good to know, I will fix it now

kaisbaccour: @JeremyParish69 it's fixed now

PR #2556, opened 2023-07-24
updated axerlar path

lufa23:
- updated to the new IBC path for axelar

### DIFF
--- a/_IBC/archway-axelar.json
+++ b/_IBC/archway-axelar.json
@@ -2,22 +2,22 @@
     "$schema": "../ibc_data.schema.json",
     "chain_1": {
         "chain_name": "archway",
-        "client_id": "07-tendermint-3",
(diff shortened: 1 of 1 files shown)


JeremyParish69:
Is there any governance approval for this change? Or can we show there have been no funds transferred with the old connection? Otherwise, switching to a new canonical route can expire many users' funds.

shahbazn:
@JeremyParish69 the change in channel was done in coordination with Axelar team to comply with recommended parameters. The new channel is already being actively used and communicated to Archway community.

There was however a small amount of funds that moved through the old channel before we switched over the relayers to the new channel. We are keeping client alive on the old channel to service any user requests to move the funds back. Having the old channel still published on the registry presents the risk of user trying to send more funds on an outdated channel.

PR #2557, opened 2023-07-24
Add desmos testnet config

giorgionocera:
Adding the Desmos "morpheus-apollo-3" testnet configurations.


testnets/desmostestnet/chain.json
```diff
@@ -0,0 +1,119 @@
...
+  "staking": {
+    "staking_tokens": [
+      {
+        "denom": "udaric"
+      }
+    ]
+  },
+  "codebase": {
+    "git_repo": "https://github.com/desmos-labs/desmos",
+    "recommended_version": "v5.0.0",
```
JeremyParish69, line 35: based on the versions list, shouldn't this (and the compatible_versions list) be v5.2.0 ?

giorgionocera: It seems that on the testnet, the [upgrade proposal to v5.2.0](https://testnet.bigdipper.live/desmos/it/proposals/47) did not pass: https://testnet.bigdipper.live/desmos/it/proposals.
What do you suggest to do in this scenario?

JeremyParish69: In this case I'd suggest removing v5.2.0 object from the versions array, and this 'recommended version should be whatever the current version is (v5.1.0?)

giorgionocera: Okay, that sounds very reasonable. I'll provide the correct implementation 🚀

giorgionocera: I provided the modification from your suggestion following the previous version of the desmos (not demostestnet) configuration file (including the v5.2.0) [here](https://github.com/cosmos/chain-registry/blob/b26c994570f11209b8f6f1155d53e3610fc2f47a/desmos/chain.json).
Do you think it is okay now?

PR #2564, opened 2023-07-25
Update Provenance registration to reflect current version 1.16.0

iramiller:
House keeping for Provenance chain registry entry:
- Updated recommended versions to match current mainnet (1.16.0).
- Updated Figure seed node address and identifier to match current running instance.
- Removed discontinued BigDipper instance hosted by Provenance

### DIFF
--- a/provenance/chain.json
+++ b/provenance/chain.json
@@ -33,13 +33,12 @@
   },
   "codebase": {
     "git_repo": "https://github.com/provenance-io/provenance",
-    "recommended_version": "v1.14.1",
+    "recommended_version": "v1.16.0",
     "compatible_versions": [
-      "v1.14.1",
-      "v1.14.0"
+      "v1.16.0"
     ],
     "binaries": {
-      "linux/amd64": "https://github.com/provenance-io/provenance/releases/download/v1.14.1/provenance-linux-amd64-v1.14.1.zip"
+      "linux/amd64": "https://github.com/provenance-io/provenance/releases/download/v1.16.0/provenance-linux-amd64-v1.16.0.zip"
     },
     "genesis": {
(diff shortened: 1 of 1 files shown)


provenance/chain.json
```diff
@@ -33,29 +33,27 @@
...
     },
     "genesis": {
       "genesis_url": "https://raw.githubusercontent.com/provenance-io/mainnet/main/pio-mainnet-1/genesis.json"
     },
     "cosmwasm_enabled": true,
     "versions": [
       {
-        "name": "v1.14.1",
-        "recommended_version": "v1.14.1",
+        "name": "v1.16.0",
```
JeremyParish69, line 49: Please do not remove the v1.14.1 version, but please create a new version below for v.1.16.0

iramiller: Would you prefer that 1.15.0, 1.15.1 were added then? Because 1.14.x has been superseded for some time now.  Alternatively we could include 1.13-1.16 which would cover the range of versions used in the last 10 months.

JeremyParish69: Yes, the more data the better, and utilizing the 'next_version_name' property really helps establish a linked list of version. I just hope it's not too inconvenient for you to add

iramiller: Not a big deal to do it ... in fact I can source a complete history here and submit that if that is what you are going for: 
https://explorer.provenance.io/network/upgrades

iramiller: Added a bunch of the new detailed version data as well as all of the chain version history going back to genesis.  Hopefully that's useful to someone at some point... seems like it would be possible to do an automated play forward from genesis using this approach if the right archive node could be found.

JeremyParish69: Absolutely fantastic addition. So many people struggle to comprehend what the versions array is meant to be, but I think you got it.

PR #2577, opened 2023-07-28
Fork Passage to v2.0.0

reversesigh:
This is a fork from passage-1 to passage-2 with a new genesis.

Chain relaunch is on Monday July 31

### DIFF
--- a/passage/chain.json
+++ b/passage/chain.json
@@ -2,7 +2,7 @@
   "$schema": "../chain.schema.json",
   "chain_name": "passage",
   "chain_id": "passage-2",
-  "pre_fork_chain_name": "passage-1",
+  "pre_fork_chain_name": "passage1",
   "pretty_name": "Passage",
   "status": "live",
   "network_type": "mainnet",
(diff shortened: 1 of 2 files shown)


JeremyParish69:
any thought on creating a folder for /passage1/ with a copy of the same files but with the original genesis and old version history and status 'killed'?

reversesigh:
> any thought on creating a folder for /passage1/ with a copy of the same files but with the original genesis and old version history and status 'killed'?

Something like this? https://github.com/reversesigh/chain-registry/tree/passage1/passage1

passage/chain.json
```diff
@@ -1,7 +1,8 @@
 {
   "$schema": "../chain.schema.json",
   "chain_name": "passage",
-  "chain_id": "passage-1",
+  "chain_id": "passage-2",
+  "pre_fork_chain_name": "passage-1",
```
JeremyParish69, line 5: I'd say pre_frok_chain_name would be passage1; i.e., the name, not the id
If used, it should be added to the schema

In your passage1 branch, I'd change the chain_name to 'passage1'

reversesigh: https://github.com/cosmos/chain-registry/pull/2598

JeremyParish69:
Are you going to merge your passage1 branch into this patch-49 branch so it caqn all be taken care of in a single PR?

PR #2593, opened 2023-08-01
Starname chain - Update version and add new endpoint providers

dsmello:
EDIT: starname/chain.json
- add one more rpc and the rest endpoint
- add one more persistent peer
- change the chain versions to the right one


starname/chain.json
```diff
@@ -30,21 +30,19 @@
...
+      "v0.11.7"
     ],
     "genesis": {
       "genesis_url": "https://gist.githubusercontent.com/davepuchyr/6bea7bf369064d118195e9b15ea08a0f/raw/genesis.json"
     },
     "versions": [
       {
-        "name": "v0.10.18",
-        "recommended_version": "v0.10.18",
+        "name": "v0.11.7",
```
JeremyParish69, line 42: within the `versions` array, can we please keep v0,10,18 where it is, but then add v0.11.7 as a new object below? We're trying to retain a version history with this

PR #2597, opened 2023-08-01
Update passage chain id and recommended binary version

LeonoorsCryptoman:


JeremyParish69:
Thanks for this, but it's already being taken care of in a certain way. Should be soon
https://github.com/cosmos/chain-registry/pull/2577

PR #2598, opened 2023-08-01
Move passage-1 information to passage1

reversesigh:
Discussion for reference: https://github.com/cosmos/chain-registry/pull/2577

### DIFF
--- a/passage1/chain.json
+++ b/passage1/chain.json
@@ -1,7 +1,7 @@
 {
   "$schema": "../chain.schema.json",
-  "chain_name": "passage",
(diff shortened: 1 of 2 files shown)


passage1/chain.json
```diff
@@ -0,0 +1,64 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "passage",
+  "chain_id": "passage1",
```
JeremyParish69, line 4: Sorry if I wasn't clear. chain_name should be passage1, and chain_id should remain as passage-1

PR #2604, opened 2023-08-02
update chain.json

CosmosSpace2:


CosmosSpace2:
> Please use better PR titles going forward.

How should I name better PRs where I update chain.json?

JeremyParish69:
Every chain has a chain.json. So maybe a chain name, and what property you're adding to. E.g., "Update Cosmos Spaces RPCs on 6 Chains"?

PR #2605, opened 2023-08-02
Add Statemine USDT 

josietyleung:


_non-cosmos/statemine/assetlist.json
```diff
@@ -0,0 +1,32 @@
...
+      ],
+      "base": "130",
+      "name": "Statemine",
+      "display": "usdt",
+      "symbol": "USDT",
+      "traces": [
+        {
+          "type": "bridge",
+          "counterparty": {
+            "chain_name": "statemine",
```
JeremyParish69, line 24: the bridge trace from itself (from statemine to statemine) doesn't make sense....

_non-cosmos/picasso/assetlist.json
```diff
@@ -77,6 +77,32 @@
...
+      "display": "usdt",
+      "symbol": "USDT",
+      "traces": [
+        {
+          "type": "bridge",
+          "counterparty": {
+            "chain_name": "statemine",
+            "base_denom": "130"
+          },
+          "provider": "Statemine"
```
JeremyParish69, line 103: I think we've been using something like Kusama Parachain or Kusama Relay for these transfers

bitcoinfinancier (approved):

_non-cosmos/statemine/assetlist.json
josietyleung: ah updated

_non-cosmos/picasso/assetlist.json
josietyleung: that was for KSM which came from the Kusama relay chain. Kusama relay chain is connected to Picasso. 
Statemine is also directly connected to Picasso so USDT will transfer directly from Statemine > Picasso

PR #2621, opened 2023-08-05
Add DGL image file via upload to Juno registry

CCrafter15:


JeremyParish69:
Why is the registry interested in this image? Is it for an asset?

CCrafter15:
Yes, I would like to add DGL to the chain registry.  I just added the asset details to the asset list.  Thank you

PR #2624, opened 2023-08-05
add Aura-cosmoshub and update Aura-axelar

ertemann:
reference: https://docs.aura.network/validator/relayer

### DIFF
--- a/_IBC/aura-cosmoshub.json
+++ b/_IBC/aura-cosmoshub.json
@@ -2,22 +2,22 @@
     "$schema": "../ibc_data.schema.json",
     "chain_1": {
       "chain_name": "aura",
-      "client_id": "07-tendermint-17",
-      "connection_id": "connection-11"
+      "client_id": "07-tendermint-24",
+      "connection_id": "connection-14"
     },
     "chain_2": {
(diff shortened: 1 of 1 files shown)


_IBC/aura-axelar.json
```diff
@@ -2,22 +2,22 @@
...
     },
     "channels": [
       {
         "chain_1": {
-          "channel_id": "channel-1",
+          "channel_id": "channel-4",
           "port_id": "transfer"
         },
         "chain_2": {
-          "channel_id": "channel-74",
```
JeremyParish69, line 20: Do we have any proof that there are no stranded tokens that have been transferred on the original connection? or that both chains have voted to switch channels? or that the original connection should never have been registered?
The protect users' transferred funds, the registered canonical IBC connection cannot be changed unless voted by governances.

ertemann: The original connection had an incorrect unbonding time, it was not used as that makes the channel itself unusable already.

It was changed within 1-2 hours of the original one being created, just never updated here.

The original one shouldnt have been registered here eventhough it was considered the correct one for a very short time it didnt have funds travel over it.

PR #2629, opened 2023-08-07
Add Archway <> JUNO IBC

Rizbe:


_IBC/archway-juno.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "../ibc_data_schema.json",
+  "chain_1": {
+    "chain_name": "archway-1",
```
JeremyParish69, line 4: there chain ids need to be changed to chain names. You can just remove the '-1'

Rizbe:
@JeremyParish69 
Noticed here looking at JUNO<>OSMO: https://github.com/cosmos/chain-registry/blob/614bb87ba1daff10b6e0e6de1fcb56d8b57b9c2d/_IBC/juno-osmosis.json#L34


Not able to IBC CW20 token (GLTO) using our relayer. is it because we need something like the above?

Rizbe:
> > @JeremyParish69 Noticed here looking at JUNO<>OSMO:
> > https://github.com/cosmos/chain-registry/blob/614bb87ba1daff10b6e0e6de1fcb56d8b57b9c2d/_IBC/juno-osmosis.json#L34
> > 
> > Not able to IBC CW20 token (GLTO) using our relayer. is it because we need something like the above?
> 
> CW20 tokens need a special CW20-to-ICS20 contract on the originating chain and another channel to be able to IBC transfer them; that's what that port is--the contract

Any doc around this I can follow to get this going <3

JeremyParish69:
Might start here, @Rizbe ?
https://github.com/CosmWasm/cw-plus/tree/main/contracts/cw20-ics20

PR #2630, opened 2023-08-07
Update KYVE info and logo

Galadrin:


JeremyParish69:
Thanks. Replaced with this (which retains a 'chain' logo)
https://github.com/cosmos/chain-registry/pull/2631

PR #2633, opened 2023-08-08
Move passage to archive dir

dylanschultzie:
@JeremyParish69 this is what I was trying to communicate in the chat, I figured creating a PR to discuss might make it more clear.

Under `_archive` would be the original chain name, `passage`, then a step down would be the chain-id, `passage-1`. If they continued to use `passage-1`, the chain-id could be incremented: `passage-1_1`, etc... however, that wouldn't be the default behavior

Here's the start of the discussion: https://t.me/c/1888994636/235

### DIFF
--- a/_archive/passage/passage-1/assetlist.json
+++ b/_archive/passage/passage-1/assetlist.json
@@ -1,6 +1,6 @@
 {
   "$schema": "../assetlist.schema.json",
-  "chain_name": "passage1",
+  "chain_name": "passage",
   "assets": [
     {
       "description": "The native staking and governance token of the Passage chain.",
--- a/_archive/passage/passage-1/chain.json
+++ b/_archive/passage/passage-1/chain.json
@@ -1,6 +1,6 @@
 {
   "$schema": "../chain.schema.json",
-  "chain_name": "passage1",
+  "chain_name": "passage",
   "chain_id": "passage-1",
   "pretty_name": "Passage",
   "status": "killed",


JeremyParish69:
-There are checks that make sure the chain_name matches the directory it's in; that would have to be modified.
-Any programmed assumptions that chain_name always == to the directory name could lead to failures.
-_archive could also go in the chain's folder? like ~/passage/_archive/passage-1/
-Could do /passage_passage-1/, /passage_passage-1_1/, /passage_passage-2/, etc. to ensure that all chain_name remain unique.
(e.g., If bitcoin forks in Bitcoin and Bitcoin SV, then Bitcoin SV would need some way to say 'pre_fork_chain_name': bitcoin_btc-1 (because Bitcoin SV's pre fork wouldn't be a Bitcoin SV chain, but a Bitcoin chain.)) 

A counter:
for current passage: /passage/chain.json::pre_fork_chain_name: passage_passage-1
pre-fork passage: /passage/_archive/passage_passage-1/chain.json, assetlist.json

benefits:
-checks don't need to be modified,
-assumptions of the chain_name's matching the directory name remains true,
-assumptions of all chain_name's being unique remains remains true,
-navigation to a chain's pre-fork history should be quite sensible this way too

dylanschultzie:
@JeremyParish69 I quite like the idea of doing `/{network_name}/_archive/`, but I'm not quite following the `passage_passage-1` bit - why not just maintain the `chain-id` as the folder name?

JeremyParish69:
> @JeremyParish69 I quite like the idea of doing `/{network_name}/_archive/`, but I'm not quite following the `passage_passage-1` bit - why not just maintain the `chain-id` as the folder name?
> 
> so it'd be `/passage/_archive/passage-1`

The reason is propose naming the folder 'passage_passage-1' instead is so that the folder name would still match the chain_name.

Now this got me thinking:
Imagine a script being given the chain_name: 'passage_passage-1_1'. The script would have to split the string at the first underscore '_' to get 'passage' and 'passage-1_1'. And then construct the path: 'passage/_archive/passage-1_1/' to find the appropriate folder.

It's easier for scripts if it doesn't need to parse strings, and much easier to locate files by chain_name if all the chains existed in a high level folder like right under ~/_archive/ with no additional depth. So maybe ~/_archive/passage_passage-1 really is the easiest way to structure it.

Something as regular as validating IBC connection files would be more complicated (to write and to compute) if it needed to worry about digging deep, or looking for a folder name that doesn't match the chain_name.

JeremyParish69:
Last conversation was over a year ago.

PR #2646, opened 2023-08-10
Modify coreum-osmosis IBC information

keyleu:
After discussion with Osmosis team, a new IBC client was created raising the trust treshhold to 2/3 for better security.
We created a new client and new connection so we need to modify this information.


JeremyParish69:
I think this is possible because it was only just created yesterday

PR #2663, opened 2023-08-12
update version { realio , planq }

Genz22:
- add latest binary version : realio , planq
- add endpoint for planq
- fix endpoint bluzelle


planq/chain.json
```diff
@@ -33,19 +33,23 @@
   },
   "codebase": {
     "git_repo": "https://github.com/planq-network/planq",
-    "recommended_version": "v1.0.3",
+    "recommended_version": "v1.0.7",
```
JeremyParish69, line 36: For planq, if v1.0.7 is incompatible with v1.0.0-v1.0.3 (consensus-breaking), then can we please add v1.0.7 as a new object under `versions` as well?

realio/chain.json
```diff
@@ -97,12 +97,26 @@
           "version": "0.34"
             },
           "ibc_go_version": "6.1.1",
+          "next_version_name": "v0.8.3"
+        },
+        {
+          "name": "v0.8.3",
+          "recommended_version": "v0.8.3",
+          "compatible_versions": [
+            "v0.8.2","v0.8.3"
```
JeremyParish69, line 106: if v0.8.2 is compatible with v0.8.3, then we simply update the existing v0.8.2 version object (but it looks like you've added a new object, which could be confusing)

Genz22: yeah v0.8.2 is compatible with v0.8.3

planq/chain.json
Genz22: yes please

Genz22:
check pls

realio/chain.json
JeremyParish69: Sorry, this was never resolved. There is still a separate version object for v0.8.2 that should be removed
image

Genz22: https://github.com/cosmos/chain-registry/pull/2663/commits/eadf6aeb331e028e299dd749c39da976ba4204c7

planq/chain.json
```diff
@@ -65,6 +70,20 @@
           "darwin/arm64": "https://github.com/planq-network/planq/releases/download/v1.0.3/planq_1.0.3_Darwin_arm64.tar.gz",
           "windows/amd64": "https://github.com/planq-network/planq/releases/download/v1.0.3/planq_1.0.3_Windows_x86_64.zip"
         }
+      },
+      {
+        "name": "v1.0.7",
+        "recommended_version": "v1.0.7",
+        "compatible_versions": [
+          "v1.0.7"
+        ],
```
JeremyParish69, line 79: can we add the cosmos_sdk_version, consensus, and ibc_go_version into the v1.0.7 version object please?

Genz22: added https://github.com/cosmos/chain-registry/pull/2663/commits/2acdf9fa1aacc451e8d541b4ca1f40f1ce8a7f13

PR #2664, opened 2023-08-12
Returning the image and adding coingecko

Kamikaza731:
Image was probably removed by accident, added coingecko api id

### DIFF
--- a/composable/assetlist.json
+++ b/composable/assetlist.json
@@ -27,7 +27,11 @@
           },
           "provider": "Composable Finance"
         }
-      ]
+      ],
+      "logo_URIs": {
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/composable/images/pica.svg"
+      },
+      "coingecko_id": "picasso"
     },
     {
       "description": "The native staking and governance token of Kusama Relay Chain.",


JeremyParish69:
Hi @Kamikaza731 ,
The image reference was actually removed intentionally.
Tracing this asset to its origin, picasso::ppica, the reference to the image does indeed exist. By opting to not reference the image again in this composable::ppica asset object, we are showing that the image (and any potential changes to it) should remain synchronized across both assets.
This makes it easier to update images, because if the name or location of the file happens to change, we longer need to track down and update multiple references to the same image (or risk breaking some forgotten references). Thus, excluding the image reference in this file more appropriately indicates the relationship between this asset and its origin, and is arguably safer and cleaner.
However, not all software reading the chain registry are necessarily sophisticated enough yet to be able to pull properties form the traced origin assets, in which case adding an image reference here could be useful for some cases. 

The coingecko id, however, was missing from picasso::ppica, and I think it'd be good to add there (and only there).

Kamikaza731:
Hello @JeremyParish69 

I only did this because I noticed some services are not working as intended or are just missing images that should have been there. For example, I noticed in https://restake.app since the last chain reg update to the asset list on the composable directory, for some reason it no longer detects it (for some reason I think because the logo is missing it won't load it which is bizarre if true). Cosmos directory https://cosmos.directory/composable also misses the images. I do not know if some other app/service was affected by this. 

I did notice you have all of the necessary data needed in _non-cosmos/picasso, but what about the data needed for Centauri-1? In this case, maybe having a symlink for assetlist.json could work. But I did notice there is a "traces" additional mintage in which case it might not work so I think every chain should have its asset list.

Anyway, I won't interfere any more since you seem to maintain these directories I only thought the logo was removed by accident.

JeremyParish69:
Hi @Kamikaza731 Thanks for pointing out the issues with these missing.

I'm starting to think you're probably right about these needing to be there. It's probably time to separate the data needs from the interface needs so that any API looking for an image doesn't need to be smart enough to crawl up the origin trace.

If interested in joining the discussion and following the development, be sure to join the Telegram working group (linked on the README)

tombeynon:
Hey @Kamikaza731 @JeremyParish69 - I'm the REStake/Cosmos.directory dev. Ultimately I think the apps should use the logo URIs defined in the main chain.json file for the _chain_ imagery, rather than the 'first' asset as it was before - I've made this update to cosmos.directory and restake.app so composable is now displaying as expected. 

I actually agree with Jeremy's initial suggestion to avoid duplicating the assets and that apps should really look them from the source chain. The Chain Registry is already pretty good from an API point of view, and I think should primarily be a sensible data source rather than jumping through hoops to make it interface suitable. It's not a big lift for apps to lookup this data, so long as all the necessary attributes are available to make the connections easily. Having said that I think both options have pros and cons and I ultimately don't mind which way the decision goes.

PR #2676, opened 2023-08-16
update oraichain version v0.41.3

hexskrt:


oraichain/chain.json
```diff
@@ -33,25 +33,25 @@
...
+      "linux/amd64": "https://orai.s3.us-east-2.amazonaws.com/v0.41.3/oraid"
     },
     "genesis": {
       "genesis_url": "https://raw.githubusercontent.com/oraichain/oraichain-static-files/master/mainnet-static-files/genesis.json"
     },
     "versions": [
       {
-        "name": "v0.40.3",
-        "recommended_version": "v0.40.3",
+        "name": "v0.41.3",
```
JeremyParish69, line 48: Can we please add this as an additional object within the 'versions' array (after v0.40.3) so we can start building up a history?

hexskrt: Apologize, i forget. Yes please

PR #2688, opened 2023-08-17
Update assetlist.json - coingeckoid = sei-network, not sei

PFC-developer:
coingecko id is sei-network.
see - https://www.coingecko.com/en/coins/sei


dylanschultzie:
@PFC-developer where are you seeing it as `sei-network` rather than `sei`? It was previously `sei-network`, but the team had changed it.

PFC-developer:
I just went to the website and it shows that

JeremyParish69:
It is indeed sei-network

PR #2690, opened 2023-08-18
add doravota testnet

tomas-dora:


testnets/doravotatestnet/chain.json
```diff
@@ -0,0 +1,99 @@
+{
+  "$schema": "../../chain.schema.json",
+  "chain_name": "doravotatestnet",
```
JeremyParish69, line 3: Will the Mainnet also be called Dora Vota? I thought I've seen somewhere a chain being called 'Dora', and assuming that is the mainnet counterpart to this chain, then I'd suggest the chain_name here be 'doratestnet' so that the mainnet can be called 'dora'. Otherwise, with this name, the mainnet's chain_name must be registered as 'doravota'

testnets/doravotatestnet/assetlist.json
```diff
@@ -0,0 +1,27 @@
...
+            "exponent": 6
+          }
+        ],
+        "base": "uDORA",
+        "name": "Dora Vota",
+        "display": "DORA",
+        "symbol": "DORA",
+        "logo_URIs": {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/cudostestnet/images/doravota.png",
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/cudostestnet/images/doravota.svg"
```
JeremyParish69, line 23: since these two images have a different appearance, I'd ask we only reference one of them to be the token logo under logo_URI's.
Although, the `images` property should allow for alternative logos.

tomas-dora: Yes, the name of our chain is `Dora Vota` (the testnet is `Dora Vota Testnet`), so that's what I'm setting up here is `doravotatestnet`.

testnets/doravotatestnet/chain.json
tomas-dora: Yes, the name of our chain is `Dora Vota` (the testnet is `Dora Vota Testnet`), so that's what I'm setting up here is `doravotatestnet`.

testnets/doravotatestnet/assetlist.json
tomas-dora: already update

PR #2699, opened 2023-08-19
Update the lava testnet2 rpc

andy-infstones:


JeremyParish69:
I think we'll deal with that in this PR: https://github.com/cosmos/chain-registry/pull/2700

andy-infstones:
> I think we'll deal with that in this PR: #2700

Thank you

PR #2702, opened 2023-08-19
Update chain.json of Sentinel

bsrinivas8687:


sentinel/chain.json
```diff
@@ -53,7 +53,14 @@
         "recommended_version": "v0.10.1",
         "compatible_versions": [
           "v0.10.1",
-          "v0.10.0"
+        ]
```
JeremyParish69, line 56: the comma here is making the check fail. Can we add back in v.0.10.0 as a compatible version here?

bsrinivas8687: v0.10.0 isn't compatible version.

PR #1851, opened 2023-04-12
Add `xion-testnet-1` network info

froch:
- the Xion testnet is the first testnet from Burnt Labs
- https://xion.burnt.com

PR #2708, opened 2023-08-21
Add `xion-testnet-1` network info

froch:
- Reviving [a PR i let go stale](https://github.com/cosmos/chain-registry/pull/1851), sorry about that.
- I've also [re-opened a PR to the satoshilabs repo](https://github.com/satoshilabs/slips/pull/1588) for the bech32 prefix.


testnets/_IBC/xiontestnet-axelartestnet.json
```diff
@@ -0,0 +1,30 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "xiontestnet",
```
JeremyParish69, line 4: do avoid duplicate files for the same connection, the chain_names are sorted alphabetically in both the name of the file and in the usage of chain_1 and chain_2, so xiontestnet must be chain_2 in all these cases

testnets/xiontestnet/assetlist.json
```diff
@@ -0,0 +1,138 @@
+{
+  "$schema": "../../assetlist.schema.json",
+  "chain_name": "xion-testnet-1",
```
JeremyParish69, line 3: please change chain_name to 'xiontestnet' in this file and the chain.json file. (xion-test-1 will still be recorded as chain_id)

testnets/xiontestnet/chain.json
```diff
@@ -0,0 +1,135 @@
...
+      {
+        "denom": "uxion"
+      }
+    ]
+  },
+  "codebase": {
+    "git_repo": "https://github.com/burnt-labs/xion",
+    "recommended_version": "v0.3.2",
+    "compatible_versions": [
+      "v0.1.0",
```
JeremyParish69, line 37: is v0.1.0 really compatible (non-consensus-breaking) with v0.3.2? I'm guessing it's not. But if they are, they should be recorded as a single version object in the versions array.

testnets/_IBC/xiontestnet-axelartestnet.json
froch: Fixed the alpha ordering of the files and the chain entries. Thanks!

testnets/xiontestnet/assetlist.json
froch: Updated `chain_name` to `xiontestnet`. Also renamed in chain cfg. Thanks!

testnets/xiontestnet/chain.json
froch: Nice catch. Removed. Thanks!

testnets/xiontestnet/assetlist.json
```diff
@@ -0,0 +1,138 @@
...
+      "type_asset": "ics20",
+      "base": "ibc/D934516FBE457F3A98AFABD87E0EFF7F95A15325C191EA8CDD7763C702FDDEC2",
+      "name": "Axelar AXL Token",
+      "display": "axl",
+      "symbol": "AXL",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "axelar",
```
JeremyParish69, line 58: sorry, I missed this before. These should be the testnet chains, e.g., 'axelartestnet' not 'axelar'.

froch: Nice catch. I fixed all three; `axelartestnet`, `nobletestnet` and `osmosistestnet`

PR #2716, opened 2023-08-22
Add gas prices to passage chain

aleem1314:
To integrate Passage network on Osmosis UI at least one of the fee token must be defined. This PR adds low, average and high gas prices.


anilCSE:
@JeremyParish69 please review and merge this

PR #2726, opened 2023-08-24
Add Wormhole Gateway ibc connection to Osmosis

nik-suri:


JeremyParish69:
Pending creation of gateway chain directory

PR #2728, opened 2023-08-25
feat (teritori): Add png logo, update chain logo

omniwired:

PR #2730, opened 2023-08-25
Fix typo on Teritori SVG logo

kleomedes:


kleomedes:
let's fix it in [#2728](https://github.com/cosmos/chain-registry/pull/2728)

PR #2731, opened 2023-08-25
Add NEOK to asset list

sirnicolaz:
This PR adds the NEOK token, a native EVMOS ERC20 token, to the asset information in the EVMOS chain `assetlist.json`.


bitcoinfinancier (approved):

evmos/assetlist.json
```diff
@@ -55,6 +55,23 @@
           }
         }
       ]
+    },
+    {
+      "description": "The token of Neokingdom DAO.",
+      "denom_units": [
```
JeremyParish69, line 61: please define the base denom (erc20/0x6...) as a denom unit with 0 exponent, and 'neok' as another denom unit with exponent 18

sirnicolaz: Thanks for the suggestion. Done

PR #2756, opened 2023-08-31
Add NEXX to Junotestnet-Osmosistestnet.json

NEXXGENAI:


testnets/_IBC/junotestnet-osmosistestnet.json
```diff
@@ -28,6 +28,38 @@
         "dex": "osmosis"
       }
     }
-  ]
-}
-  
\ No newline at end of file
+  ],
```
JeremyParish69, line 31: this doesn't follow the schema, as it only supports one set of connection ids and client ids. Please use the same client+connection ids, and set up the channel using those.

testnets/_IBC/junotestnet-osmosistestnet.json
```diff
@@ -28,6 +28,38 @@
...
+      "chain_2": {
+        "chain_name": "osmosistestnet",
+        "client_id": "07-tendermint-832",
+        "connection_id": "connection-739"
+      },
+      "channels": [
+        {
+          "chain_1": {
+            "channel_id": "channel-622",
+            "port_id": "transfer"
```
JeremyParish69, line 47: I'm surprised to see the transfer port id, when I'm used to seeing that cw20s need a contract and wasm.contract port for moving cw20s

testnets/_IBC/junotestnet-osmosistestnet.json
piscesyuma: Hi JeremyParish69, 
We've created a new relayer between juno testnet(uni-6) and osmosis testnet(osmo-test-5).
how can we add our own new set of connection ids and client ids?
We can not find and use the connection id and client id that was already registered at junotestnet-osmosistestnet.json file.

testnets/_IBC/junotestnet-osmosistestnet.json
piscesyuma: I think juno factory token is not follow cw20 format.

testnets/_IBC/junotestnet-osmosistestnet.json
piscesyuma: The relayer info(channel id, client id, connection id) already registered does not exists.
Is it possible for us to swap original relayer info to our new relayer info for junotestnet-osmosistestnet.json?

testnets/_IBC/junotestnet-osmosistestnet.json
JeremyParish69: That's correct, my mistake

testnets/_IBC/junotestnet-osmosistestnet.json
JeremyParish69: I'll need to investigate further to prove it, but I believe it does indeed exist...

piscesyuma: I've already investigated about the existence of already registered channel.
If it was exists, we didn't need to make a new relayer between uni6 to osmo-test-5.

image

NEXXGENAI: > I'll need to investigate further to prove it, but I believe it does indeed exist...

It does not appear to exist, which is why we created a new relayer between uni-6 and osmo-test5.  Please provide further direction so we can get this moving.  Thank you

JeremyParish69: The original client maybe be revived via governance

PR #2758, opened 2023-08-31
akash: prep for mainnet6

andy108369:
## Do not merge yet

Can be merged once the network gets [upgraded](https://www.mintscan.io/akash/proposals/224), i.e. when the validators will start indexing new blocks after [12606074](https://www.mintscan.io/akash/blocks/12606074) height

### DIFF
--- a/akash/chain.json
+++ b/akash/chain.json
@@ -27,27 +27,27 @@
   },
   "codebase": {
     "git_repo": "https://github.com/akash-network/node/",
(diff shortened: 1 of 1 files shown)


andy108369:
# OK TO MERGE

JeremyParish69:
Please add a new version object in the `versions` array rather than just writing over the previous version

JeremyParish69:
Change already applied, but more correctly. Thanks

PR #2760, opened 2023-08-31
Update Description for Crypto.org Chain

aw126:


cryptoorgchain/assetlist.json
```diff
@@ -3,7 +3,7 @@
     "chain_name": "cryptoorgchain",
     "assets": [
       {
-        "description": "CRO coin is the token for the Crypto.com platform.",
+        "description": "Crypto.org Chain - driving mass adoption of blockchain technology through use cases like Payments, DeFi and NFTs, prioritizing speed and seamless transactions.",
```
JeremyParish69, line 6: This description should be reserved for the CRO asset. This kind of description can go under chain.json

aw126: Noted, we will revise it. Thank you.

PR #2793, opened 2023-09-07
Update Chronic Network Status To 'Killed'

discoverdefiteam:
update status to 'killed' & remove public endpoints

### DIFF
--- a/chronicnetwork/chain.json
+++ b/chronicnetwork/chain.json
@@ -26,14 +26,22 @@
     "compatible_versions": [
       "v1.1.0"
     ],
-    "binaries": {},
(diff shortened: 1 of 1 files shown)


chronicnetwork/chain.json
```diff
@@ -26,22 +26,14 @@
     "compatible_versions": [
       "v1.1.0"
     ],
-    "binaries": {
-      "linux/amd64": "https://github.com/ChronicNetwork/cht/releases/download/v.1.1.0/cht"
```
JeremyParish69, line 30: Can we leave all this stuff in here, but only change the status?

PR #2794, opened 2023-09-07
corrected gas for fxcore

LanceLan96:


fxcore/chain.json
```diff
@@ -20,7 +20,7 @@
     "fee_tokens": [
       {
         "denom": "FX",
-        "fixed_min_gas_price": 4200000000000,
+        "fixed_min_gas_price": 4000000000000,
```
JeremyParish69, line 23: This minimum is still greater than the low amount

JeremyParish69: 4000000000000
400000000000
4200000000000
500000000000

LanceLan96: corrected

PR #2802, opened 2023-09-08
correct names and network names for terra chains

gregnuj:


reversesigh:
The network type should probably remain "mainnet". The other option would be "testnet"

gregnuj:
> The network type should probably remain "mainnet". The other option would be "testnet" or "devnet". Classic is more colloquial than an accurate network type.

Okay, I will change it

gregnuj:
> The network type should probably remain "mainnet". The other option would be "testnet" or "devnet". Classic is more colloquial than an accurate network type.

Resolved

PR #2805, opened 2023-09-08
Update comercionetwork software version

marcotradenet:


commercionetwork/chain.json
```diff
@@ -21,19 +21,19 @@
...
+      "v5.0.0"
     ],
     "genesis": {
       "genesis_url": "https://raw.githubusercontent.com/commercionetwork/chains/master/commercio-3/genesis.json"
     },
     "versions": [
       {
-        "name": "v4.0.0",
-        "recommended_version": "v4.0.0",
+        "name": "v5.0.0",
```
JeremyParish69, line 33: instead of writing over v4, could you please keep v4 in this array and add a new object below for v5?

marcotradenet: Updated versions array. Thank you

PR #2810, opened 2023-09-09
added OIN token asset (sei)

Dizz02:


sei/assetlist.json
```diff
@@ -23,6 +23,25 @@
...
             },
             "coingecko_id": "sei-network"
+        },
+
+        {
+            "description": "OIN Token ($OIN) is a pioneering digital asset developed on the $SEI Blockchain. Representing more than just a cryptocurrency, $OIN is a symbol of the future of decentralized finance, designed with the vision of reshaping the crypto landscape.",
+            "denom_units": [
+                {
+                    "denom": "factory/sei12q0zv3c4cd9jkupn0krazdycc5ftw9wzt9vmhu/OIN",
+                    "exponent": 6
```
JeremyParish69, line 33: exponent for base denom must be 0.

sei/assetlist.json
```diff
@@ -23,6 +23,25 @@
...
+                }
+            ],
+            "base": "OIN",
+            "name": "OIN",
+            "display": "oin",
```
JeremyParish69, line 38: this demon isn't defined. You can add another denom_unit above called 'oin' with exponent 6.

sei/assetlist.json
Dizz02: Done thanks

sei/assetlist.json
Dizz02: Done too

PR #2811, opened 2023-09-10
Update TerpNetTestNet Fee

Xiti922:
- Added THIOLX like Fee tokens.
- Added TERPX Like Stake tokens
- Added Key_algos
- Added slip44


Xiti922:
@chain-registry modules is fetching info from 24 days ago.

PR #2819, opened 2023-09-13
Update version

activenodes:
- https://github.com/bandprotocol/chain/releases/tag/v2.5.3


bandchain/chain.json
```diff
@@ -42,10 +42,10 @@
         ]
       },
       {
-        "name": "v2.5.2",
-        "recommended_version": "v2.5.2",
+        "name": "v2.5.3",
```
JeremyParish69, line 45: Please add a new version object instead of writing over the previous version

PR #2833, opened 2023-09-14
rename terra to terraclassic

gregnuj:


JeremyParish69:
I'm not a fan of this change as many APIs and service providers are already using chain_name 'terra', which is supposed to be immutable.

PR #2836, opened 2023-09-15
Updated Unification

jo-chainmasters:


unification/chain.json
```diff
@@ -29,49 +29,68 @@
...
+      "version": "0.34"
     },
+    "cosmwasm_enabled": false,
+    "ibc_go_version": "5.3.1",
     "genesis": {
       "genesis_url": "https://raw.githubusercontent.com/unification-com/mainnet/master/latest/genesis.json"
     },
     "versions": [
       {
-        "name": "v1.6.3",
```
JeremyParish69, line 48: Instead of writing of v1.6.3 here, can we please just add v1.8.2 as a new version object below?

jo-chainmasters: Will do

jo-chainmasters: Done. Hopefully nothing is missing

unification/chain.json
```diff
@@ -29,49 +29,146 @@
...
+        },
+        "cosmos_sdk_version": "0.46.14",
+        "ibc_go_version": "5.3.1",
+        "ics_enabled": [
+          "ics20-1"
+        ],
+        "cosmwasm_enabled": false,
+        "next_version_name": "v1.8.2"
+      },
+      {
```
JeremyParish69, line 118: Thanks for adding all this. More info is better. But can we roll v1.8.0, v1.8.1, and v1.8.2 into a single version object please? (since they're compatible/non-conensus-breaking).
Really, just keep this v.1.8.2 object, (maybe choose a generic name like 'v1.8' or better yet use the name in the proposal), and just remove the separate v1.8.1 and v1.8.0 objects above

jo-chainmasters: Next try :)

PR #2796, opened 2023-09-07
Add EntryPoint public testnet

migueldingli1997:

PR #2838, opened 2023-09-15
Update EntryPoint fee token to ATOM

migueldingli1997:
As of #2796 the fee token on the EntryPoint public testnet has been changed to ATOM.

For the record, here's an example of a transaction paid for using ATOM: https://explorer.entrypoint.zone/entrypoint/tx/56D647EE82F479674A8817AC182436C66483CC5F4B5F2D14296130750CEEA332

### DIFF
--- a/testnets/entrypointtestnet/assetlist.json
+++ b/testnets/entrypointtestnet/assetlist.json
@@ -29,6 +29,27 @@
         "governance",
         "vaults"
       ]
+    },
+    {
+      "description": "The native staking and governance token of the Theta testnet version of the Cosmos Hub, via Osmosis.",
+      "denom_units": [
+        {
+          "denom": "ibc/E774302AE43D5FA03522C42B14823288E7DD1B2F54F85DFD3D6FC3E5FCF54645",
+          "exponent": 0
+        },
+        {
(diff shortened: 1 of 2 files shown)


JeremyParish69:
Can we please add this denom in assetlist.json so it appears as a valid asset, and also register the cosmoshubtestnet-entrypointtestnet ibc connection that would result in that ibc hash?

migueldingli1997:
> Can we please add this denom in assetlist.json so it appears as a valid asset, and also register the cosmoshubtestnet-entrypointtestnet ibc connection that would result in that ibc hash?

I've added the ATOM into assetlist.json. However I did not register a new IBC connection because this ATOM is obtained over the connection with Osmosis. The denom trace of the ATOM is: `transfer/channel-0/transfer/channel-1497` where `channel-1497` is `Osmosis <> Cosmos Hub` and `channel-0` is `EntryPoint <> Osmosis`.

Source: https://testnet-rest.entrypoint.zone/ibc/apps/transfer/v1/denom_traces/E774302AE43D5FA03522C42B14823288E7DD1B2F54F85DFD3D6FC3E5FCF54645

PR #2842, opened 2023-09-15
Update oraichain to v0.41.4

hexskrt:
https://github.com/oraichain/orai/releases/tag/v0.41.4

### DIFF
--- a/oraichain/chain.json
+++ b/oraichain/chain.json
@@ -56,18 +56,6 @@
         }
       },
       {
-        "name": "v0.41.3",
-        "height": 12353514,
-        "proposal": 185,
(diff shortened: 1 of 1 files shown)


oraichain/chain.json
```diff
@@ -65,6 +66,19 @@
         "binaries": {
           "linux/amd64": "https://orai.s3.us-east-2.amazonaws.com/v0.41.3/oraid"
         }
+      },
+      {
+        "name": "v0.41.4",
```
JeremyParish69, line 71: actually, since v0.41.4 is compatible with v0.41.3, we don't create a new version object in this case--we simply update the version object above

PR #2854, opened 2023-09-18
Update Impacthub Testnet to v2

Michael-Ixo:
Update Impacthub Testnet to v2, already happened.
Also adding testnet ibc impacthubtestnet-osmosistestnet


Michael-Ixo:
Hi @JeremyParish69 sorry I didnt notice that locally, fixed and pushed

PR #2873, opened 2023-09-20
Update composable-injective.json

lichdu29:


JeremyParish69:
Changing IBC connections can be quite problematic. Do we have any proof the the originally used connection has little to no funds transferred (and to be abandoned)?

lichdu29:
Well the composable team ask us to do that, I am also the one wgo create the previous channel and it is expried

lichdu29:
and yes there are no funds on that channel @JeremyParish69 so we decide to create a new channel instead of reviving it by gov

JeremyParish69:
OK I found this: https://www.mintscan.io/injective/relayers/channel-141/not-established/channel-10

PR #2903, opened 2023-09-27
add Secretnetwork SNIP channels

ertemann:
supported widely by Crosnest, Lavender.Five, cryptocrew and others.


ertemann:
@JeremyParish69 lemme know how you want me to format these names. but for example the Axelar channel has more value passed over it than any other standard Secret transfer channel. Them being in the registry would be great.

JeremyParish69:
What we've been doing so far @ertemann is adding new channels under the same connection so it all goes in a single file. E.g.,
https://github.com/cosmos/chain-registry/blob/master/_IBC/osmosis-secretnetwork.json

ertemann:
That works for some but the Axelar channel uses a different client and connection. So that wont work then right?

JeremyParish69:
> That works for some but the Axelar channel uses a different client and connection. So that wont work then right?

Correct. A requirement for different client/connection was never planned for these files. Is it possible to create a new channel using the same clients?

ertemann:
no it is not possible, over 10m value is locked on that Axelar channel. 

there have been more cases where Wasm/ICA channels dont use the same clients/connections. we should be able to accomodate for that.

In the RLY software we just an extension on the name, there it doesnt break anything.

maybe @dylanschultzie has some tips on how to handle this exception as well.

tac0turtle (approved):

JeremyParish69:
@tac0turtle the proposed changes in the PR go against design and makes future PRs fails due to the ibc CI check failing

JeremyParish69:
I'd like to revert this until we come up with a way to design it in and adjust the verification checks where necessary

PR #2914, opened 2023-09-28
feat(chain): Add polkachu rpc and rest to chain and reorder its grpc …

0xFable:
…entry. Add one explorer too

### DIFF
--- a/migaloo/chain.json
+++ b/migaloo/chain.json
@@ -179,7 +179,7 @@
     ],
     "rest": [
       {
-        "address": "https://migaloo-rest.polkachu.com:443",
+        "address": "https://migaloo-api.polkachu.com:443",
         "provider": "Polkachu"
       },
       {
--- a/migaloo/chain.json
+++ b/migaloo/chain.json
@@ -137,6 +137,10 @@
   "apis": {
     "rpc": [
       {
+        "address": "https://migaloo-rpc.polkachu.com:443",
+        "provider": "Polkachu"
+      },
+      {
         "address": "https://migaloo-rpc.lavenderfive.com",
         "provider": "Lavender.Five Nodes 🐝"
       },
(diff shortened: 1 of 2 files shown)


0xFable:
Checked this PR again this morning and noticed changes in asset file, these are unrelated. Removed them.

migaloo/chain.json
```diff
@@ -212,6 +220,10 @@
       }
     ],
     "grpc": [
+      {
+        "address": "migaloo-grpc.polkachu.com:20790",
+        "provider": "Polkachu"
```
dylanschultzie, line 225: Why did you reorder this?

migaloo/chain.json
```diff
@@ -174,6 +178,10 @@
       }
     ],
     "rest": [
+      {
+        "address": "https://migaloo-rest.polkachu.com:443",
```
dylanschultzie, line 182: This doesn't work

dylanschultzie (changes requested):

migaloo/chain.json
0xFable: Your right for some reason its API and not rest. Not sure why I was thinking of rest. Fixed now

migaloo/chain.json
0xFable: Only because I added Polkachu as the top entry in the other two areas and wanted to be consistent. 
Tools which use this registry to get these URLS tend to only consider the first one in the list.

0xFable:
@dylanschultzie it is ready for review again. Any other changes needed?

0xFable:
@JeremyParish69 Thank you for merging this !! 
It ended up being needed for a very specific purpose. The top node in each section seems to be the default most tools use and the lavender five node was giving quite a few 429s for multisig tools.

PR #2919, opened 2023-09-29
Update: composable v5.2.0

hexskrt:
https://github.com/notional-labs/composable-centauri/releases/tag/v5.2.0

### DIFF
--- a/composable/chain.json
+++ b/composable/chain.json
@@ -33,9 +33,10 @@
   },
   "codebase": {
     "git_repo": "https://github.com/notional-labs/composable-centauri",
-    "recommended_version": "v5.1.0",
+    "recommended_version": "v5.2.0",
     "compatible_versions": [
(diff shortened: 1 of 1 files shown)


composable/chain.json
```diff
@@ -118,19 +119,19 @@
...
         ],
         "cosmos_sdk_version": "v0.47.3",
         "ibc_go_version": "v7.0.1",
         "consensus": {
           "type": "cometbft",
           "version": "0.37.1"
         },
-        "height": 1515288,
-        "proposal": 10,
+        "height": 1769900,
```
JeremyParish69, line 134: The height property should reflect when v5 started, and not when v5.2.0 started.

reversesigh: Closing as this has been resolved with https://github.com/cosmos/chain-registry/pull/2919

PR #2929, opened 2023-10-03
Add Observatory to explorers list

livthomas:
[Observatory](https://observatory.zone/) is not a typical chain explorer where you would be able to browse accounts or transactions. But it provides some insightful data especially about the state of decentralization of individual chains that cannot be found anywhere else.

Many foundations and validators in the Cosmos ecosystem already use Observatory to make informed decisions on how to improve their chains. We would like to add Observatory to `chain-registry` so even more users will become aware of it and will make use of this data.

### DIFF
--- a/agoric/chain.json
+++ b/agoric/chain.json
@@ -253,6 +253,10 @@
       "kind": "Stakeflow",
       "url": "https://stakeflow.io/agoric",
       "account_page": "https://stakeflow.io/agoric/accounts/${accountAddress}"
+    },
+    {
+      "kind": "observatory",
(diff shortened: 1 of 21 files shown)


JeremyParish69:
Apologies for my silence. I did see it but don't have a solution ready at this time. I do want to include this into a growing idea I'm working on called the interface registry.

JeremyParish69:
I don't think the Chain Registry is ready to include interfaces like these at this time. Discussion can continue--perhaps in the TG chat? For now I'd like to cleanup stale PRs.

PR #2961, opened 2023-10-12
update exponent of FX

LanceLan96:


fxcore/assetlist.json
```diff
@@ -6,8 +6,12 @@
       "description": "The native staking token of the Function X",
       "denom_units": [
         {
-          "denom": "FX",
+          "denom": "fx",
```
JeremyParish69, line 9: if the exp:0 denom is 'fx', then the 'base' must match and must also be changed to 'fx' (and same with staking token and gas token on the chain.json file)

devon-chain: @JeremyParish69  @LanceLan96 base denom must be FX, because the staking token and gas token on f(x)Core are FX, not fx, so the new denom will be a different name

JeremyParish69: Thanks @devon-chain , in this case 'WFX' should be specified as the display unit here: https://github.com/FunctionX/chain-registry/blob/6666eee85dcb3f397de5a5653e8be56bed518d4e/fxcore/assetlist.json#L19

devon-chain: @JeremyParish69 OK, have been modified, thanks

PR #2966, opened 2023-10-13
Add Source Mainnet

Source-Protocol-Cosmos:


source/assetlist.json
```diff
@@ -0,0 +1,34 @@
...
+      ],
+      "base": "usource",
+      "name": "Source",
+      "display": "source",
+      "symbol": "SOURCE",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/source/images/source.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/source/images/source.svg"
+      },
+      "coingecko_id": "source-protocol",
```
JeremyParish69, line 25: Is this the same token? Feels like it could be different
https://www.coingecko.com/en/coins/source-protocol

PR #2969, opened 2023-10-14
IBC Path kujira <> terra2

danbryan:
This path has existed for 470 days, but was not included in the chain registry.  I confirmed the clients, connections and channels were open.


dylanschultzie:
Thanks @danbryan !

PR #2973, opened 2023-10-16
Rebranding of `Crypto.org Chain` to `Cronos POS Chain`

lezzokafka:
As of October 16th, 2023, the Crypto.org Chain has been rebranded as the Cronos POS Chain, with 'POS' representing Proof-of-Stak.

This pull request is intended to update the chain's name and its logos.

Announcement can be found on https://cronospos.substack.com/p/cryptoorg-chain-becomes-cronos-pos for reference.


cryptoorgchain/chain.json
```diff
@@ -1,9 +1,9 @@
 {
   "$schema": "../chain.schema.json",
-  "chain_name": "cryptoorgchain",
+  "chain_name": "cronosposchain",
```
JeremyParish69, line 3: Can we please leave the chain_name alone? but everything else is fine to change.

lezzokafka: Thanks @JeremyParish69! Centernally, and keeping it as `cryptoorgchain`

JeremyParish69:
I also separated the logo images into separate image object because they look significantly different.

PR #3033, opened 2023-10-24
Update chain.json

antrixy:
Added Antrix validators end point

### DIFF
--- a/evmos/chain.json
+++ b/evmos/chain.json
@@ -616,7 +616,7 @@
         "provider": "StakeTown"
       },
       {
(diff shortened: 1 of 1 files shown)


evmos/chain.json
```diff
@@ -616,8 +616,12 @@
         "provider": "StakeTown"
       },
       {
-        "address": "https://evmos.jsonrpc.liveraven.net",
+        "address": "https://evmos.jsonrpc.liveraven.,net",
```
JeremyParish69, line 619: why this added comma?

antrixy: Must have been added by mistake

PR #3046, opened 2023-10-25
Add BITS token for Juno

kevinroleke:


juno/assetlist.json
```diff
@@ -1849,6 +1849,33 @@
...
+      "description": "BITS - Native token of Bitswift Cash",
+      "type_asset": "cw20",
+      "address": "juno14fz92ehqt37e096xr95kmy8nc0kz803uezxtg4fwx7agjjma86sqm8mg3h",
+      "denom_units": [
+        {
+          "denom": "cw20:juno14fz92ehqt37e096xr95kmy8nc0kz803uezxtg4fwx7agjjma86sqm8mg3h",
+          "exponent": 0
+        },
+        {
+          "denom": "BITS",
```
JeremyParish69, line 1863: should be lower case 'bits', according to the 'display' property

juno/assetlist.json
```diff
@@ -1849,6 +1849,33 @@
...
+          "exponent": 8
+        }
+      ],
+      "base": "cw20:juno14fz92ehqt37e096xr95kmy8nc0kz803uezxtg4fwx7agjjma86sqm8mg3h",
+      "name": "BITS",
+      "display": "bits",
+      "symbol": "BITS",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/juno/images/bits.png"
```
JeremyParish69, line 1872: I'm a little concerned about the low fidelity of the image here. Do we have a version that looks cleaner? There' compression artifacts, blurry edges, no transparency layer, and some random dot in the top right corner...

kevinroleke:
Updated. Thanks for the feedback @JeremyParish69

JeremyParish69:
> Updated. Thanks for the feedback @JeremyParish69

Thanks. I still see the image now has transparency, but it still has that random marking in the top right...

kevinroleke:
> > Updated. Thanks for the feedback @JeremyParish69
> 
> Thanks. I still see the image now has transparency, but it still has that random marking in the top right...

Sorry about that. Should be all good now :)

PR #3083, opened 2023-10-30
Add ibc usdc and ibc atom as gas prices for neutron-1

NeverHappened:


JeremyParish69:
Can you please add these tokens to Neutron's assetlist? It's so viewers aren't guessing which tokens these are referring to, especially if the original ibc channel changes, it could be hard to find out what it is.

JeremyParish69:
Thanks, there actually a lot that could be fixed here, thought. Are you able to grant modify permissions to repo maintainers so I can fix it up?

NeverHappened:
Oh, I think so, but we can only do it tomorrow. Or maybe you can use review suggestions to show correct code?

JeremyParish69:
This should do it:
https://github.com/neutron-org/chain-registry/pull/4

NeverHappened:
Oh, nice! Thanks a lot for helping!

PR #3102, opened 2023-10-31
add $LOCAL on Kujira and Terra.

desamtralized:


terra2/assetlist.json
```diff
@@ -459,6 +459,34 @@
...
+        },
+        {
+          "denom": "whLOCAL",
+          "exponent": 6
+        }
+      ],
+      "base": "cw20:terra1xzkel96e5e8vfmqw7valzdzzv9hqasfyslclvnmvxdejvpda3xwsskxzus",
+      "name": "whLOCAL",
+      "display": "whLOCAL",
+      "symbol": "whLOCAL",
```
JeremyParish69, line 480: is this being bridged from Kujira via Wormhole?

desamtralized: This is $LOCAL minted on Terra Classic and moved to Terra2 via Wormhole.

terra2/assetlist.json
```diff
@@ -459,6 +459,34 @@
...
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/terra2/images/local.png"
+      },
+      "coingecko_id": "local-money",
```
JeremyParish69, line 484: Please remove this CoinGecko since it's being used on Kujira's LOCAL. The trace to Kujira (once defined herein) will allow interfaces tofind the coingecko id

terra2/assetlist.json
```diff
@@ -459,6 +459,34 @@
...
+          "denom": "whLOCAL",
+          "exponent": 6
+        }
+      ],
+      "base": "cw20:terra1xzkel96e5e8vfmqw7valzdzzv9hqasfyslclvnmvxdejvpda3xwsskxzus",
+      "name": "whLOCAL",
+      "display": "whLOCAL",
+      "symbol": "whLOCAL",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/terra2/images/local.png"
```
JeremyParish69, line 482: please remove this duplicate image, since we only need one copy. The trace will create a pointer to Kujira's $LOCAL's image.

terra2/assetlist.json
desamtralized: Will remove, what will link the whLOCAL on Terra2 to the native LOCAL on Kujira?

A bit of history of the LOCAL token, it was first minted as a cw20 on Terra before the UST collapse and the chain split. Then it was moved to Terra2 via Wormhole, a new version of the token with the same circulating supply was minted on Kujira, all the supply was transferred to a contract that allows the swap of whLOCAL to native LOCAL on Kujira, 1:1. whLOCAL can be transferred to Kujira via IBC/ICS20.

JeremyParish69: Did all the LOCAL supply on Kujira come from the bridge and converted using the contract? Or is Kujira's LOCAL an additional mintage of LOCAL, where the 1:1 converting contract is just extra convenience for users?

JeremyParish69: The linking is handled via the `traces` property, which points each asset's origin. For what it sounds like, LOCAL in Terra Classic was the first, so we'll treat this as the original--I see it is already defined in terra/assetlist.json.
Now Terra2's whLOCAL should have a `traces` property pointing it to Terra Classic (using type 'bridge'). For Kujira's LOCAL, I think it depends where the supply first came from.

desamtralized: Yes, all tokens came from the bridge. We minted a new native token on Kujira with the same supply and sent it all to the contract where users can convert 1:1 from the wormhole version to the native on Kujira.

JeremyParish69: In this case, I would say we should define the Terra2 to Kujira ibc-transferred $LOCAL, as well (with its trace origin pointing to Terra2, using type 'ibc'), and then this Kujira LOCAL that you already drafted would have a trace origin to the ibc Kujira LOCAL, using type 'wrapped'.
I hope that makes sense? If needed, I could try to define some of it for you

desamtralized:
If you can set it up it would be of great help. At this point I'm not sure
what the correct JSON is for each $LOCAL token.

Here's a list of the Token Addresses:

Original LOCAL on Terra
Classic: terra1vchw83qt25j89zqwdpmdzj722sqxthnckqzxxp

whLOCAL on Terra2 - (LOCAL moved from Terra Classic to Terra2 via Wormhole)
terra1xzkel96e5e8vfmqw7valzdzzv9hqasfyslclvnmvxdejvpda3xwsskxzus

whLOCAL on KUJIRA - (LOCAL moved from Terra Classic to Terra2 via Wormhole,
and from Terra2 to KUJIRA via IBC)
ibc/DE92DB84EA7E8CB5EC76BC395CF634D6483B67414AD28BEF3F71EBFD6A780D36
https://finder.kujira.network/kaiyo-1/token/ibc%2FDE92DB84EA7E8CB5EC76BC395CF634D6483B67414AD28BEF3F71EBFD6A780D36

New, native LOCAL on KUJIRA
factory/kujira1swkuyt08z74n5jl7zr6hx0ru5sa2yev5v896p6/local
https://finder.kujira.network/kaiyo-1/token/factory%2Fkujira1swkuyt08z74n5jl7zr6hx0ru5sa2yev5v896p6%2Flocal

whLOCAL:LOCAL Swap contract, (in case you need or curious about it:
https://finder.kujira.network/kaiyo-1/address/kujira1hf84x7k4nkmult2daw65fe8yujr4yetvsv2zs2y33h2lq0smxu4sun5mmh
)

Really appreciate it.

kujira/assetlist.json
```diff
@@ -233,6 +233,65 @@
...
+      "name": "whLOCAL",
+      "display": "whLOCAL",
+      "symbol": "whLOCAL",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "terra2",
+            "base_denom": "cw20:terra1xzkel96e5e8vfmqw7valzdzzv9hqasfyslclvnmvxdejvpda3xwsskxzus"
+            "channel_id": "channel-XX"
```
JeremyParish69, line 260: I wasn't able to find the remote (the terra2 side) channel id, and this IBC channel hasn't been registered--please register it to _IBC/kujira-terra2.json

eRFvyZ: @JeremyParish69 I have made some changes to the code to what I believe is the right channel to use base on my research on https://chainsco.pe/terra2/ibc/connections/connection-13 and https://finder.kujira.network/kaiyo-1/token/ibc%2FDE92DB84EA7E8CB5EC76BC395CF634D6483B67414AD28BEF3F71EBFD6A780D36

Happy to amend anything base on your feedback

kujira/assetlist.json
```diff
@@ -233,6 +233,65 @@
...
+        }
+      ],
+      "type_asset": "ics20",
+      "base": "ibc/DE92DB84EA7E8CB5EC76BC395CF634D6483B67414AD28BEF3F71EBFD6A780D36",
+      "name": "whLOCAL",
+      "display": "whLOCAL",
+      "symbol": "whLOCAL",
+      "traces": [
+        {
+          "type": "ibc",
```
JeremyParish69, line 256: The type here should be 'ibc-cw20', which also requires the port properties

kujira/assetlist.json
JeremyParish69: Sure, we can use that info, but the IBC Connection needs to first be defined in '_IBC/kujira-terra2.json', which it currently is not.

kujira/assetlist.json
eRFvyZ: Just applied those changes hope that's ok

kujira/assetlist.json
eRFvyZ: @JeremyParish69 I'm gonna need some help with this one as I couldn't see anything specific on the schema for that file, I'm sorry very new to IBC, could you expand a bit on what's the connection you are expecting to see in `_IBC/kujira-terra2.json`, thanks

JeremyParish69: Sure, if you look at the Juno-osmosis ibc connection file, for example, we can see that they've defined an ibc connection with a native channel and also a [cw20-specific channel](https://github.com/cosmos/chain-registry/blob/1a9d79ed3affee1bfaf6bad1696dbaae3d36cbbc/_IBC/juno-osmosis.json#L31C2-L46C10). We need a cw20 specific channel to be defined in the kujira-terra2.json file.

kujira/assetlist.json
```diff
@@ -233,6 +233,66 @@
...
+      "name": "whLOCAL",
+      "display": "whLOCAL",
+      "symbol": "whLOCAL",
+      "traces": [
+        {
+          "type": "ibc-cw20",
+          "counterparty": {
+            "chain_name": "terra2",
+            "base_denom": "cw20:terra1xzkel96e5e8vfmqw7valzdzzv9hqasfyslclvnmvxdejvpda3xwsskxzus",
+            "port": "transfer",
```
JeremyParish69, line 260: For ibc-cw20 send, I don't think the port is 'transfer'. I think it would be a wasm. port of some kind.

kujira/assetlist.json
```diff
@@ -233,6 +233,66 @@
...
+            "channel_id": "channel-10"
+          },
+          "chain": {
+            "channel_id": "channel-43",
```
JeremyParish69, line 264: the 'chain' object is missing the 'port' property. The value is likely 'transfer'

kujira/assetlist.json
JeremyParish69: As another example, it looks like someone else also registered a different wasm channel from terra2 to kujira.
But it looks like this token isn't using the same channel. It would be best to use the same channel for all cw20s, but we are able to register mulitple.
https://github.com/cosmos/chain-registry/blob/b539099ada02a25033be7fcdb868f5321296fd26/_IBC/kujira-terra2.json#L37

eRFvyZ: So wait so does that means that they already did it? so we don't need to change the kujira-terra2.json file? because my changes looked exactly the same that those

kujira/assetlist.json
eRFvyZ: I'm guessing the wasm port here match what we have specified on the _IBC/kujira-terra2.json file right?

kujira/assetlist.json
eRFvyZ: I did implemented some changes as you suggested @JeremyParish69 but still a bit confuse as it seems like this changes https://github.com/cosmos/chain-registry/blob/b539099ada02a25033be7fcdb868f5321296fd26/_IBC/kujira-terra2.json#L37 are the changes I was aiming for however can't see how is that block related to our specific token

eRFvyZ: Thinking about it shouldn't this a block like this what we need? for the whLOCAL token? 🤔

```
{
      "chain_1": {
        "channel_id": "channel-10",
        "port_id": "wasm.terra1e0mrzy8077druuu42vs0hu7ugguade0cj65dgtauyaw4gsl4kv0qtdf2au"
      },
      "chain_2": {
        "channel_id": "channel-43",
        "port_id": "transfer"
      },
      "ordering": "unordered",
      "version": "ics20-1",
      "tags": {
        "status": "live",
        "preferred": true
      }
    }
```

JeremyParish69: So, currently I see a standard ibc connection between kujira and terra2 (5/10), and I also see an additional 28/36 wasm ibc connection there as well. But there is still no 10/43 wasm ibc connection registered. You might be able to use the same (28/36) ibc connection that's already registered, but that would result in a different ibc denomination for $LOCAL, and any tokens already transferred via 10/43 would be invalid.

JeremyParish69: Also I'm skeptical that both the 10/43 and the 28/36 would use the same wasm port, but it may indeed be correct...

JeremyParish69: But yes, to have ibc/DE92DB84EA7E8CB5EC76BC395CF634D6483B67414AD28BEF3F71EBFD6A780D36 be considered valid, you'd need a block like the one you posted 3 comments above to register the channel-43 connection.

JeremyParish69:
Will there still be effort to register this?

PR #3110, opened 2023-11-01
Remove Temporal

george-aj:


reversesigh:
It might be better to change the status to killed rather than delete the entire history.

JeremyParish69:
Although a 'killed' status is generally preferred, if the chain hasn't made any significant impact, never had ibc transfers, and never made it past testnet, it might be OK to allow it to be completely removed.

PR #3132, opened 2023-11-06
Add mUSDC to Migaloo

Ange-FE:
Add mUSDC to Migaloo


JeremyParish69:
Should this have a trace to USDC on Migaloo?

Ange-FE:
I don't believe so

PR #3145, opened 2023-11-08
Update Sifchain Version

snobbee:


reversesigh:
If v1.2.0-beta is the upgrade_handler for the new chain version, include that as a new version in "versions" rather than removing the current one. Refer to Cosmos Hub chain history as an example - https://github.com/cosmos/chain-registry/blob/master/cosmoshub/chain.json

snobbee:
@reversesigh I appreciate you sharing your review, I updated the json as you suggested, let me know if there is anything other changes you find that need to be done.

reversesigh:
The order of the version history is reversed - it should be the oldest version first and the most recent version last.

JeremyParish69:
@reversesigh I just went in a reversed the order. Look OK now?

reversesigh:
Just made one change. Looks good to me.

PR #3147, opened 2023-11-08
update humans chain

0x963D:


humans/chain.json
```diff
@@ -1,29 +1,25 @@
 {
   "$schema": "../chain.schema.json",
-  "chain_name": "humans",
+  "chain_name": "Humans.ai",
```
JeremyParish69, line 3: Please don not change the chain_name. It must match the folder name and it may never change. It's purely an internal key, so interfaces should not be displaying this, but instead the pretty_name

0x963D: Hello @JeremyParish69, my bad I overlooked this aspect, added a new commit to correct this, thank you!

PR #3173, opened 2023-11-12
add Sr20 de endpoints for SGE testnet

Sr20dem:


testnets/sgetestnet/chain.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "../chain.schema.json",
   "chain_name": "sgetestnet",
-  "chain_id": "sge-network-3",
+  "chain_id": "sge-network-4",
```
JeremyParish69, line 4: With this chain-id change, do the ibc connections remain functional?
https://github.com/cosmos/chain-registry/blob/master/testnets/_IBC/osmosistestnet-sgetestnet.json

reversesigh: sge-network-4 is the new testnet, so it should probably be created in testnets/sgetestnet2 and this endpoint should be added there after, correct?

PR #3176, opened 2023-11-13
feat(ibc/teritori): Add Teritori IBC values

omniwired:


omniwired:
not sure about these 2 values, I don't know where I can confirm them
https://github.com/cosmos/chain-registry/blob/ee61a5920a6bf9becca984efdd57fd11003a0b85/_IBC/cosmoshub-teritori.json#L5-L6

PR #3188, opened 2023-11-14
Add Indonode Source

elangrr:


reversesigh:
Please remove the p2p port from your RPC url or change it to 443.

JeremyParish69:
> Please remove the p2p port from your RPC url or change it to 443.

Is this port not OK for the seed?
I don't see any port on the designated RPC--just https...

reversesigh:
> > Please remove the p2p port from your RPC url or change it to 443.
> 
> Is this port not OK for the seed? I don't see any port on the designated RPC--just https...

No you are correct. I saw RPC in the URL for the seed and my brain must have switched to thinking I was testing the RPC. Approving and merging.

PR #3202, opened 2023-11-15
kujira assets

codehans:


dynalz (approved):

kujira/assetlist.json
```diff
@@ -361,6 +360,31 @@
...
+          "denom": "factory/kujira1swkuyt08z74n5jl7zr6hx0ru5sa2yev5v896p6/local",
+          "exponent": 0
+        },
+        {
+          "denom": "local",
+          "exponent": 6
+        }
+      ],
+      "base": "factory/kujira1swkuyt08z74n5jl7zr6hx0ru5sa2yev5v896p6/local",
+      "name": "LOCAL",
```
sunnya97, line 377: I think the name should be "Local Money" and that the description should be something longer

JeremyParish69:
I helped draft out what I believe is the situation with LOCAL on Kujura:
https://github.com/cosmos/chain-registry/pull/3102/files#diff-1d9d2c8e24e586a0741ec2191ce697e973a65c99122e648f715c99be34eeca45
Although I don't know if the submitter has abandoned that PR.
Note how there is a trace for the asset and how we don't need to upload a duplicate copy of the logo image. I think we were just blocked on needing the IBC connection to Terra to be registered.

codehans:
Yeah just adding this in so Pulsar can index it. Will leave it up to the Local DAO team to deal with the details. 

Not sure the context for the ibc channel here though. This is a factory token originating on the kujira network

sunnya97:
I think they were also trying to register their legacy token from Terra Classic

sunnya97:
But for this PR, should be fine

PR #3209, opened 2023-11-16
adding sgetestnet4

reeves89:


dylanschultzie (approved):

testnets/sgetestnet4/assetlist.json
```diff
@@ -0,0 +1,33 @@
...
+      "base": "usge",
+      "name": "SGE Testnet4",
+      "display": "sge",
+      "symbol": "SGE",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/sgetestnet4/images/sge.png"
+      },
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/sgetestnet4/images/sge.png"
```
JeremyParish69, line 28: can we point these to the existing sge token logos instead? We have several sets of sge images when we only need one.

We could even use an image_sync and point it to the same image mainnet uses

testnets/sgetestnet4/assetlist.json
```diff
@@ -0,0 +1,33 @@
...
+          "aliases": []
+        },
+        {
+          "denom": "usge",
+          "exponent": 6,
+          "aliases": []
+        }
+      ],
+      "base": "usge",
+      "name": "SGE Testnet4",
```
JeremyParish69, line 20: SGE testnet 4 is fine, although, would if this is not going to be the primary Testnet, then I'd add an 'additional-mintage' trace to the primary SGETestnet.
However is this is going to be primary and the ibc connections retained, I'd say this should renamed to 'sgetestnet', while the current sgetestnet should be renamed to 'sgetestnet3'.
Either is fine, but I would like to add traces here.

testnets/sgetestnet4/chain.json
```diff
@@ -0,0 +1,128 @@
...
+      },
+      {
+        "id": "58556b5fb572e20d41ce686149ab7b1646ad63a9",
+        "address": "65.108.15.170:26656",
+        "provider": "ramuchi.tech"
+      }
+    ]
+  },
+  "logo_URIs": {
+    "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/sgetestnet/images/sge.png",
```
JeremyParish69, line 78: again, we can use image sync so we don't need duplicate images

PR #3213, opened 2023-11-16
add dear doge testnet

tomas-dora:


testnets/deardogetestnet/chain.json
```diff
@@ -0,0 +1,103 @@
+{
+  "$schema": "../../chain.schema.json",
+  "chain_name": "deardogetestnet",
```
JeremyParish69, line 3: please rename chain_name (in both files) and the directory to doravotatestnet.

testnets/deardogetestnet/assetlist.json
```diff
@@ -0,0 +1,31 @@
...
+      "base": "peaka",
+      "name": "Dear Doge Testnet",
+      "display": "DORA",
+      "symbol": "DORA",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/deardogetestnet/images/deardoge.png"
+      },
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/deardogetestnet/images/deardoge.png"
```
JeremyParish69, line 26: A duplicate image is not needed. Please point this to Doravota.  An image_sync would be best.

testnets/deardogetestnet/chain.json
tomas-dora: These are two different testnets, so we need to keep both deardogetestnet and doravotatestnet

testnets/deardogetestnet/assetlist.json
tomas-dora: deardoge's logo may change after that.

testnets/deardogetestnet/chain.json
JeremyParish69: Sorry, I'd forgotten there was already a doravotatestnet when I commented. But I presume it's still a testnet for doravota, so I think in this case it should be doravotatestnet2, a convention we see with many other chains with multiple testnets

tomas-dora:
@JeremyParish69 hi, Our new test network is up and running, I've updated the latest configuration info, please help merge this pr. Thanks!

tomas-dora:
@JeremyParish69  I've updated some more detail data and now it's ready, can you merge this pr?

testnets/deardogetestnet/chain.json
JeremyParish69: Sorry, I don't see how this is resolved. The git_repo and genesis_url still point to dora vota, so it still looks like this should be doravota2...

tomas-dora: I fixed this problem. @JeremyParish69 
https://github.com/cosmos/chain-registry/pull/3213/files#diff-14bbdcbd15770599c05050af74677168f99c7bf946a103deaf11139e23d365e3R29-R34

PR #3221, opened 2023-11-17
add Aura version details

kienvc:


aura/chain.json
```diff
@@ -37,6 +40,13 @@
     "compatible_versions": [
       "v0.7.1"
     ],
+    "cosmos_sdk_version": "0.47.4",
```
JeremyParish69, line 43: can we please add these details to the version obejct?
[Here's ](https://github.com/aura-nw/chain-registry/pull/2)a PR that does that if it helps

kienvc: thanks, merged it.

PR #3230, opened 2023-11-18
Update cosmoshub-realio.json to using new client

hakuno2000:
@MasterPi-2124 done


JeremyParish69:
The registered IBC connection cannot be changed unless we can show that both chains' governance votes to approve so, or if we can show that the previous connection hasn't been used or would leave no abandoned tokens (small amounts from test txs are fine to abandon).

PR #3238, opened 2023-11-20
chore: improve metadata for Noble Argon upgrade

johnletey:


reversesigh (changes requested):
Can you also update ibc_go_version in codebase as well to match argon? https://github.com/cosmos/chain-registry/blob/da5f4ed164fdb7e8a38fef1fe6a394cc0cdd7c89/noble/chain.json#L56C1-L56C32

johnletey:
Just updated @reversesigh!

PR #3240, opened 2023-11-21
Add QSR, DYDX and YETH assets to Osmosis assetlist.json

magiodev:
This PR aims to add QSR, DYDX and YETH ibc assets to the Osmosis assets list


osmosis/assetlist.json
```diff
@@ -433,6 +433,183 @@
...
+          "aliases": []
+        },
+        {
+          "denom": "qsr",
+          "exponent": 6,
+          "aliases": []
+        }
+      ],
+      "type_asset": "sdk.coin",
+      "base": "uqsr",
```
JeremyParish69, line 601: This asset does no exist (in this form) on Osmosis chain

JeremyParish69:
Although fine to add, it's generally redundant to add here for Osmosis' purposes. Can I ask what is the purpose of adding here?

osmosis/assetlist.json
magiodev: You are right, this is wrong but also is redundant as the previous object is also about $QSR. See here: https://github.com/cosmos/chain-registry/pull/3240/files/4bfc09a6600d8d9740b42887159044d3f3c62097#diff-afc1495905a465ffe02b6fa4b659d5aabd602f9eb8c9276b06125b3591fc98a1R562

I removed the whole redundant block. It should be fine now. Thanks for reviewing this.

magiodev:
> Although fine to add, it's generally redundant to add here for Osmosis' purposes. Can I ask what is the purpose of adding here?

We are adding this for FE purposes, especially for 18 decimals denoms. We still are hardcoding all of this on our FE and it would be nice to be fetching this from an external, shared, source of truth.

PR #3244, opened 2023-11-21
Upgrade Sei to v3.3.0

reversesigh:
https://sei.explorers.guru/proposal/45

https://sei.explorers.guru/block/40562773

### DIFF
--- a/sei/chain.json
+++ b/sei/chain.json
@@ -37,9 +37,13 @@
     "compatible_versions": [
       "v3.3.1"
     ],
-    "ibc_go_version": "v3.3.0",
-    "cosmos_sdk_version": "v0.45.10",
-    "cosmwasm_version": "v0.27.0",
+    "ibc_go_version": "sei-ibc-go/v3 v3.3.0",
+    "cosmos_sdk_version": "sei-cosmos v0.2.63",
+    "consensus": {
+      "type": "sei-tendermint",
+      "version": "v0.2.28"
+    },
+    "cosmwasm_version": "sei-wasmd v0.0.2",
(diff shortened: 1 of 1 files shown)


reversesigh:
@JeremyParish69 - sei uses a lot of customized versions of these modules with their own version tags. I'm using the originals for now rather than the replacements, but I'm wondering what the best way to express these are custom modules are.

Example:
`github.com/CosmWasm/wasmd => github.com/sei-protocol/sei-wasmd v0.0.3`

The base here is `wasmd v0.27.0` but the replacement is `sei-wasmd v0.0.3`. I wouldn't want to just list `v0.0.3` as that could be misleading.

Would this be acceptable instead?
```
"cosmwasm_version": "sei-wasmd v0.0.3"
```

dylanschultzie (approved):

reversesigh:
Refer to https://github.com/cosmos/chain-registry/pull/3245

JeremyParish69:
> @JeremyParish69 - sei uses a lot of customized versions of these modules with their own version tags. I'm using the originals for now rather than the replacements, but I'm wondering what the best way to express these are custom modules are.

Although I don't like that some teams customize their modules like this (especially cosmos_sdk), I think it's better to include the 'sei-' descriptor somewhere so others don't assume it functions exactly as the base module would at the same version number. In cases where there exists a `type` property, I think the sei- part should go there instead of where the numbering is. To tangent off that, if it makes sense, perhaps we should create objects for each of these so `cosmos_sdk` and `cosmwasm` would have an entire object to capture type, version number, whether it's enabled, path, etc.).

PR #3245, opened 2023-11-21
Add "sei-tendermint" to chain.schema.json

reversesigh:
Sei uses their own forked custom modules for consensus, wasm, ibc, etc., In order to properly list "sei-tendermint" with its proper version for sei, we should add this to the chain schema.


JeremyParish69 (approved):
I don't love this but I think it's necessary. A more sustainable solution might need to be developed later on.

PR #3248, opened 2023-11-21
Create IBC source-osmosis.json

Source-Protocol-Cosmos:
https://ping.pub/source/ibc/connection/connection-0


Source-Protocol-Cosmos:
Fixed from source-osmosis to osmosis-source

PR #3250, opened 2023-11-22
Bump EntryPoint version to v1.2.0 and fix gas price

migueldingli1997:
Reference: https://github.com/entrypoint-zone/testnets/tree/main/entrypoint-pubtest-2#testnet-details

### DIFF
--- a/testnets/entrypointtestnet/chain.json
+++ b/testnets/entrypointtestnet/chain.json
@@ -16,9 +16,9 @@
     "fee_tokens": [
       {
         "denom": "ibc/8A138BC76D0FB2665F8937EC2BF01B9F6A714F6127221A0E155106A45E09BCC5",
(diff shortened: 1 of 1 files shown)


testnets/entrypointtestnet/chain.json
```diff
@@ -45,10 +45,10 @@
     },
     "versions": [
       {
-        "name": "v1.1.1",
-        "recommended_version": "v1.1.1",
+        "name": "v1.2.0",
```
JeremyParish69, line 48: Can you please keep v1.1.1, and add v1.2.0 as a new object at the bottom of the `versions` array?

migueldingli1997: Totally missed this message, oops. Should be OK now.

PR #3261, opened 2023-11-25
Add SIX Network Mainnet

dDeedev:
Add SIX Network Mainnet


six/assetlist.json
```diff
@@ -0,0 +1,26 @@
...
+          },
+          {
+            "denom": "six",
+            "exponent": 6
+          }
+        ],
+        "base": "usix",
+        "name": "Six token",
+        "display": "six",
+        "symbol": "six",
```
JeremyParish69, line 20: Please capitalize to SIX

six/assetlist.json
```diff
@@ -0,0 +1,26 @@
...
+        "logo_URIs": {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/sixtestnet/images/six.png"
```
JeremyParish69, line 22: Please point to the mainnet /six/images/ directory instead of the testnet one

JeremyParish69, line 22: This image provided might make an appropriate chain logo (which, currently there is no chain logo defined), but isn't as appropriate for asset logo due to not having square aspect ratio and the text ('SIX Protocol') at the bottom. Perhaps we mean to add another copy of the image but without the text?

six/chain.json
```diff
@@ -0,0 +1,98 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "six",
+  "status": "live",
+  "network_type": "mainnet",
+  "pretty_name": "SIX Protocol",
+  "chain_id": "sixnet",
+  "bech32_prefix": "6x",
```
JeremyParish69, line 8: Please register this prefix

six/assetlist.json
dDeedev: Which one all of them?
1. chain_name
2. name
3. display
4. symbol

six/chain.json
dDeedev: https://github.com/satoshilabs/slips/pull/1637

six/assetlist.json
```diff
@@ -0,0 +1,26 @@
...
+            "denom": "six",
+            "exponent": 6
+          }
+        ],
+        "base": "usix",
+        "name": "Six token",
+        "display": "six",
+        "symbol": "SIX",
+        "logo_URIs": {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/six/images/six.png"
```
JeremyParish69, line 22: we are OK to proceed, but note that due to the aspect ratio of the image not being exactly square, it could end up looking squished on some interfaces that stretch the image to force it to be square.

dDeedev: Ok, that's acceptable for us. Thank you.

PR #3280, opened 2023-11-28
Add Nolus<>Neutron IBC transfer channels for mainnet and testnet

metodi96:


_IBC/nolus-neutron.json
```diff
@@ -0,0 +1,31 @@
+{
+    "$schema": "../ibc_data.schema.json",
+    "chain_1": {
```
JeremyParish69, line 3: not following alphabetical order. Should be Neutron<>Nolus

JeremyParish69: Not only the file names, but also, chain_1 and chain_2 must follow alphabetical order, and the corresponding channels must also be reversed to reflect the order change.

metodi96: Oh, right. Just pushed the changes, should be good now.

PR #3287, opened 2023-11-28
Update chain.json

0xMrmoon:
Add grpc,api,rpc

### DIFF
--- a/jackal/chain.json
+++ b/jackal/chain.json
@@ -262,7 +262,11 @@
       {
         "address": "https://jackal.rpc.silknodes.io",
         "provider": "Silk Nodes"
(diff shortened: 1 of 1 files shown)


jackal/chain.json
```diff
@@ -262,7 +262,11 @@
       {
         "address": "https://jackal.rpc.silknodes.io",
         "provider": "Silk Nodes"
-      }
+      },
+      {
+        "address": "https://jackal.rpc.skynodejs.net",
+        "provider": "skynodejs"
+      },
```
JeremyParish69, line 269: These commas are causing validation checks to fail

PR #3308, opened 2023-11-29
Update pion-1 chain info

pr0n00gler:


testnets/neutrontestnet/chain.json
reversesigh, line 1: Can you add `"next_version_name": v2.0.0` to the end of at least the previous version, and if you can a blank dummy for `"next_version_name":` in the v2.0.0 object?

PR #3315, opened 2023-11-30
Update chain.json

obajay:


reversesigh:
Is this a necessary change? The domain is not case sensitive, and it seems to me its standard to present URLs with lower-case regardless.

obajay:
> Is this a necessary change? The domain is not case sensitive, and it seems to me its standard to present URLs with lower-case regardless.

you follow these two links in different browsers. There is a difference

https://explorer.stavr.tech/aura-mainnet
https://explorer.stavr.tech/Aura-Mainnet

dylanschultzie:
@obajay please use an appropriate title

reversesigh:
> > Is this a necessary change? The domain is not case sensitive, and it seems to me its standard to present URLs with lower-case regardless.
> 
> you follow these two links in different browsers. There is a difference
> 
> https://explorer.stavr.tech/aura-mainnet https://explorer.stavr.tech/Aura-Mainnet

Noted. When I was testing in Chrome, it will default to the lower case page (no matter which I type). But in another browser I was able to see the two different versions. I'll merge but it might be worth considering that some user's browser settings will not respect capitalization.

PR #3335, opened 2023-12-01
lava testnet endpoints by🔥STAVR🔥

obajay:


dylanschultzie:

reversesigh:

testnets/lavatestnet/chain.json
```diff
@@ -116,22 +116,44 @@
         "provider": "Polkachu"
       }
     ]
+  },
```
reversesigh, line 119: I think this line isn't needed.

JeremyParish69:
Sorry to add more, but you're adding endpoints above what looks like the developers' endpoints. Please only added nodes and interfaces to the bottom of arrays.

PR #3338, opened 2023-12-01
Update chain.json

obajay:


dylanschultzie:

reversesigh:

obajay:
> Please return chain.json to the filename `chain.json`

done

testnets/okp4testnet/chain.json
```diff
@@ -97,6 +97,14 @@
         "address": "91.211.251.232:28656"
       }
     ]
+  },
```
reversesigh, line 100: Same as your other PR, I think this line needs to be removed and then your check will pass.

PR #3361, opened 2023-12-04
Add milkTIA token to Osmosis's asset list 

josietyleung:


osmosis/assetlist.json
```diff
@@ -1023,6 +1023,31 @@
...
+          "exponent": 0
+        },
+        {
+          "denom": "milktia",
+          "exponent": 6
+        }
+      ],
+      "base": "factory/osmo1f5vfcph2dvfeqcqkhetwv75fda69z7e5c2dldm3kvgj23crkv6wqcn47a0/umilktia",
+      "name": "milkTIA",
+      "display": "milkTIA",
```
JeremyParish69, line 1041: 'milkTIA' is not defined under denom_units. (looks like this should be 'milktia')

josietyleung: Hey @JeremyParish69 updating the denom_units to be milkTIA  to align

JeremyParish69:
Can I ask why the symbol for this has upper and lower case letters? We'd like to have all token symbols use only uppercase unless they are a derivative of another token. (Is this a liquid staked $TIA perhaps?)

josietyleung:
> Can I ask why the symbol for this has upper and lower case letters? We'd like to have all token symbols use only uppercase unless they are a derivative of another token. (Is this a liquid staked $TIA perhaps?)

Hey yup this is liquid staked $TIA indeed!

JeremyParish69:
Ok, we should add a trace to TIA, in that case. A rather simple change I don't mind taking care of. I just need to know the provider.

josietyleung:
> Ok, we should add a trace to TIA, in that case. A rather simple change I don't mind taking care of. I just need to know the provider.

See that you've added it already. Thank you!

PR #3364, opened 2023-12-05
Remove non-public iris endpoints

hyehan:
Remove non-public iris endpoints from the chain-registry.

### DIFF
--- a/irisnet/chain.json
+++ b/irisnet/chain.json
@@ -8,9 +8,7 @@
   "bech32_prefix": "iaa",
   "daemon_name": "iris",
   "node_home": "$HOME/.iris",
-  "key_algos": [
-    "secp256k1"
-  ],
+  "key_algos": ["secp256k1"],
   "slip44": 118,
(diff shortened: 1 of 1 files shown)


irisnet/chain.json
```diff
@@ -8,9 +8,7 @@
   "bech32_prefix": "iaa",
   "daemon_name": "iris",
   "node_home": "$HOME/.iris",
-  "key_algos": [
-    "secp256k1"
-  ],
+  "key_algos": ["secp256k1"],
```
reversesigh, line 11: Can you edit to maintain the original formatting on this and compatible_versions please?

hyehan: Changed back to the original format, thanks @reversesigh

JeremyParish69: why is this duplicating the key_algos property?

dylanschultzie (changes requested):
@hyehan please remove duplicate `key_algos` property

PR #3369, opened 2023-12-05
Add go version to chain schema

dylanschultzie:


chain.schema.json
```diff
@@ -169,6 +169,11 @@
         "recommended_version": {
           "type": "string"
         },
+        "go_version": {
+          "type": "string",
+          "pattern": "^[0-9]+\\.[0-9]+\\.[0-9]+$",
```
nullmames, line 174: change to `^[0-9]+\\.[0-9]+\\.[0-9]?$` so as to match both `1.19` and `1.19.1` as an example.

dylanschultzie: Oh dang, you're right. I thought new versions ended with `.0`. Thanks!

Screenshot 2023-12-07 at 6 13 39 AM

dylanschultzie: Revised, this is the new regex:

```
^[0-9]+\\.[0-9]+(\\.[0-9]+)?$
```

PR #3377, opened 2023-12-07
add path between dydx and terra2

DreamGallery:


JeremyParish69:
Thank you for the submission, but this was just added via the PR prior https://github.com/cosmos/chain-registry/pull/3376

DreamGallery:
oh, I just want to run the relayer between dydx and terra2 yesterday but didn't found the path. Someone has already submitted a PR, that is great!

PR #3398, opened 2023-12-10
Update chain.json

caotre:


reversesigh:
Approved but please use an appropriate PR title next time briefly describing the change.

PR #3403, opened 2023-12-11
Add POIL token to Juno asset list

zwilling:
Let me know if anything needs to be changed.
Thanks


juno/assetlist.json
```diff
@@ -1879,6 +1879,33 @@
           "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/juno/images/bits.png"
         }
       ]
+    },
+    {
+      "description": "The least memable memecoin in existence. The cats have had their day, it's time for $POIL coin to take reign.",
```
JeremyParish69, line 1884: do you mean 'memeable' instead of 'memable'?

zwilling: Yes, good catch.
Updated the description in the assetlist and cw20 contract.

PR #3417, opened 2023-12-11
Add sDAI and Gravity Bridge sDAI to registry

monet-supply:
Includes images and updated files with sDAI information.


_non-cosmos/ethereum/assetlist.json
```diff
@@ -171,6 +171,49 @@
...
+        {
+          "denom": "sdai",
+          "exponent": 18
+        }
+      ],
+      "base": "0x83F20F44975D03b1b09e64809B757c47f942BEeA",
+      "name": "Savings Dai",
+      "display": "sdai",
+      "symbol": "sDAI",
+      "traces": [
```
JeremyParish69, line 195: Based on a quick Google, this is a wrapped or synthetic DAI? I think we should reference DAI in the traces here.

gravitybridge/assetlist.json
```diff
@@ -296,6 +296,51 @@
...
+      "display": "gsdai",
+      "symbol": "sDAI",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/ethereum/images/sdai.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/ethereum/images/sdai.svg"
+      },
+      "traces": [
+        {
+          "type": "synthetic",
+          "counterparty": {
```
JeremyParish69, line 322: Since there might be traces in between, I think it's probably safest to remove this trace. A full trace can be compiled by external software, but adding an incomplete trace path here may interfere with that since it's ambiguous which origin path to follow.

_non-cosmos/ethereum/assetlist.json
monet-supply: It is tokenized version of Dai deposited into Dai savings rate contract on ethereum, so with respect to DAI it is wrapped I guess. But DAI is a synthetic USD stablecoin, so idk if it's more appropriate to say it's wrapped or synthetic?

monet-supply: Not 100% familiar with how these terms are used here, apologies if I missed something.

gravitybridge/assetlist.json
JeremyParish69: Fix this myself

_non-cosmos/ethereum/assetlist.json
JeremyParish69: Ok. I've made it so that Gravity Bridge sDAI references Ethereum sDAI, and Ethereum sDAI points to Ethereum DAI.

gravitybridge/assetlist.json
```diff
@@ -296,6 +296,47 @@
...
+            "chain_name": "ethereum",
+            "base_denom": "0x83F20F44975D03b1b09e64809B757c47f942BEeA"
+          },
+          "provider": "Gravity Bridge"
+        }
+      ],
+      "images": [
+        {
+          "image_sync": {
```
JeremyParish69, line 331: I've also added an image pointer so we only need the one copy and deleted the duplicate images you'd added to gravitybridge's images folder.

PR #3419, opened 2023-12-11
added union testnet

danbryan:


danbryan:
union uses, `tendermint/PrivKeyBn254` for its key type.

JeremyParish69:
> union uses, `tendermint/PrivKeyBn254` for its key type.

Do we need to add bn254 to the chain schema to the key_algos enum, then?

PR #3426, opened 2023-12-12
Update chain.json

PRO-Delegators:


dydx/chain.json
```diff
@@ -182,7 +186,11 @@
       }
     ],
     "rest": [
-      {
+	  {
```
reversesigh, line 189: Can you fix the formatting on this and the following lines please? Also please use an appropriate title for PRs - like "Update Pro Delegators Endpoints"

PRO-Delegators: Sorry for the formatting issue, i fix it in the last commit, with a better message too.

reversesigh: The formatting was still off. I went ahead and fixed it so you can review the changes for next time

PR #3443, opened 2023-12-15
WhisperNode - add updated mainnet RPC, API, gRPC, & seed nodes

gh0stdotexe:
updated public endpoints for currently supported mainnet networks where we run a validator.


sentinel/chain.json
```diff
@@ -102,10 +102,10 @@
         "provider": "BadgerBite"
       },
       {
-        "id": "1a409560619766355f818ef2e42c935453782635",
+        "id": "c28827cb96c14c905b127b92065a3fb4cd77d7f6",
         "address": "seeds.whispernode.com:17256",
-        "provider": "WhisperNode🤐"
-      }
+        "provider": "WhisperNode 🤐"
+      },
```
JeremyParish69, line 108: this comma will make checks fail

nomic/chain.json
```diff
@@ -71,14 +71,9 @@
...
       },
       {
-        "id": "10beadbcd4bc5fef8a1f5f57353bdb8646f7a554",
-        "address": "seeds.whispernode.com:26656",
-        "provider": "WhisperNode🤐"
-      },
-      {
-        "id": "27b6d74c8408e033e2e5a9e966a0d15782e33596",
-        "address": "seeds.nethernode.xyz:12756",
-        "provider": "carbonZERO🌲"
```
JeremyParish69, line 81: Not sure why CarbonZero nodes are being removed?

sentinel/chain.json
gh0stdotexe: whoops, good looks. will fix.

nomic/chain.json
gh0stdotexe: team is no longer providing support / validating for the chains I cleaned up (they're friends of mine and was just helping them out). can have them submit this change if necessary, but they only validate AssetMantle / Nomic / Stride (governor) at this point, with no additional support being provided currently.

PR #3447, opened 2023-12-15
Add Stake Village

semalis:


composable/chain.json
```diff
@@ -320,6 +337,12 @@
       "kind": "NodeStake Explorer",
       "url": "https://explorer.nodestake.top/composable",
       "tx_page": "https://explorer.nodestake.top/composable/tx/${txHash}"
+    },
+    {
+      "kind": "Stake Village",
+      "url": "https://exp.stakevillage.net/composable",
+      "tx_page": "https://exp.stakevillage.net/composable/tx/${txHash}",
+      "account_page": "https://exp.stakevillage.net/composable/account/{$accountAddress}"
```
JeremyParish69, line 345: the dollar sign $ is supposed to be outside and before the bracket {.

JeremyParish69: looks like some other explorers you've added (but not all) have the same issue

PR #3455, opened 2023-12-16
add $CORGI / Baby Corgi to Neutron

BabyCorgiToken:
add $CORGI / Baby Corgi to Neutron


neutron/assetlist.json
```diff
@@ -366,6 +366,34 @@
...
+          ]
+        },
+        {
+          "denom": "corgi",
+          "exponent": 6
+        }
+      ],
+      "base": "factory/neutron1tklm6cvr2wxg8k65t8gh5ewslnzdfd5fsk0w3f/corgi",
+      "name": "Baby Corgi",
+      "display": "corgi",
```
dylanschultzie, line 387: I believe this needs to be in all caps.

dylanschultzie (changes requested):

neutron/assetlist.json
JeremyParish69: Display does not need to be caps

PR #3456, opened 2023-12-16
Fabo/change milkTIA base denom to IBC trace

faboweb:
Instead of using the base denom of the token on the native chain we need to provide the base denom of the token on Osmosis


JeremyParish69:
Thinking it over. osmosis' TIA gets deposited, but in the backend it transferres back to celestia and stakes TIA, so maybe it should stay as is.

JeremyParish69:
Closing for now. Thank you

PR #3458, opened 2023-12-17
add doravota ibc channel: cosmoshub, nolus, coreum

tomas-dora:


tomas-dora:
@JeremyParish69 Hi, We have new ibc channel that need to be merged ASAP, thanks!

PR #3460, opened 2023-12-18
feat(teritori): Update Assetlist with IBC denoms

omniwired:


teritori/assetlist.json
```diff
@@ -29,6 +29,173 @@
...
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "cosmoshub",
+            "base_denom": "uatom",
+            "channel_id": "channel-10"
+          },
+          "chain": {
+            "channel_id": "channel-431",
+            "path": "transfer/channel-431/uatom"
```
JeremyParish69, line 109: For all these assets I think you put the channel_id with the opposite chain/counterparty channel_id property. I think cosmoshub/counterparty would be channel-431 and chain channel_id would be channel-10. The path also doesn't yield the ibc/hash denom and swapping the channel order would fix this.

teritori/assetlist.json
```diff
@@ -29,6 +29,173 @@
...
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "juno",
+            "base_denom": "ujuno",
+            "channel_id": "channel-11"
+          },
+          "chain": {
+            "channel_id": "channel-164",
```
JeremyParish69, line 140: Adding this requires there to be a juno-teritori ibc connection registered.

teritori/assetlist.json
```diff
@@ -29,6 +29,173 @@
...
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/osmo.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/osmo.svg"
+      },
+      "coingecko_id": "osmosis",
+      "keywords": [
+        "dex",
+        "staking"
+      ],
+      "images": [
+        {
```
JeremyParish69, line 74: Please add an image_sync here (e.g., https://github.com/cosmos/chain-registry/blob/e1a5d27659072d77a5f5ecca42ea3c7969bd3a37/dyson/assetlist.json#L65)

omniwired:
@JeremyParish69 thank you always for your review, I made the requested changes

PR #3468, opened 2023-12-19
add 3 andromeda IBC channels

ertemann:


ertemann:
Made by L5, also supported by Crosnest

PR #3469, opened 2023-12-19
add doravota incentive testnet

tomas-dora:


testnets/doravotaincentivetestnet/chain.json
```diff
@@ -0,0 +1,88 @@
+{
+  "$schema": "../../chain.schema.json",
+  "chain_name": "doravotaincentivetestnet",
```
JeremyParish69, line 3: can we please change this to `"chain_name": "doravotatestnet2"`?

tomas-dora: good idea. I'm already changed

PR #3488, opened 2023-12-22
AUTISM on Osmosis

PFC-developer:
I think i've got the right type 'ics20' here

### DIFF
--- a/osmosis/assetlist.json
+++ b/osmosis/assetlist.json
@@ -1169,6 +1169,53 @@
         "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/ethereum/images/page.png",
         "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/ethereum/images/page.svg"
       }
+    },
+    {
+      "description": "AUTISM on Osmosis",
+      "denom_units": [
+        {
+          "denom": "ibc/9DDF52A334F92BC57A9E0D59DFF9984EAC61D2A14E5162605DF601AA58FDFC6D",
+          "exponent": 0,
+          "aliases": [
+            "uautism"
+          ]
+        },
+        {
+          "denom": "autism",
+          "exponent": 6
+        }
+      ],
+      "type_asset": "ics20",
+      "base": "ibc/9DDF52A334F92BC57A9E0D59DFF9984EAC61D2A14E5162605DF601AA58FDFC6D",
+      "name": "Injective AUTISM",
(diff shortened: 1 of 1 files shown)


osmosis/assetlist.json
```diff
@@ -1169,6 +1169,53 @@
...
+            "channel_id": "channel-8",
+            "path": "transfer/channel-8/factory/inj14lf8xm6fcvlggpa7guxzjqwjmtr24gnvf56hvz/autism"
+          }
+        }
+      ],
+      "images": [
+        {
+          "image_sync": {
+            "chain_name": "injective",
+            "base_denom": "utia"
```
dylanschultzie, line 1211: This doesn't seem right?

dylanschultzie (changes requested):

osmosis/assetlist.json
dylanschultzie: @JeremyParish69 can you advise? I'm bad at these.

JeremyParish69: It shouldn't be trying to synchronize to the TIA image. It should be pointing to the original AUTISM image. So base_denom is wrong here

PFC-developer: I'll fix shortly.. does everything else  look ok?
I haven't done a token factory one before

osmosis/assetlist.json
```diff
@@ -1169,6 +1169,53 @@
...
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "injective",
+            "base_denom": "factory/inj14lf8xm6fcvlggpa7guxzjqwjmtr24gnvf56hvz/autism",
+            "channel_id": "channel-122"
+          },
+          "chain": {
+            "channel_id": "channel-8",
+            "path": "transfer/channel-8/factory/inj14lf8xm6fcvlggpa7guxzjqwjmtr24gnvf56hvz/autism"
```
JeremyParish69, line 1203: This path doesn't yield the ibc hash denom above when run through a sha256 function. But it does when using channel 122. Either the channels are backwards or the base hash is wrong

JeremyParish69: yep, channels are backwards

osmosis/assetlist.json
JeremyParish69: Other than my other comment, I think everything else is fine

osmosis/assetlist.json
PFC-developer: thx. box changes have been done.

PR #3517, opened 2023-12-29
ADD chihuahuawifhat/baddog.png to chihuahua chain registry

SmolFatality:


chihuahua/assetlist.json
```diff
@@ -57,5 +57,31 @@
...
+          "exponent": 0
+        },
+        {
+          "denom": "achihuahuawifhat",
+          "exponent": 6
+        }
+      ],
+      "base": "factory/chihuahua1x4q2vkrz4dfgd9hcw0p5m2f2nuv2uqmt9xr8k2/achihuahuawifhat",
+      "name": "Chihuahuawifhat",
+      "display": "chihuahuawifhat",
```
JeremyParish69, line 76: there is no unit of this name defined. Is this supposed to be 'achihuahuawifhat'?

chihuahua/assetlist.json
```diff
@@ -57,5 +57,31 @@
         "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/chihuahua/images/puppyhuahua_logo.png"
       }
     }
-  ]
-}
\ No newline at end of file
+  ],
+
```
JeremyParish69, line 61: I think you messed up the formatting a bit. Can you please restore this to the correct schema?

chihuahua/assetlist.json
SmolFatality: Yes sorry,

chihuahua/assetlist.json
SmolFatality: Yes

chihuahua/assetlist.json
JeremyParish69: I think the name was better before. And the `display` value is still invalid.

chihuahua/assetlist.json
JeremyParish69: I think the issue was that you added another asset object but outside the `assets` array. This is still not following the schema correctly

SmolFatality: okay, i see what you mean, i'll correct it need to edit it.

SmolFatality: corrected. hope this is the last review you will have to do. sorry in advance!

PR #3531, opened 2024-01-03
fix lava testnet to testnet-2

omerlavanet:
the current entry is outdated

### DIFF
--- a/testnets/lavatestnet2/chain.json
+++ b/testnets/lavatestnet2/chain.json
@@ -4,8 +4,8 @@
   "status": "live",
   "network_type": "testnet",
   "website": "https://www.lavanet.xyz/",
-  "pretty_name": "Lava",
-  "chain_id": "lava-testnet-1",
+  "pretty_name": "Lava Testnet",
+  "chain_id": "lava-testnet-2",
   "bech32_prefix": "lava@",
   "daemon_name": "lavad",
   "node_home": "$HOME/.lava",
@@ -17,8 +17,8 @@
     "fee_tokens": [
       {
         "denom": "ulava",
-        "fixed_min_gas_price": 0.0025,
-        "low_gas_price": 0.0025,
+        "fixed_min_gas_price": 0.000001,
+        "low_gas_price": 0.000001,
         "average_gas_price": 0.025,
         "high_gas_price": 0.04
       }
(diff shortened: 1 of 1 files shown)


testnets/lavatestnet2/chain.json
```diff
@@ -56,91 +56,9 @@
     ],
     "genesis": {
       "name": "v0.3.0",
-      "genesis_url": "https://raw.githubusercontent.com/lavanet/lava-config/main/testnet-1/genesis_json/genesis.json"
+      "genesis_url": "https://raw.githubusercontent.com/lavanet/lava-config/main/testnet-2/genesis_json/genesis.json"
     },
     "versions": [
```
JeremyParish69, line 61: Please add the current version (v0.32.0) to the `versions` [] array

omerlavanet: thanks, done

JeremyParish69:
Looking over this, I think we need to revert this PR. I think this was meant to be an archived chain with id lava-testnet-1, whereas lava-testnet-2 has chain_name: "lavatestnet". I think most the properties here are meant to be reverted, the status set to killed, and the chain_name changed to lavatestnet1 to properly archive this as having been abandoned post a hard-fork--the one leading to the creating of lava-testnet-2.
Does this sound acceptable? For now, to correct the duplicate chain-id problem, I need to at least set the chain_id back to lava-testnet-1, but it aught to be fully archived.

omerlavanet:
> Looking over this, I think we need to revert this PR. I think this was meant to be an archived chain with id lava-testnet-1, whereas lava-testnet-2 has chain_name: "lavatestnet". I think most the properties here are meant to be reverted, the status set to killed, and the chain_name changed to lavatestnet1 to properly archive this as having been abandoned post a hard-fork--the one leading to the creating of lava-testnet-2. Does this sound acceptable? For now, to correct the duplicate chain-id problem, I need to at least set the chain_id back to lava-testnet-1, but it aught to be fully archived.

yes, makes sense. we can kill the lava-testnet-1 and create a new lava-testnet-2, what triggered the need for the change?

JeremyParish69:
The trigger for this was the addition of automation to detect chain-id conflicts. The chain-id for multiple chains had lava-testnet-2 which will no longer be allowed at the chain registry.
I'd like to essentially revert the changes in this PR to restore the version history and other details of lavatestnet1, plus set its status to 'killed'(, assuming it to be the pre-fork chain of lava-testnet-2), and to update the chain_name for lavatestnet2 to just 'lavatestnet'. This way, no matter which version of Lava Testnet is current, it will always keep the chain_name: lavatestnet.  This is how other chains have handled archiving pre-fork chains.

JeremyParish69:
This is the PR I propose to correct the handling of the pre-form chain instance:
https://github.com/cosmos/chain-registry/pull/4201

PR #3542, opened 2024-01-04
Update firmachain peer by ramuchi

ramuchi:


JeremyParish69:
Please use a descriptive, non-generic PR title.

ramuchi:
ok

JeremyParish69:
> ok

This is also a poor title. There is no peer.json. It doesn't communicate anything. You could maybe use: 'Update firmachain peer by ramuchi'

ramuchi:
Ok👌

PR #3566, opened 2024-01-07
Update BeeZee chain json

faneaatiku:


beezee/chain.json
```diff
@@ -32,33 +32,33 @@
...
+      "windows/amd64": "https://github.com/bze-alphateam/bze/releases/download/v6.1.0/bze-6.1.0-win64.zip"
     },
     "genesis": {
       "genesis_url": "https://raw.githubusercontent.com/bze-alphateam/bze/main/genesis.json"
     },
     "versions": [
       {
-        "name": "v6.0.0",
-        "recommended_version": "v6.0.0",
+        "name": "v6.1.0",
```
JeremyParish69, line 51: Within the versions array, please add the next_version_name property and instead add a new version to the bottom of the list

faneaatiku: checked cosmoshub to really understand what you meant. Basically versions should be the exact "description" on how to sync from scratch. Will modify as required asap.

faneaatiku: Done! Thanks for the help.

PR #3567, opened 2024-01-07
Add Lambo Token

wenlambotron:
Add Lambo Token


neutron/assetlist.json
```diff
@@ -472,6 +472,32 @@
       "images": [
         {
           "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/circus.png"
+        },
```
JeremyParish69, line 475: Missing some brackets so the schema is broken

JeremyParish69:
PR seems abandoned. Can be reopened if needed

PR #3571, opened 2024-01-08
Add WTF Token

whatthefuckcoin:
Add WTF Token


JeremyParish69:
This PR breaks the schema--please repair

neutron/assetlist.json
```diff
@@ -500,6 +500,32 @@
       "images": [
         {
           "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/jimmy.png"
+        },
+        {
```
JeremyParish69, line 504: Schema is broken here. This should not be a continuation of Jimmy's images, but rather a new asset.

PR #3578, opened 2024-01-08
adding Wormhole wrapped USDC from Avalanche

rupebac:
Adding the Wormhole USDC wrapper for Avalanche bridged USDC. An example of a Portalbridge transaction can be found here:

https://wormholescan.io/#/tx/0x7baf631114042d8da16cef66aeec331fbc58374bb32ed42d0cbb463b9105ebae

transaction in osmosis side:

https://www.mintscan.io/osmosis/tx/B13881F5CA02555DEA80CAE58DBFFAD433CEE03D855848D08543500EEA8ED006?height=13106768

transaction in Avalanche side:

https://snowtrace.io/tx/0x7baf631114042d8da16cef66aeec331fbc58374bb32ed42d0cbb463b9105ebae?chainId=43114


osmosis/assetlist.json
```diff
@@ -165,6 +165,50 @@
...
+      "display": "avalanche-usdc.hole",
+      "symbol": "avaxUsdc.hole",
+      "traces": [
+        {
+          "type": "bridge",
+          "counterparty": {
+            "chain_name": "avalanche",
+            "base_denom": "0xB97EF9Ef8734C71904D8002F8b6Bc66Dd9c48a6E"
+          },
+          "provider": "Wormhole"
```
JeremyParish69, line 194: This is good but not quite right. There is a no direct Wormhole bridge from Avalanche to Osmosis. We'll need to register this avalanche.USDC asset under the /gateway/ chain, defining this Wormhole bridge in the traces there. And then we may define the asset (avalanche.USDC.wh) here under /osmosis/, and the trace on this one would point to gateway::avalanche.USDC.

osmosis/assetlist.json
```diff
@@ -165,6 +165,50 @@
...
+        }
+      ],
+      "images": [
+        {
+          "image_sync": {
+            "chain_name": "osmosis",
+            "base_denom": "avalanche-uusdc"
```
JeremyParish69, line 201: this is a circular reference??

osmosis/assetlist.json
```diff
@@ -165,6 +165,50 @@
...
+          },
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/usdc.hole.png",
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/usdc.hole.svg"
+        }
+      ],
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/usdc.hole.png",
```
JeremyParish69, line 208: Does Wormhole use the same logo for USDC from Ethereum and for USDC from Avalanche?

osmosis/assetlist.json
rupebac: sorry, was a copy paste leftover. I don't think we need to add any image_sync right?

osmosis/assetlist.json
rupebac: they just use a generic USDC logo, at least in portalbridge.... Do you want to use the same generic logo? I find more explicit a specific one for Avalanche, since the unwrapped liquidity rest there. But you tell me :)

osmosis/assetlist.json
rupebac: alright, let me attempt that.

rupebac: done, how does it look like now? I am only dubious about the images, can you check that up, in both the gateway and osmosis side? thanks!

rupebac:
please, review again. I added the chain to the gateway, in a similar way as it is done for Kujira and Sui...

Thanks!

gateway/assetlist.json
```diff
@@ -504,6 +504,48 @@
...
+      "address": "wormhole1utjx3594tlvfw4375esgu72wa4sdgf0q7x4ye27husf5kvuzp5rsr72gdq",
+      "base": "factory/wormhole14ejqjyq8um4p3xfqj74yld5waqljf88fz25yxnma0cngspxe3les00fpjx/5ZLmAZpcbaP4EGyihSmpfwryzDr84h51tboV392BCjW4",
+      "display": "wormhole/5ZLmAZpcbaP4EGyihSmpfwryzDr84h51tboV392BCjW4/6",
+      "name": "Avalanche USD Coin (Wormhole)",
+      "symbol": "avalanche.USDC.wh",
+      "traces": [
+        {
+          "type": "bridge",
+          "counterparty": {
+            "chain_name": "osmosis",
```
JeremyParish69, line 530: In this gateway file, the trace here should point to avalanche USDC (not to Osmosis). Luckily, Avalanche USDC has already been registered to the Chain Registry, so I think you just need to update the chain_name and base_denom here.

gateway/assetlist.json
```diff
@@ -504,6 +504,48 @@
...
+            "base_denom": "ibc/0B3C3D06228578334B66B57FBFBA4033216CEB8119B27ACDEE18D92DA5B28D43"
+          },
+          "provider": "Wormhole"
+        }
+      ],
+      "images": [
+        {
+          "image_sync": {
+            "chain_name": "osmosis",
```
JeremyParish69, line 539: image sync likely shouldn't point to osmosis (downstream), but preferably upstream to avalanche USDC (if they just use the USDC logo)

osmosis/assetlist.json
JeremyParish69: I think this is fine for now, but Osmosis will have to create a new image for this asset anyway, and I'll update this once it gets added.

gateway/assetlist.json
rupebac: ok, just changed it

osmosis/assetlist.json
rupebac: alright, thanks!

PR #3591, opened 2024-01-09
add dydx testnet

danbryan:
Worked with the DYDX team to get the correct chain info for their persistent testnet.


testnets/dydxtestnet/assetlist.json
```diff
@@ -0,0 +1,86 @@
+{
+  "$schema": "../assetlist.schema.json",
+  "chain_name": "dydx",
```
JeremyParish69, line 3: Chain name cannot be dydx. Please use dydxtestnet

testnets/dydxtestnet/chain.json
```diff
@@ -0,0 +1,155 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "dydx",
```
JeremyParish69, line 3: Chain name cannot be dydx. Please use dydxtestnet

PR #3592, opened 2024-01-09
Add status page for Neutron

keylogic1:
To provide up-to-date information on the status of our RPCs we have included a dedicated page to be able to provide updates.


JeremyParish69:
I don't think we have a space for this here. It would be more appropriate for an interface registry

PR #3606, opened 2024-01-10
Update chain.json

alibabaedge:
Change validator alias


reversesigh:
Please use an appropriate title for your PR

PR #3634, opened 2024-01-12
Add Sei-Nois to ibc

purpleguy99:


_IBC/sei-nois.json
```diff
@@ -0,0 +1,47 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
```
JeremyParish69, line 3: Please use alphabetical order to name the file and to determine which chain is chain1 and which is chain2. This needs to be reversed

purpleguy99: Done!

PR #3642, opened 2024-01-16
Add DSR

dinosaurgamelol:


neutron/assetlist.json
```diff
@@ -560,6 +560,30 @@
       ]
     },
     {
+      "description": "Visit Dinosaurgame.lol",
```
JeremyParish69, line 563: Please do no use URLs in the description. Please repair the schema

JeremyParish69 (changes requested):
This addition breaks the schema. Please repair it and address the comments

JeremyParish69:
Submitter is ignoring comments and re-requesting reviews without any changes or attempt to communicate.

PR #3669, opened 2024-01-19
Add multi tokenfactory on chihuahua

atmoner:


chihuahua/assetlist.json
```diff
@@ -106,7 +106,107 @@
...
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/chihuahua/images/bdog.png"
+      },
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/chihuahua/images/bdog.png"
+        }
+      ]
+    },
+    {
+      "description": "Bulldog",
```
JeremyParish69, line 186: Is this description perhaps incorrect? Copied from the Bulldog token and not updated for this one?

atmoner: True! Fixed here: https://github.com/cosmos/chain-registry/pull/3669/commits/be33c36f898a8d125f64478b9d580f435408b7ef
Thanks o/

PR #3713, opened 2024-01-26
Remove Asset from Bitsong

angelorc:
remove discontinued fantokens

### DIFF
--- a/bitsong/assetlist.json
+++ b/bitsong/assetlist.json
@@ -380,31 +380,6 @@
           "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/bitsong/images/ftD4B6290EDEE1EC7B97AB5A1DC6C177EFD08ADCC3.png"
         }
       ]
-    },
-    {
-      "description": "L DON a BitSong Music FanToken",
-      "denom_units": [
-        {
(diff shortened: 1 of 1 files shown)


angelorc:
I agree with you. However, the artist had an agreement with us and wishes for their image and name to be removed. We would also prefer it to be taken off the asset lists, so it doesn't appear among the other tokens. Moreover, in a few days, artists will be able to customize metadata using the URI field available on the fan token itself within the BitSong chain

JeremyParish69:
Perhaps we can decommission it without removing it to give artists freedom to remove their identity. Could we update the name, description, and images to remove their identity, and use placeholder values instead?

JeremyParish69:
@angelorc Please see this PR: https://github.com/cosmos/chain-registry/pull/3760
I think this new property could signal to interfaces to exclude this asset by default.

angelorc:
That's good, but we should remove also the name and the image, the artist no longer wants any reference to him.

JeremyParish69:
> That's good, but we should remove also the name and the image, the artist no longer wants any reference to him.

The images were already removed. But I've made a few more changes, removing more of the references to them, substituting a generic value or [Redacted] where their name was. The only remaining reference now is in the symbol property.

angelorc:
Thank you Jeremy

PR #3714, opened 2024-01-26
RAPTR Token

Rapture-INC:
First ever P2P insurance platform on $OSMO


osmosis/assetlist.json
```diff
@@ -1521,6 +1521,32 @@
         }
       ] 
     },
+    {
```
JeremyParish69, line 1524: Please move to the bottom of the list

Rapture-INC:
Done @JeremyParish69

PR #3720, opened 2024-01-28
Added Starname REST & RPC

jo-chainmasters:


starname/chain.json
```diff
@@ -75,6 +75,10 @@
   },
   "apis": {
     "rpc": [
+      {
```
JeremyParish69, line 78: Please move these additions to the bottom

jo-chainmasters: Done

PR #3722, opened 2024-01-29
Added axlUSDC and noble USDC as fee tokens for Terra

terencelimzhengwei:
Majority of validators have updated their config to support axlUSDC and noble USDC on Terra.

https://commonwealth.im/terra/discussion/14892-add-stablecoins-as-gas-fee-tokens-on-terra).


JeremyParish69:
Thanks for the addition. Please define these in the asssetlist first so we know which assets these are.

PR #3732, opened 2024-01-30
Page token coingecko_id update

rbpollock:
- Added coingecko_id to ethereum, gravitybridge, and osmosis assetlists
- Removed ".grv" from symbol on osmosis (was being duplicated in frontends)


JeremyParish69:
Please only add the cgid to the original asset listed under Ethereum, but not any others

rbpollock:
> Please only add the cgid to the original asset listed under Ethereum, but not any others

I've removed the other commits and added the symbol update for osmosis

PR #3747, opened 2024-01-31
Add Panthea EU endpoints for Shentu

jarnek73:
Added our new endpoints


reversesigh:
Please use an appropriate PR title such as "Add Panthea EU endpoints for Shentu"

PR #3748, opened 2024-01-31
Fix: remove duplicated info for Teritori

DavideSegullo:
In the teritori chain configuration file, the codebase information was repeated twice


reversesigh:
Good catch - thank you!

PR #3760, opened 2024-02-02
Add Deprecated Asset property

JeremyParish69:
Add Deprecated Asset property
Deprecate Bitsong::L DON, and remove it's images


JeremyParish69:
@angelorc Sorry to keep bothering you, just looking to close this out. I hope the additional steps taken to decommission/deprecate/redact the asset is more acceptable?

angelorc:
looks good!

PR #3788, opened 2024-02-06
Update chain.json

takeshi-val:


dymension/chain.json
```diff
@@ -72,6 +72,10 @@
   },
   "apis": {
     "rpc": [
+      {
```
JeremyParish69, line 75: Please move to bottom of the list

JeremyParish69:
Please use a descriptive, non-generic PR title

PR #3804, opened 2024-02-07
Add RPC/API for chain Dymension, by PRO Delegators

PRO-Delegators:


dymension/chain.json
```diff
@@ -72,6 +72,10 @@
   },
   "apis": {
     "rpc": [
+      {
```
JeremyParish69, line 75: Please move these to the bottom of the list

PRO-Delegators (commented):
done

PR #3815, opened 2024-02-08
Update Osmosis assetlist.json 3

JeremyParish69:
Update Osmosis assetlist.json 3
Adds many of the assets previously missing from this assetlist. Includes everything listed on osmosis-labs' assetlists repo, plus a few extra that are only registered here.

The process still cannot be automated because the generated output is missing the schema property and the assets that are only registered here in the Chain Registry (but not on Osmosis Zone) 

One glaring issue is the inclusion of CoinGecko ID, even though it should be omitted for all ics20 assets.


JeremyParish69:
Replaced by #3823

PR #3818, opened 2024-02-08
Add NODEJUMPER rpc/api/grpc enpoints for Dymension

vbloher:


dymension/chain.json
```diff
@@ -148,6 +148,10 @@
         "address": "https://rpc.dymension.silentvalidator.com:443",
         "provider": "silent"
       },
+      {
```
JeremyParish69, line 151: Please add to bottom of the arrays

vbloher: done

PR #3843, opened 2024-02-10
Update chain.json  Shentu by 🔥STAVR🔥

obajay:


shentu/chain.json
```diff
@@ -161,6 +161,11 @@
         "address": "shentu.ramuchi.tech:56656",
         "provider": "ramuchi.tech"
       },
+      {
```
JeremyParish69, line 164: Please instead add these to the bottom of their respective array

PR #3850, opened 2024-02-11
Include a top-level chainlists

refundvalidator:
This PR proposes to add top-level chain lists for both mainnets, and testnets. This would allow for simpler querying when retrieving data directly from GitHub.


ankurdotb:
Seems redundant since Github provides an API to [fetch the contents of a repository as JSON](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content)? Although yes, I do understand that the JSON generated by this bash script will be simpler than the one from Github's API. I'm just not sure how big of a problem it is or not addressed downstream in clients using Github-specific packages for parsing this repo.

refundvalidator:
> Seems redundant since Github provides an API to [fetch the contents of a repository as JSON](https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28#get-repository-content)? Although yes, I do understand that the JSON generated by this bash script will be simpler than the one from Github's API. I'm just not sure how big of a problem it is or not addressed downstream in clients using Github-specific packages for parsing this repo.

The main thing I'm trying to address is automation. If somebody wanted to query every mainnet/testnet from GitHub or another platform hosting the repo, they would need to manually specify each chains' endpoint. With something like this, it could be inferred automatically.

Just as an example, in python, we have to do something like the following to get the data from each chain:
```python
# Query all chain data
url = "https://raw.githubusercontent.com/cosmos/chain-registry/master"
chains = [ "cosmoshub", "osmosis", "celestia", "etc" ] 
result = [ ] 
for chain in chains:
    result.append(requests.get(f'{url}/{chain}/chain.json').json())
```

This is very rigid, so the only way to do this without cloning/constantly updating a local repo  or web scraping would be to hardcode each chain. With the approach im suggesting, we could automatically parse any new chains when they are added:

```python
chains_url = "https://raw.githubusercontent.com/cosmos/chain-registry/master/mainnets.json"
url = "https://raw.githubusercontent.com/cosmos/chain-registry/master"
chains = requests.get(url).json
result = [ ] 
for chain in chains["mainnets"]:
    result.append(requests.get(f'{url}/{chain}/chain.json')
```

JeremyParish69:
IMO it's extra complexity, and maintenance, and makes structure changes harder to implement. I think there are several tools that can derive chain names without needing this top-level chainlist.

JeremyParish69:
For now I think we should leave this to tooling. Can be brought up for discussion again, but for now I'm closing this PR because it's just sitting here.

PR #3851, opened 2024-02-11
Add AgoraNodes Loadbalanced Endpoints for Dymension Mainnet

AgoraNodes:
Adding load balanced end points from Agora Nodes

### DIFF
--- a/dymension/chain.json
+++ b/dymension/chain.json
@@ -97,6 +97,10 @@
   "apis": {
     "rpc": [
       {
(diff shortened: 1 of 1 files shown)


dymension/chain.json
```diff
@@ -174,6 +178,10 @@
       }
     ],
     "rest": [
+      {
+        "address": "https://dym-m-api.agoranodes.com/"
```
reversesigh, line 182: Missing a comma here

AgoraNodes: fixed ty

PR #3858, opened 2024-02-12
updates humans IBC

0x963D:


JeremyParish69:
IBC channels cannot be updated without governance approval on both side, or unless it can be proven that no transferred tokens would be stuck on the counterparty chain.

0x963D:
No transfered tokens will be stuck as can be seen on Mintscan:

0$
0 txn

https://www.mintscan.io/analytics/network/relayers/trends?source=humans

PR #3862, opened 2024-02-13
Updated IBC connection details

ashishchandr70:
Update Saga Testnet IBC Connection to Osmosis Testnet 5 due to expired IBC client.


JeremyParish69:
There was concern expressed over being able to make use of the existing TSAGA tokens on Osmosis. Abandoning the old connection will make the tokens lose relevance and likely be trapped forever. Are we sure we want to change channels rather than try to recover the expired connection?

JeremyParish69:
No longer needed.

PR #3869, opened 2024-02-14
Update odin-osmosis IBC info,  add 2 + change 2 Odin tokens in Osmosis assets

Cordtus:
- updated IBC clients in /_IBC/odin-osmosis
- added 2 tokens to /osmosis/assetlist - [MYRK, DOKI]
- set 'osmosis_unlisted' status for the above
- removed 'osmosis_unstable' status from existing Odin tokens

### DIFF
--- a/_IBC/odin-osmosis.json
+++ b/_IBC/odin-osmosis.json
@@ -2,12 +2,12 @@
   "$schema": "../ibc_data.schema.json",
   "chain_1": {
     "chain_name": "odin",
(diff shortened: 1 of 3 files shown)


JeremyParish69:
Thanks for the update! Please don't update the Osmosis assetlist here; the changes will only be overwritten and removing "osmosis-unstable" would not affect the Osmosis frontend.

Cordtus:
Sorry, restored Osmosis assetlist.json to current state in last commit.

PR #3871, opened 2024-02-14
Add $BADKID (IBC) (Injective - Neutron)

Dizz02:


injective/assetlist.json
```diff
@@ -285,6 +285,50 @@
...
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "osmosis",
+            "base_denom": "factory/osmo10n8rv8npx870l69248hnp6djy6pll2yuzzn9x8/BADKID",
+            "channel_id": "channel-8"
+          },
+          "chain": {
+            "channel_id": "channel-122",
```
JeremyParish69, line 315: I think you have these channels backwards (it's 122 on the osmosis side, not the injective side), which is further evidenced by the calculated ibc denoms not matching what you have here. I think you have the same issue with the other asset on the neutron assetlist

PR #3877, opened 2024-02-15
fix ibc - fix $BADKID img

Dizz02:


injective/assetlist.json
```diff
@@ -285,6 +285,50 @@
...
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "osmosis",
+            "base_denom": "factory/osmo10n8rv8npx870l69248hnp6djy6pll2yuzzn9x8/BADKID",
+            "channel_id": "channel-122"
+          },
+          "chain": {
+            "channel_id": "channel-8",
+            "path": "transfer/channel-8/factory/factory/osmo10n8rv8npx870l69248hnp6djy6pll2yuzzn9x8/BADKID"
```
JeremyParish69, line 316: you have /factory/ twice (i.e.e, "/factory/factory/"), which is incorrect.

injective/assetlist.json
```diff
@@ -285,6 +285,50 @@
...
+          }
+        }
+      ],
+      "images": [
+        {
+          "image_sync": {
+            "chain_name": "osmosis",
+            "base_denom": "factory/osmo10n8rv8npx870l69248hnp6djy6pll2yuzzn9x8/BADKID"
+          },
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/badkid2.png"
```
JeremyParish69, line 326: If you could please have the replacement image badkid.png to use the same name. This way, apps that have already integrated the badkid image won't lose the reference file.

Dizz02: Done

PR #3897, opened 2024-02-17
Upgrade Planq to v1.1.0

kii-awesome:
Software upgrade without props

### DIFF
--- a/planq/chain.json
+++ b/planq/chain.json
@@ -73,26 +73,6 @@
       },
       {
         "name": "v1.0.7",
-        "recommended_version": "v1.0.7",
-        "compatible_versions": [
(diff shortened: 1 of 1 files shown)


reversesigh:
Was this a halt-height upgrade?

kii-awesome:
Nope, Basically a fork . The operator must upgrade before blockheight 7516700 or it will get apphash.

reversesigh:
Based on my understanding your update to the codebase is correct but instead of adding v1.1.0 to the version history after v1.0.7, the changes need to be made within v1.0.7 instead. The version name is still v1.0.7, but the recommended and compatible binary should be v1.1.0.

PR #3908, opened 2024-02-18
Add SRCX in Planq assets

ivanshukhov:


planq/assetlist.json
```diff
@@ -29,6 +29,32 @@
           "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/planq/images/planq.svg"
         }
       ]
+    },
+    {
+      "description": "The token of Source Protocol bridged from BSC.",
```
JeremyParish69, line 34: Since this is bridged, can we added the traces[] point back to it origin? I imagine it might be something like Source -> BSC > Planq?

ivanshukhov: I'm a bit unsure what to do here. It's an ERC20 token originating from BSC. This got then bridged to Planq.
So for now its BSC > Planq.

The traces def will fall out of schema since its not originating from ibc. This is what I can add though:

"traces": [
        {
          "type": "erc20",
          "counterparty": {
            "chain_name": "bsc",
            "base_denom": "erc20/srcx",
            "channel_id": "0"
          },
          "chain": {
            "channel_id": "",
            "path": ""
          }
        }
      ],

Let me know, thanks

JeremyParish69: This is not quite right. the type: would be `bridge`; there is a $def for non-ibc transfers. It take bridge provider and contract addresses.
The srcx token will need to be defined in /non-cosmos/binancesmartchain/assetlist.json.
Just to make sure, the Source Protocol here is native to BSC, and it's not bridged from some other network to get to BSC?

ivanshukhov: >Just to make sure, the Source Protocol here is native to BSC, and it's not bridged from some other network to get to BSC?
Correct.

Made adjustments like you suggested.

planq/assetlist.json
```diff
@@ -29,6 +29,46 @@
...
+        },
+        {
+          "denom": "srcx",
+          "exponent": 9
+        }
+      ],
+      "base": "erc20/0x091F9A57A3F58d758b6572E9d41675918EAC7F09",
+      "name": "Source Token",
+      "display": "srcx",
+      "coingecko_id": "source-protocol",
```
JeremyParish69, line 48: Please remove the coingecko_id here. It will be defined only in bsc's assetlist

planq/assetlist.json
```diff
@@ -29,6 +29,46 @@
...
+          "provider": "DeltaSwap.io"
+        }
+      ],
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/planq/images/srcx.png"
+      },
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/planq/images/srcx.png"
+        }
```
JeremyParish69, line 70: Can you please add the image only to: /non-cosmos/binancesmartchain/images/, and in this Planq-bridged asset, please use image_sync to point to the image in the bsc image folder

_non-cosmos/binancesmartchain/assetlist.json
```diff
@@ -174,6 +174,34 @@
...
+      ],
+      "type_asset": "erc20",
+      "address": "0x454b90716a9435e7161a9aea5cf00e0acbe565ae",
+      "base": "0x454b90716a9435e7161a9aea5cf00e0acbe565ae",
+      "name": "Source Token",
+      "display": "SRCX",
+      "symbol": "SRCX",
+      "coingecko_id": "source-protocol",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/planq/images/srcx.png"
```
JeremyParish69, line 198: Please change the location to .../non-cosmos/binancesmartchain/images/srcx.png

_non-cosmos/binancesmartchain/assetlist.json
```diff
@@ -174,6 +174,34 @@
...
+        {
+          "denom": "srcx",
+          "exponent": 9
+        }
+      ],
+      "type_asset": "erc20",
+      "address": "0x454b90716a9435e7161a9aea5cf00e0acbe565ae",
+      "base": "0x454b90716a9435e7161a9aea5cf00e0acbe565ae",
+      "name": "Source Token",
+      "display": "SRCX",
```
JeremyParish69, line 194: please lowercase the display denom so that it's pointing to a defined denom_unit

JeremyParish69:
Can you please remove Source Testnet's coinGekco Id? It never should have been added there, and is the wrong one anyway,  and now it would clash with this. https://github.com/cosmos/chain-registry/blob/edb94fe36918ad46afc693300419ea101e12739e/testnets/sourcetestnet/assetlist.json#L25

ivanshukhov:
Well, at least for the next bridged asset I know what to do now.

JeremyParish69:
Thanks for your cooperation. Everything looks good now

PR #3911, opened 2024-02-19
Upgrade Migaloo to v4.1.3

reversesigh:
https://ping.pub/Migaloo/gov/24

https://ping.pub/Migaloo/block/5689150


reversesigh:
Team is requesting that validators vote no on the prop and wait for another one - will edit this based on the correct new prop / height when released.

PR #3915, opened 2024-02-19
Add fee suggestions for Pylons.

MikeSofaer:
Need fee suggestions to pass Osmosis checks.

### DIFF
--- a/pylons/chain.json
+++ b/pylons/chain.json
@@ -17,7 +17,10 @@
     "fee_tokens": [
       {
         "denom": "ubedrock",
-        "fixed_min_gas_price": 0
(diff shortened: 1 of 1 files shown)


pylons/chain.json
```diff
@@ -17,7 +17,10 @@
     "fee_tokens": [
       {
         "denom": "ubedrock",
-        "fixed_min_gas_price": 0
+        "fixed_min_gas_price": 0,
+        "low_gas_price": 0
```
JeremyParish69, line 21: missing a comma

MikeSofaer: ty, added it in.

PR #3921, opened 2024-02-20
update odin-osmosis client_id

danbryan:
I believe this is the correct client.  

cc @Cordtus 

```
osmosisd q ibc connection connections --limit 50000 | jq '.connections[] | select(.id == "connection-1551") | {client_id, counterparty}'
{
  "client_id": "07-tendermint-2007",
  "counterparty": {
    "client_id": "07-tendermint-10",
    "connection_id": "connection-9",
    "prefix": {
      "key_prefix": "aWJj"
    }
  }
}
```
Also, odin-protocol requested a 2nd channel be created for use by doki.

image

### DIFF
--- a/_IBC/odin-osmosis.json
+++ b/_IBC/odin-osmosis.json
@@ -7,7 +7,7 @@
   },
   "chain_2": {
     "chain_name": "osmosis",
-    "client_id": "07-tendermint-3088",
+    "client_id": "07-tendermint-2007",
     "connection_id": "connection-1551"
   },
   "channels": [
--- a/_IBC/odin-osmosis.json
+++ b/_IBC/odin-osmosis.json
@@ -27,6 +27,23 @@
         "preferred": true,
         "dex": "osmosis"
       }
+    },
+    {
+      "chain_1": {
+        "channel_id": "channel-74",
+        "port_id": "transfer"
+      },
+      "chain_2": {
+        "channel_id": "channel-20925",
+        "port_id": "transfer"
+      },
+      "ordering": "unordered",
+      "version": "ics20-1",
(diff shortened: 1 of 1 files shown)


Cordtus:
@danbryan 
Hello, that is the old client which has been [substituted in proposal 727](https://gist.github.com/Cordtus/72c94dd263983eda718049f32bec9ebb). 

I'm not entirely sure how it's possible that this client and connection are still associated with that channel on Osmosis. 
The counterparty client for this channel pair was also  replaced, but this one somehow remains.
The parameters you suggest are currently "working", however there is no valid counterparty configuration and it is only possible to make transfers in one direction -  `osmosis-1` >> `odin-mainnet-freya`.


additional information attached

Matching channels:

odin_channel-3.txt

osmo_channel-258.txt


Matching connections:

osmo_connection-2573.txt

odin_connection-90.txt


Matching clients:

odin_client-113.txt

osmo_client-3088.txt




Odd connection that should no longer be active:

odin_connection-9.txt

JeremyParish69:
New channels are pretty much only added for a wasm port. These ports are 'transfer' on both ends, which creates ambiguity about which would be used. I don't think we can add this.

danbryan:
cc @AddySiz

AddySiz:
> cc @JeremyParish69 removed the 2nd channel for the time being until justified by odin lead @AddySiz 

Yes @danbryan I had a chat with @JeremyParish69 about it. It's good to remove it for the time being.

JeremyParish69:
Shall we proceed with the client update?

AddySiz:
> Shall we proceed with the client update?

Yes please

PR #3937, opened 2024-02-21
chore to update nibiru url

cgilbe27:


cgilbe27:
@k-yang @CalicoNino

reversesigh:
I just checked and the github link still works for the genesis. Is there a reason to change it?

reversesigh:
I've reached out to Nibiru to see which they'd prefer. In the meantime can you please return the formatting of key_algos and compatible_versions? Once I hear back from them on the genesis URL I'll go ahead and merge.

k-yang:
Let's keep the GitHub URL. It'll have the same content as the networks.nibiru.fi URL, but GitHub has better availability than a web server that we spun up.

PR #3945, opened 2024-02-22
Add Titan Testnet

InnerPeace080:


testnets/titantestnet/chain.json
```diff
@@ -0,0 +1,163 @@
...
+    "staking_tokens": [
+      {
+        "denom": "atkx"
+      }
+    ]
+  },
+  "codebase": {
+    "git_repo": "https://github.com/tokenize-titan/titan",
+    "recommended_version": "v1.0.0",
+    "compatible_versions": [
```
JeremyParish69, line 40: Can you please add these version details (the full list of compatible versions, cosmos sdk version, consensus, ibc go version, cosmwasm , etc...) into the object inside the versions array as well?

InnerPeace080: Thank you. I just correct version information in `versions` array.

PR #3949, opened 2024-02-22
orai -- secret IBC channel

PFC-developer:
add's new channel between secret & Oraichain


_IBC/oraichain-secretnetwork.json
```diff
@@ -0,0 +1,32 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "Oraichain",
```
JeremyParish69, line 4: must be lowercase 'o'

PR #3993, opened 2024-02-27
Rename doki_Logo.png to doki.png

AddySiz:


JeremyParish69:
This changes the file name, but not the reference(s) to it--which should be updated to be the same.

JeremyParish69:
I believe this has since been updated so this PR can likely be closed.

JeremyParish69:
PR seems abandoned

PR #3999, opened 2024-02-27
Added Injective-odin IBC route

AddySiz:


JeremyParish69:
Please correct the schema

PR #4002, opened 2024-02-27
Add SRCX to Osmosis assetlist

Source-Protocol-Cosmos:
Add IBC SRCX from Source Chain.


JeremyParish69:
Again some detail are wrong (like the path in traces), but you shouldn't need to worry about updating this assetlist since it's won't affect what Osmosis shows. We will be doing regular semi-automated updates that will bring all listed assets back to this assetlist.

Source-Protocol-Cosmos:
Ok sounds good. We can close this PR.

PR #4012, opened 2024-02-29
Add ibc assets `tkx` and `ausdc` in osmosis testnet

InnerPeace080:


testnets/osmosistestnet/assetlist.json
```diff
@@ -94,6 +94,115 @@
...
+        },
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "axelartestnet",
+            "base_denom": "uausdc",
+            "channel_id": "channel-521"
+          },
+          "chain": {
+            "channel_id": "channel-4118",
```
JeremyParish69, line 189: This connection between axelartestnet and osmosistestnet is not registered. Would need to register this to replace the old (currently registered) connection

InnerPeace080: Thank u. I have changed back to use old connection.

PR #4013, opened 2024-02-29
Add TAKUMI

takumiasano:


juno/assetlist.json
```diff
@@ -2015,6 +2015,33 @@
...
+          "exponent": 0
+        },
+        {
+          "denom": "takumi",
+          "exponent": 6
+        }
+      ],
+      "base": "cw20:juno1gwstg5ycwzxkecqat46kw3s0hwt6c9afsuvllu7cec0xh7nquhqszgf82c",
+      "name": "TAKUMI ASANO",
+      "display": "takumi asano",
```
JeremyParish69, line 2035: This value is invalid is not defined as a denom_unit. I think you mean to have just 'takumi'

PR #4025, opened 2024-03-03
BSKT non-cosmos Token Addition

crypt0xg:
PR for adding BSKT to token registry.


_non-cosmos/solana/assetlist.json
```diff
@@ -163,6 +163,34 @@
...
+          "exponent": 0
+        },
+        {
+          "denom": "bskt",
+          "exponent": 5
+        }
+      ],
+      "base": "6gnCPhXtLnUD76HjQuSYPENLSZdG8RvDB1pTLM5aLSJA",
+      "name": "Basket",
+      "display": "BSKT",
```
JeremyParish69, line 183: display denom must reference a denom_unit. make this lowecase to make it true.

_non-cosmos/solana/assetlist.json
```diff
@@ -163,6 +163,34 @@
...
+      "symbol": "BSKT",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/solana/images/bskt.png"
```
JeremyParish69, line 186: Are we able to find an image (or compress this) to be less than 250KB?

PR #4046, opened 2024-03-06
Upgrade Source v3.0.1

Source-Protocol-Cosmos:
Small Security Update. Organized with Validators via Discord. All Validators currently running v3.0.1.

### DIFF
--- a/source/chain.json
+++ b/source/chain.json
@@ -50,9 +50,9 @@
     "versions": [
       {
         "name": "v3.0.0",
-        "recommended_version": "v3.0.0",
+        "recommended_version": "v3.0.1",
         "compatible_versions": [
-          "v3.0.0"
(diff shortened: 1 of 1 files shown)


reversesigh:
Is this a prop based upgrade or a patch for the current version? I couldn't find a prop for this on ping pub.

Source-Protocol-Cosmos:
Patch to the current version. No Gov Prop.

Source-Protocol-Cosmos:
Updated

source/chain.json
reversesigh, line 1: The changes to v3.0.0 are correct, but you still need to remove lines 65-78 for v3.0.1 then we can merge.

Source-Protocol-Cosmos: Done

PR #4049, opened 2024-03-06
add Wind token

poonc:
This PR adds the Wind Token tokenfactory to the repository


juno/assetlist.json
```diff
@@ -1854,6 +1854,31 @@
         }
       ]
     },
+    {
+      "denom_units": [
```
JeremyParish69, line 1858: Please move to the bottom of the list of assets

PR #4050, opened 2024-03-07
add Babylon testnet 3 

SebastianElvis:
This PR adds Babylon testnet 3 metadata.


reversesigh:
You have a few options here:

Close this PR and simply update the correct information in the current babylontest folder. As long as the original testnet is dead, this is acceptable until you reach mainnet.

Or assuming the original testnet is dead and you'd like to preserve the history, change the status of the original babylontestnet to `killed` and change its folder name to `babylontestnet1` and rename this folder `babylontestnet`

SebastianElvis:
> You have a few options here:
> 
> Close this PR and simply update the correct information in the current babylontest folder in a new PR. As long as the original testnet is dead, this is acceptable until you reach mainnet.
> 
> Or assuming the original testnet is dead and you'd like to preserve the history, change the status of the original babylontestnet to `killed` and change its folder name to `babylontestnet1` and rename the folder for the new chain to `babylontestnet`

Hi, thanks for the feedback. I have chosen to go for the second option, where I changed the status of the original testnet to `killed` and renamed it to `babylontestnet1`. I kept the new testnet name as `babylontestnet3` to reflect the testnet version number (have seen some other projects doing this e.g., `celestiatestnet3`). Lmk if this sounds good to you.

reversesigh:
@SebastianElvis Thank you!!

I'll refer to @JeremyParish69 on keeping the new testnet named `babylontestnet3`, but I do think we are trying to establish a standard and some of those examples might be before that.

A few more things and I think it should be good to go- On the new chain chain.json file, after `"chain_id": "bbn-test3"` add the line `"pre_fork_chain_name": "babylontestnet1"`. Also please remove the images folder from the previous testnet as it should only exist for the current chain.

JeremyParish69:
I think the plan was to keep the unnumbered chain (so babylontestnet) as the current chain, and all the images would just stay in there. As there are forks, a new chain directory (e.g., babylongtestnet1) can be created with an incrementing number (referring to the old fork), but the images I think should remain in the original directory which always remains as the current testnet.

SebastianElvis:
Thanks @reversesigh @JeremyParish69 for the clarifications, which totally make sense to me.

I have revised the PR to 1) rename the current testnet folder as `babylontestnet`, 2) add the `pre_fork_chain_name` entry, 3) remove the duplicated logo files, and 4) change the logo URLs accordingly.

Please feel free to take a look again

testnets/babylontestnet/chain.json
```diff
@@ -1,17 +1,16 @@
 {
   "$schema": "../../chain.schema.json",
-  "chain_name": "babylontestnet",
+  "chain_name": "babylontestnet3",
```
JeremyParish69, line 3: This value should also be 'bablyontestnet' (without the 3)

testnets/babylontestnet1/assetlist.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "../../assetlist.schema.json",
+  "chain_name": "babylontestnet",
```
JeremyParish69, line 3: this value should be 'babylontestnet1' to match the directory name

testnets/babylontestnet1/chain.json
```diff
@@ -0,0 +1,118 @@
+{
+  "$schema": "../../chain.schema.json",
+  "chain_name": "babylontestnet",
```
JeremyParish69, line 3: this value should be 'babylontestnet1' to match the directory name

JeremyParish69:
Looks good to me and passes checks. @reversesigh looks good to you?

PR #4061, opened 2024-03-08
Update cosmwasm_enabled for Oraichain

tubackkhoa:


oraibridge/chain.json
```diff
@@ -0,0 +1,107 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "oraibridge",
+  "status": "live",
+  "network_type": "mainnet",
+  "website": "https://orai.io/",
+  "pretty_name": "OraiBridge",
+  "chain_id": "oraibridge-subnet-2",
+  "bech32_prefix": "oraib",
```
JeremyParish69, line 9: Please register this prefix

oraibtcbridge/assetlist.json
```diff
@@ -0,0 +1,62 @@
...
+      ],
+      "base": "uoraibtc",
+      "name": "oraibtcbridge",
+      "display": "ORAIBTC",
+      "symbol": "ORAIBTC",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/oraichain/images/orai-white.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/oraichain/images/orai-white.svg"
+      },
+      "coingecko_id": "oraibtc-token",
```
JeremyParish69, line 25: This coingecko id doesn't exist. Any coingecko id that doesn't exist must be removed (even if it will exist soon).

oraibtcbridge/assetlist.json
```diff
@@ -0,0 +1,62 @@
...
+          "denom": "ORAIBTC",
+          "exponent": 6
+        }
+      ],
+      "base": "uoraibtc",
+      "name": "oraibtcbridge",
+      "display": "ORAIBTC",
+      "symbol": "ORAIBTC",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/oraichain/images/orai-white.png",
```
JeremyParish69, line 22: It should be using the same logo as ORAI token?
In which case, it might make sense to use image sync, so that any updates to the image will update everywhere without anything (like this reference) breaking.

oraibtcbridge/assetlist.json
```diff
@@ -0,0 +1,62 @@
...
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/oraichain/images/orai-white.svg"
+      },
+      "coingecko_id": "oraibtc-token",
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/oraichain/images/orai-white.png",
```
JeremyParish69, line 28: I want to make sure that this should be using the same token logo?
If so, it might be best to use an image_sync so that logos will be kept the same

oraibtcbridge/chain.json
```diff
@@ -0,0 +1,106 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "oraibtcbridge",
+  "status": "live",
+  "network_type": "mainnet",
+  "website": "https://orai.io/",
+  "pretty_name": "Orai Bitcoin Bridge",
+  "chain_id": "oraibtc-subnet-1",
+  "bech32_prefix": "oraibtc",
```
JeremyParish69, line 9: This must be registered as well

oraibtcbridge/chain.json
```diff
@@ -0,0 +1,106 @@
...
+  "daemon_name": "nomic",
```
JeremyParish69, line 10: Pretty sure this daemon name (and probably the daemon name in the other file are left over value and are incorrect). I'm guessing you copied Gravity Bridges' and Nomic's chain.json files and updated value here and there.

oraichain/assetlist.json
```diff
@@ -29,6 +29,406 @@
...
+        {
+          "denom": "ibc/A2E2EEC9057A4A1C2C0A6A4C78B0239118DF5F278830F50B4A6BDD7A66506B78",
+          "exponent": 0
+        },
+        {
+          "denom": "ATOM",
+          "exponent": 6
+        }
+      ],
+      "base": "ibc/A2E2EEC9057A4A1C2C0A6A4C78B0239118DF5F278830F50B4A6BDD7A66506B78",
```
JeremyParish69, line 45: If there's a base denom with an ibc hash, then this was likely ibc transferred. Where did it come from? Please use full trace paths for all tokens.

oraichain/assetlist.json
```diff
@@ -29,6 +29,406 @@
...
+      "name": "Cosmos",
+      "display": "ATOM",
+      "symbol": "ATOM",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/cosmoshub/images/atom.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/cosmoshub/images/atom.svg"
+      },
+      "coingecko_id": "cosmos",
```
JeremyParish69, line 53: The 'cosmos' coingecko id is already being used by Cosmos Hub ATOM, and cannot be reused. Please remove the coingecko ID for anything that doesn't describe an asset native to these chains.

oraibridge/chain.json
```diff
@@ -0,0 +1,107 @@
...
+  "explorers": [
+    {
+      "kind": "oraiscan",
+      "url": "https://scan.bridge.orai.io",
+      "tx_page": "https://scan.bridge.orai.io/txs/${txHash}"
+    }
+  ],
+  "images": [
+    {
+      "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/oraichain/images/orai.png",
```
JeremyParish69, line 103: Since we're looking to the /oraichain/images/ directory for the orai logo, I don't see a need to copy them to /oraibridge/images/ directory. Can we remove those please?

oraibridge/chain.json
```diff
@@ -0,0 +1,107 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "oraibridge",
+  "status": "live",
+  "network_type": "mainnet",
+  "website": "https://orai.io/",
+  "pretty_name": "OraiBridge",
+  "chain_id": "oraibridge-subnet-2",
+  "bech32_prefix": "oraib",
+  "daemon_name": "gravity",
```
JeremyParish69, line 10: May be legitimate, but also may be leftover from working on a copy of Gravity Bridge's chain.json file?

oraibridge/assetlist.json
```diff
@@ -0,0 +1,274 @@
...
+      },
+      "coingecko_id": "tether",
+      "images": [
+        {
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/oraichain/images/tether.svg"
+        }
+      ]
+    },
+    {
+      "description": "wTRX bridge tron",
```
JeremyParish69, line 196: For all these tokens, please define full trace paths, remove the coingecko id, and point to images added under its original chain. Some of these tokens will have to be registered under /_non-cosmos/... , e.g., TRON I think needs to be registered, but something like USDT is already registered.

JeremyParish69:
Can you please update the PR title to better describe the changes?

PR #4063, opened 2024-03-08
ADD TORO

vegerardoz:


osmosis/assetlist.json
```diff
@@ -14147,6 +14147,33 @@
...
+        {
+          "denom": "toro",
+          "exponent": 6
+        }
+      ],
+      "type_asset": "sdk.coin",
+      "address": "osmo1nr8zfakf6jauye3uqa9lrmr5xumee5n42lv92z",
+      "base": "factory/osmo1nr8zfakf6jauye3uqa9lrmr5xumee5n42lv92z/toro",
+      "name": "TORO",
+      "display": "TORO",
```
JeremyParish69, line 14167: this value is invalid. It must point to a denom_unit, which TORO is not defined. (You could use 'toro')

PR #4075, opened 2024-03-09
Update furya-1 chain.json

qf3l3k:
Updated repo and version based on current details from running chain.

### DIFF
--- a/furya/chain.json
+++ b/furya/chain.json
@@ -28,20 +28,20 @@
     ]
   },
   "codebase": {
-    "git_repo": "https://github.com/furysport/furya-chain",
-    "recommended_version": "v1.0.0",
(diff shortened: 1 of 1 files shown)


furya/chain.json
```diff
@@ -28,20 +28,20 @@
...
+      "v3.1.0"
     ],
     "genesis": {
       "genesis_url": "https://github.com/furysport/furya-chain/blob/main/network/mainnet/genesis.json"
     },
     "versions": [
       {
-        "name": "v1.0.0",
-        "recommended_version": "v1.0.0",
+        "name": "v3.1.0",
```
JeremyParish69, line 41: Instead of replacing v1 with v3 inside the `versions` array, please add v3 as a new object below v1. This allow us to build up a version history.

PR #4076, opened 2024-03-09
Update chain.json

Panvasi:


JeremyParish69:
This PR appears to remove and add unrelated endpoints due to having branched from an outdated state.
Also, please use a more descriptive PR title like "add {provider} endpoints to {chain}".

PR #4079, opened 2024-03-10
add lava mainnet

omerlavanet:


lava/assetlist.json
```diff
@@ -0,0 +1,46 @@
...
+        {
+          "denom": "lava",
+          "exponent": 6
+        }
+      ],
+      "type_asset": "sdk.coin",
+      "base": "ulava",
+      "name": "Lava",
+      "display": "lava",
+      "symbol": "Lava",
```
JeremyParish69, line 21: Please use all caps for ticker symbols

lava/chain.json
```diff
@@ -0,0 +1,119 @@
...
+        "consensus": {
+            "type": "cometbft",
+            "version": "0.37.4"
+        },
+        "cosmwasm_enabled": false,
+        "ibc_go_version": "v7.2.0",
+        "ics_enabled": [
+            "ics20-1"
+        ],
+        "versions": [
```
JeremyParish69, line 58: fyi, when using a versions.json file, the versions array is not required. (In fact, I recommend against it because it means two places to update and can lead to de-sync). But it's not a strict rule at this point.

lava/assetlist.json
omerlavanet: done

PR #4084, opened 2024-03-11
Add Osmosis v23

nullmames:
Have only added as current version


JeremyParish69:
When updating the version, the versions array must also be updated.

PR #4085, opened 2024-03-11
Update Humans'ai and HEART Images

ToDy95:
removed other asset images as they caused mixed identity on certain websites


JeremyParish69:
If removing images, please update all un-synchronised references to it (as there are in the chain.json file)

ToDy95:
i updated the photos :D sorry about that

PR #4087, opened 2024-03-11
Neutron v3 for testnet

pr0n00gler:


JeremyParish69:
If this is a current version, please also update the section above the versions array with similar info

pr0n00gler:
> If this is a current version, please also update the section above the versions array with similar info

updated

PR #4090, opened 2024-03-11
Add Stride's stDYM

shellvish:
Adds info for stDYM


stride/assetlist.json
```diff
@@ -366,6 +366,43 @@
...
     },
+    {
+      "description": "Stride's liquid staked DYM",
+      "denom_units": [
+        {
+          "denom": "stadym",
+          "exponent": 0
+        },
+        {
+          "denom": "stTIA",
```
JeremyParish69, line 377: this looks like a typo

stride/assetlist.json
```diff
@@ -366,6 +366,43 @@
...
+      ],
+      "base": "stadym",
+      "name": "Stride Staked DYM",
+      "display": "stDYM",
+      "symbol": "stDYM",
+      "traces": [
+        {
+          "type": "liquid-stake",
+          "counterparty": {
+            "chain_name": "dymension_1100-1",
```
JeremyParish69, line 389: although this is the chain_id, we need chain_name which is 'dymension'

shellvish: Makes sense, updated!

stride/assetlist.json
shellvish: Agreed, good catch! Updated

PR #4100, opened 2024-03-11
Updated genesisL1 chain configs

zenodeapp:
Hi!

GenesisL1 has changed since we've forked to a different codebase. This PR aims to update its configurations.

### DIFF
--- a/genesisl1/chain.json
+++ b/genesisl1/chain.json
@@ -37,11 +37,27 @@
     },
     "versions": [
       {
+        "name": "v0.3.0",
+        "tag": "v0.3.0",
+        "height": 0,
+        "recommended_version": "v0.5.0",
+        "compatible_versions": [
+          "v0.3.0",
+          "v0.5.0"
+        ],
+        "cosmos_sdk_version": "0.44.5",
+        "next_version_name": "v1.0.0"
+      },
+      {
         "name": "v1.0.0",
+        "tag": "v1.0.0",
+        "height": 7400000,
+        "proposal": 92,
         "recommended_version": "v1.0.0",
         "compatible_versions": [
           "v1.0.0"
-        ]
+        ],
+        "cosmos_sdk_version": "0.46.15"
       }
     ]
   },
--- a/genesisl1/chain.json
+++ b/genesisl1/chain.json
@@ -7,7 +7,7 @@
   "chain_id": "genesis_29-2",
   "bech32_prefix": "genesis",
   "daemon_name": "genesisd",
-  "node_home": "$HOME/.genesisd",
+  "node_home": "$HOME/.genesis",
   "slip44": 118,
   "fees": {
     "fee_tokens": [
@@ -27,20 +27,20 @@
     ]
   },
   "codebase": {
-    "git_repo": "https://github.com/alpha-omega-labs/genesisd",
-    "recommended_version": "v0.3.0",
+    "git_repo": "https://github.com/alpha-omega-labs/genesis-crypto",
+    "recommended_version": "v1.0.0",
     "compatible_versions": [
-      "v0.3.0"
+      "v1.0.0"
     ],
     "genesis": {
-      "genesis_url": "https://github.com/alpha-omega-labs/genesisd/raw/neolithic/genesis_29-1-state/genesis.json"
+      "genesis_url": "https://raw.githubusercontent.com/alpha-omega-labs/genesis-parameters/main/genesis_29-2/genesis.json"
     },
     "versions": [
       {
-        "name": "v0.3.0",
-        "recommended_version": "v0.3.0",
+        "name": "v1.0.0",
+        "recommended_version": "v1.0.0",
         "compatible_versions": [
-          "v0.3.0"
+          "v1.0.0"
         ]
       }
     ]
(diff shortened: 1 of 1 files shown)


reversesigh:
Since this is a new chain id and new genesis, you'll need to rename the old folder to `genesisl11` and change its status to `killed` and change the chain name to `genesisl11` in chain.json (also remove images from the renamed old chain). 

Create a new folder named `genesisl1` for the new chain, and add a field for `"pre_fork_chain_name": "genesisl11"` after the chain_id field.

zenodeapp:
> Since this is a new chain id and new genesis, you'll need to rename the old folder to `genesisl11` and change its status to `killed` and change the chain name to `genesisl11` in chain.json (also remove images from the renamed old chain).
> 
> Create a new folder named `genesisl1` for the new chain, and add a field for `"pre_fork_chain_name": "genesisl11"` after the chain_id field.

Heya :)! Familiar display picture hehe.

It's not a new chain-id, we've hard-forked to a different code-base but kept everything intact (luckily).

zenodeapp:
I see that the genesis file might have caused this assumption, but as you can see I haven't touched the chainId field itself.

zenodeapp:
> > I see that the genesis file might have caused this assumption, but as you can see I haven't touched the chainId field itself. Also that genesis file is still the same, just placed somewhere else.
> > **In short:** The config before my PR was already based on genesis_29-2.
> 
> Ah I see - so at some point there was a change to the chain_id but it was never reflected here before when that event happened. Is v0.3.0 part of `genesis_29-2` then or `genesis_29-1`?

v0.3.0 was part of genesis_29-2, yes.

zenodeapp:
Hmm, maybe for both chain IDs though.

Let me check back the commit history.

zenodeapp:
Okay @reversesigh, I checked back and it has always been known as `genesis_29-2` in the chain-registry.

reversesigh:
Got it!! 

Ok, so if v0.3.0 is the first version for `genesis_29-2` but it's patched to v1.0.0 without a proposal, then your changes should be ok I think? Can you confirm that there was no prop for v1.0.0 and that it's acting more of a patch for the current version? If so we'll go ahead and merge, otherwise we'll need to create v1.0.0 as a new version in the version history (but only if it's a prop upgrade).

zenodeapp:
> Got it!!
> 
> Ok, so if v0.3.0 is the first version for `genesis_29-2` but it's patched to v1.0.0 without a proposal, then your changes should be ok I think? Can you confirm that there was no prop for v1.0.0 and that it's acting more of a patch for the current version? If so we'll go ahead and merge, otherwise we'll need to create v1.0.0 as a new version in the version history (but only if it's a prop upgrade).

Okay so I was in charge of this upgrade. We did have a proposal to do this. Our versioning system was never really good (we didn't have any). But in this revamp we now do maintain this.

For those who wish to full sync their nodes they have to first use v0.5.0 of the https://github.com/alpha-omega-labs/genesis-ethermint repo, which is the old codebase. Then once the proposal upgrade height gets triggered they should hop onto V1.0.0 in the https://github.com/alpha-omega-labs/genesis-crypto repository. But those who wish to statesync could immediately use the newer repo. I've documented this all in the readme's of both repos. 

So the chain ID didn't change but we did hop onto a different repo as to mark this new beginning.

I hope this makes a bit sense!

zenodeapp:
Extra info: v0.5.0 was basically 0.3.0, the only difference were useful scripts added and a repo cleanup.

reversesigh:
No this is fine. What should be done then is this:

Keeping your changes to codebase as v1.0.0 is fine! Leave this.

On the version history, keep v0.3.0 as it was before your merge, but change to v0.5.0 as the recommended version, and `"v0.3.0, v0.5.0" as compatible versions. Add a `"next_version_name": v1.0.0"` field, then create a new set for `v1.0.0` afterward with the correct information for that version (include new fields for `"proposal":` and `"height":`). 

I think this is the best approach for now. Let's take a look after these changes are made and we should be able to merge then.

As for the repo conflict between v0.5.0 and v1.0.0 I think that is ok for now. I can talk with @JeremyParish69 about ways to approach this since its a little unorthodox. It might be possible to simply add a one-time "git_repo" field in the version history for v0.3.0 to signify the different codebase (but I wouldn't make a change like this yet).

JeremyParish69:
Probably OK to just use the latest codebase as git_repo value for now.

zenodeapp:
Hey @reversesigh and @JeremyParish69; I changed it!

Also, I let the version name reflect the plan name in the proposal itself. And gave the first version (0.3.0) a similar name to keep it concise. This is the intended naming convention I believe? This should also hint towards the change in repo.

reversesigh:
@JeremyParish69 if you're ok with the added "name" to genesis then I think this is good to go.

zenodeapp:
> @JeremyParish69 if you're ok with the added "name" to genesis then I think this is good to go.

Saw it in the readme, though tell me if this needs to go, then I'll remove it asap.

reversesigh:
> > @JeremyParish69 if you're ok with the added "name" to genesis then I think this is good to go.
> 
> Saw it in the readme, though tell me if this needs to go, then I'll remove it asap.

We'll see what he says. I'm still learning all the nuances of small changes like that so I didn't watch to merge yet just in case. But everything else looks good to me!

JeremyParish69 (approved):

JeremyParish69:
@reversesigh has definitely put more care into reviewing this, but it looks OK to me!

zenodeapp:
> @reversesigh has definitely put more care into reviewing this, but it looks OK to me! 

Love the extra care ♥️!

PR #4103, opened 2024-03-12
add genesis nibiru IBC channels

ertemann:


_IBC/nibiru-osmosis.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "chain_2": {
+    "chain_name": "osmosis",
+    "client_id": "07-tendermint-3104",
+    "connection_id": "connection-2580"
+  },
+  "channels": [
+    {
+      "chain_1": {
+        "channel_id": "channel-139",
```
JeremyParish69, line 16: hey @ertemann this looks like a copypastypo, because channel-139 is registered as the chain1 channel for the axelar-nibiru connection, while for Osmosis, the Nibiru-side channel appears to be channel-0, not -139.

JeremyParish69: https://github.com/cosmos/chain-registry/pull/4210

PR #4105, opened 2024-03-12
Updated Asset logos

AddySiz:
Updated Doki and MYRK logos to new one


JeremyParish69:
Can we find an image file for MYRK that has exactly square aspect ratio? It's showing 250 by 254

AddySiz:
> Can we find an image file for MYRK that has exactly square aspect ratio? It's showing 250 by 254

done added 250x250

PR #4114, opened 2024-03-13
Elys: Adds USDC and ATOM fee currencies

Wgil:
It update Elys chain definition to support different fee currencies in wallet providers.

### DIFF
--- a/testnets/elystestnet/assetlist.json
+++ b/testnets/elystestnet/assetlist.json
@@ -8,16 +8,12 @@
         {
           "denom": "uelys",
           "exponent": 0,
-          "aliases": [
-            "microelys"
-          ]
+          "aliases": ["microelys"]
         },
         {
           "denom": "melys",
           "exponent": 3,
-          "aliases": [
-            "millielys"
-          ]
+          "aliases": ["millielys"]
         },
         {
           "denom": "elys",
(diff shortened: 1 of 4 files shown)


testnets/elystestnet/assetlist.json
```diff
@@ -74,6 +66,97 @@
...
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "noble",
+            "base_denom": "uusdc",
+            "channel_id": "channel-19"
+          },
+          "chain": {
+            "channel_id": "channel-12",
+            "path": "transfer/channel-12/uusdc"
```
JeremyParish69, line 97: Neither of the IBC connections are registered. I assume you meant to actually put nolustestnet and cosmoshubtestnet ?

Wgil: Just updated the names. Is it ok now? These are Noble Testnet and Cosmos Testnet IBC connections w/ Elys

Wgil: Thanks for checking

JeremyParish69: Thanks for correcting the name. Although neither of these IBC connections are registered here: https://github.com/cosmos/chain-registry/tree/master/testnets/_IBC

Wgil: Updated. thanks

PR #4138, opened 2024-03-18
fix: remove ibc connection coreum with sei

akhlopiachyi:
This channels have not been ever used


JeremyParish69:
It there a site that shows that it's never been used?

keyleu:
Hey @JeremyParish69 

{
  "amount": {
    "denom": "ibc/81BD95B0890B8D0130755E8338CF4AA48C7CBBD149C1D66EC9F9B62AFAE5C4F3",
    "amount": "16928523"
  }
}

This is the total supply that was bridged (about 16 SEI in total (quite a small amount) the first days the channel was open (and that we bridged for testing) and before it expired because it wasn't maintained not used by anyone. The channel hasn't been maintained for months.

PR #4154, opened 2024-03-19
Add IBC info for Oraichain

ducphamle2:


_IBC/oraichain-neutaro.json
```diff
@@ -0,0 +1,32 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "oraichain",
+    "client_id": "07-tendermint-156",
+    "connection_id": "connection-120"
+  },
+  "chain_2": {
+    "chain_name": "neutaro",
```
JeremyParish69, line 9: this chain is not registered

_IBC/oraichain-cosmoshub.json
```diff
@@ -0,0 +1,32 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "oraichain",
```
JeremyParish69, line 4: for all of these, oraichain must be chain_2, since it comes later alphabetically. (same with file names)

PR #4158, opened 2024-03-20
Add an extra ApeironNodes Seed

Panvasi:


chain4energy/chain.json
```diff
@@ -177,6 +177,11 @@
   },
   "peers": {
     "seeds": [
+      {
```
JeremyParish69, line 180: Please add to the bottom of the array

Panvasi (commented):
Looking good👍

chain4energy/chain.json
JeremyParish69: This conversation was not resolved.

Panvasi: > This conversation was not resolved.

Now?

JeremyParish69: No, it's still not at the very bottom, but I see you've added it after an existing Seed on the same provider, and it's not worth additional back and forth.

PR #4163, opened 2024-03-20
akiro

akiroinu:


reversesigh:
Since the release isn't versioned and the source is closed, I suggest adding modules to both codebase and version history. This is what I got from the binary that's available Dropbox:

```
"cosmos_sdk_version": "v0.45.4",
"ibc_go_version": "v0.34.19",
        "consensus": {
          "type": "tendermint",
          "version": "v0.34.19"
        }
```

akiroinu:
The source code is here:
https://github.com/akiroinu/akiro/releases/tag/v1.0.0

How can I do the versioning process? Is there any documentation about this?

reversesigh:
> The source code is here: https://github.com/akiroinu/akiro/releases/tag/v1.0.0
> 
> How can I do the versioning process? Is there any documentation about this?

I believe the version is defined in the Makefile used to compile the source? You'll need to refer to documentation thats available online or speak with other devs. I think we should return to adding this after you have this sorted out.

reversesigh:
> I solved it with makefile.can you check again.thank you

Cool.

I was able to pull and build using the main repo and see it posts v0.1. The problem is your Makefile isn't part of v0.1, only the main branch. So either list the recommended_version and compatible_version as "main" in chain.json (for now) or create a new version tag (i.e. v0.2) that includes the makefile and posts v0.2 in the output and use that as your entries instead. Either solution should be ok for now.

akiroinu:
> > > > I solved it with makefile.can you check again.thank you
> > > 
> > > 
> > > Cool.
> > > I was able to pull and build using the main repo and see it posts v0.1. The problem is your Makefile isn't part of v0.1, only the main branch. So either list the recommended_version and compatible_version as "main" in chain.json (for now) , or add Makefile to v0.1 (if possible). If not possible to add to v0.1 maybe create a new version tag (i.e. v0.2) that includes the makefile and posts v0.2 in the output and use that as your entries instead. Any of these solutions should be ok.
> > > Also don't forget to add the module information I referenced in [#4163 (comment)](https://github.com/cosmos/chain-registry/pull/4163#issuecomment-2010779203). Refer to something like the Cosmos Hub chain.json as ref for formatting if you need help.
> > 
> > 
> > thank you
> 
> One more thing - are you actually planning to add your explorer to Ping.Pub? Or do you have an alternative lined up? We should probably leave the explorer fields blank until you're added to one.

https://github.com/ping-pub/ping.pub/pull/56

I added for ping. We will add other explorers later.

PR #4170, opened 2024-03-21
Update `ibc` axelar-bitsong

clemensgg:
Previous client on the axelar side has expired due to a relayer misconfiguration. The channel was only used for testing. New channels have been created as per request, the config error has been fixed on our side.


JeremyParish69:
Is there any interface that can show that this channel only ever transferred a small quantity of tokens?

PR #4176, opened 2024-03-22
update uptick chain cosmoswasm Version

xianxianzhou:
update uptick chain cosmoswasm Version config

### DIFF
--- a/uptick/chain.json
+++ b/uptick/chain.json
@@ -42,8 +42,8 @@
     },
     "cosmos_sdk_version": "v0.47.5",
     "consensus": {
-      "type": "tendermint",
-      "version": "0.34"
+      "type": "cometbft",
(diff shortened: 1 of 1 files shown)


reversesigh:
Looking over the version history, instead of replacing v0.2.4 you need to add a `"next_version_name": ""` field and then add new versions between 0.2.4 and the current version according to prop history (also include prop number and height for each). According to https://uptick.explorers.guru/proposals that will be v0.2.8, v0.2.11, v0.2.17, v0.2.18, and v0.2.19 (and checking go.mod for each to include proper module info). Refer to the Cosmos Hub chain.json for a reference on how to handle version history.

JeremyParish69:
Should we close this or try to repair it?

reversesigh:
> Should we close this or try to repair it?

I went ahead and added a basic chain history sans modules so we can push this through.

PR #4184, opened 2024-03-22
Updated and Added BackBone Labs Liquid Staked Assets.

tvl83:
Updated logos and information for BackBone Labs LSTs across the cosmos.

### DIFF
--- a/chihuahua/assetlist.json
+++ b/chihuahua/assetlist.json
@@ -215,7 +215,7 @@
           "exponent": 0
         },
         {
-          "denom": "bhuahua",
+          "denom": "bHUAHUA",
           "exponent": 6
         }
       ],
@@ -232,7 +232,8 @@
       ],
       "logo_URIs": {
         "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/chihuahua/images/bhuahua.png"
-      }
+      },
+      "coingecko_id": "backbone-labs-staked-huahua"
     }
   ]
 }
--- a/terra2/assetlist.json
+++ b/terra2/assetlist.json
@@ -377,7 +377,7 @@
           "exponent": 0
         },
         {
-          "denom": "bluna",
+          "denom": "bLUNA",
           "exponent": 6
         }
       ],
@@ -392,7 +392,8 @@
         {
           "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/terra2/images/bLUNA.png"
         }
-      ]
+      ],
+      "coingecko_id": "backbone-labs-staked-luna"
     },
     {
       "description": "Sayve is a revolutionary language learning app in the Web3 era that combines gamification, blockchain technology, and a Metaverse experience to motivate users to learn languages while earning rewards.",
--- a/juno/assetlist.json
+++ b/juno/assetlist.json
@@ -2060,7 +2060,7 @@
           "exponent": 0
         },
         {
-          "denom": "bjuno",
+          "denom": "bJUNO",
           "exponent": 6
         }
       ],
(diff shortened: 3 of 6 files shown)


terra2/assetlist.json
```diff
@@ -382,15 +382,15 @@
         }
       ],
       "base": "cw20:terra17aj4ty4sz4yhgm08na8drc0v03v2jwr3waxcqrwhajj729zhl7zqnpc0ml",
-      "name": "boneLuna",
-      "display": "bluna",
+      "name": "BackBone Labs Liquid Staked LUNA",
+      "display": "bLUNA",
```
JeremyParish69, line 386: This case-sensitive display denom is not defined as "bLUNA" in denom_units.

chihuahua/assetlist.json
```diff
@@ -206,6 +206,33 @@
...
+        {
+          "denom": "bhuahua",
+          "exponent": 6
+        }
+      ],
+      "type_asset": "cw20",
+      "address": "chihuahua1jz5n4aynhpxx7clf2m8hrv9dp5nz83k67fgaxhy4p9dfwl6zssrq3ymr6w",
+      "base": "cw20:chihuahua1jz5n4aynhpxx7clf2m8hrv9dp5nz83k67fgaxhy4p9dfwl6zssrq3ymr6w",
+      "name": "BackBone Labs Liquid Staked HUAHUA",
+      "display": "bHUAHUA",
```
JeremyParish69, line 226: In all cases, there seems to be a misunderstanding of the "display" [denom] property. It is meant to reference one of the defined denom_units, and in all your additions, the value for display is not defined. I recommend making the display values fully lowercase (since that's how it's defined in denom_units)

tvl83: okay, ill fix these things.

JeremyParish69:
I noted an issue that needs correction, but otherwise this looks acceptable.

chihuahua/assetlist.json
tvl83:
```json
{
      "description": "BackBone Labs Liquid Staked WHALE",
      "denom_units": [
        {
          "denom": "factory/migaloo1mf6ptkssddfmxvhdx0ech0k03ktp6kf9yk59renau2gvht3nq2gqdhts4u/boneWhale",
          "exponent": 0
        },
        {
          "denom": "boneWHALE",
          "exponent": 6
        }
      ],
      "base": "factory/migaloo1mf6ptkssddfmxvhdx0ech0k03ktp6kf9yk59renau2gvht3nq2gqdhts4u/boneWhale",
      "name": "boneWHALE",
      "display": "boneWHALE",
      "symbol": "bWHALE",
      "logo_URIs": {
        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/migaloo/images/bWHALE.png"
      },
      "images": [
        {
          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/migaloo/images/bWHALE.png"
        }
      ]
    }
```

so this one is ok because the `denom_units[1].denom` matches `display`?  i was following the pattern of others which put the 2nd denom in the denom_units array as the last part of the contract address.

tvl83:
@JeremyParish69 I just noticed that there's another bJUNO token and it's conflicting on the mintscan website. for juno it shows 2 entries for bJUNO and has our logo but the contract address of the other bJUNO. 

I am curious if there is another solution besides going by a different name?

JeremyParish69:
IMO, it would be ideal for interfaces to show the names "BackBone Labs Liquid Staked JUNO" vs "StakeEasy bJUNO" to differentiate the assets, however there is no guarantee that users wouldn't confuse assets with a similar or identical ticker symbol.

The symbol of an asset is allowed to match another asset's symbol--the only property that guarantees disambiguation is the base denomination. Any improperly handled differentiation is the responsibility of which 3rd party app isn't displaying correctly.

Sounds like Mintscan aught to correct how they manage asset identifiers, and should be brought up with the Cosmostation team. I'm guessing that some automation they use scrapes the images from the chain registry, but then renames the image files copied to their own repo by symbol, causing any formerly scraped images to be overwritten by the last one.
image

PR #4189, opened 2024-03-23
added harpoon-4 <> pisco-1

danbryan:
Added new path for terra2-testnet to kujira-testnet


testnets/_IBC/kujiratestnet-terra2testnet.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "../../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "terra2testnet",
```
JeremyParish69, line 4: terra2testnet must be chain_2

danbryan: just swapped them, thanks for catching that!

PR #4192, opened 2024-03-23
Update pretty_name for Akiro

akiroinu:


JeremyParish69:
Change is OK, but please use appropriate PR titles, like "Update pretty_name for Akiro"

PR #4198, opened 2024-03-24
Delete akiro directory

akiroinu:
I think I need to improve myself. so i request my akiro inu project to be deleted Sincelery


reversesigh:
The network is still live, and there's another validator who has already joined and submitted their info here as well. If you're not planning on shutting down the network, I think its best to keep this listed.

akiroinu:
I stopped the network.

akiroinu:
I decided to continue running the network. I gave up on deleting the folder.

PR #4202, opened 2024-03-25
Add Neutaro to chain-registry 

Neutaro:


Neutaro/assetlist.json
```diff
@@ -0,0 +1,34 @@
+{
+  "$schema": "../assetlist.schema.json",
+  "chain_name": "Neutaro",
```
JeremyParish69, line 3: Please do not use uppercase letters for chain_name. assetlist.json::chain_name, chain.json::chain_name, and the directory name should all be `neutaro`.

Neutaro/assetlist.json
```diff
@@ -0,0 +1,34 @@
...
+        },
+        {
+          "denom": "neutaro",
+          "exponent": 6
+        }
+      ],
+      "base": "uneutaro",
+      "name": "neutaro",
+      "display": "neutaro",
+      "symbol": "$ntmpi",
```
JeremyParish69, line 20: Please use all caps for ticker symbols and exclude the `$`

Neutaro/assetlist.json
```diff
@@ -0,0 +1,34 @@
...
+          "denom": "uneutaro",
+          "exponent": 0
+        },
+        {
+          "denom": "neutaro",
+          "exponent": 6
+        }
+      ],
+      "base": "uneutaro",
+      "name": "neutaro",
```
JeremyParish69, line 18: This can capitalize the 'n' since it will be be displayed as this name on interfaces.

Neutaro/chain.json
```diff
@@ -0,0 +1,109 @@
+﻿{
+  "$schema": "../chain.schema.json",
+  "chain_name": "Neutaro",
+  "status": "live",
+  "network_type": "mainnet",
+  "website": "https://neutaro.com/",
+  "pretty_name": "Neutaro",
+  "chain_id": "Neutaro-1",
+  "bech32_prefix": "neutaro",
```
JeremyParish69, line 9: Please register this prefix: https://github.com/satoshilabs/slips/blob/master/slip-0173.md

Neutaro: Will work on that now.

Neutaro/assetlist.json
Neutaro: Done

Neutaro/assetlist.json
Neutaro: Done

Neutaro/assetlist.json
Neutaro: also done

JeremyParish69:
Wow I was stumped on that last failure. Not sure how you knew that character was there.

Neutaro:
Just luck, I've had something like that happen once and the - '' character error - looked familiar. But I can't tell you where it's from. I suspect from the browser editor GitHub has, but not sure.

PR #4220, opened 2024-03-26
docs: remove blockspace race testnet Celestia

jcstein:
This PR removes an old testnet that is no longer live for Celestia. cc @Bidon15


Bidon15 (approved):

reversesigh:
This is a good chance to actually fix up the testnet situation with celestia on here.

Instead of completely removing, I suggest:

The folder that is currently labeled `testnets/celestiatestnet3` needs to be renamed `testnets/celestianet`. Sidenote - I don't think Mocha was a fork of anything, but if it was then we'll add `  "pre_fork_chain_name": "celestiatestnet3"`. If it was a new launch with a new genesis and from block 1 then ignore adding that field.

I only joined Celestia during Mocha so i can't speak what was before, but if blockspace race was the testnet before mocha, then rename the blockspace race testnet folder to `celestiatestnet3` and change the status to `killed` and change the `chain_name` to `celestiatestnet3`.  Also delete the images folder for this.

JeremyParish69:
PR has been abandoned.

jcstein:
will open a new PR for this

PR #4228, opened 2024-03-27
Add BRNZ to osmosis assetlist

triccs:
Let me know if the SVG needs to be resized.

Also I set the exponents to 0, so I only added 1 entry to the denom_units list, hope this is fine.


triccs:
Note for the merger: 

I'm trying to make the smallest unit 'BRNZ' & not the full denom so if the denom_units allows 2 names for the smallest decimal it should work, if not plz don't merge, thank you in advance.

osmosis/assetlist.json
```diff
@@ -14261,6 +14261,26 @@
...
+          "denom": "brnz",
+          "exponent": 0
+        }
+      ],
+      "base": "factory/osmo13gu58hzw3e9aqpj25h67m7snwcjuccd7v4p55w/brnz",
+      "name": "BRNZ",
+      "display": "brnz",
+      "symbol": "BRNZ",
+      "logo_URIs": {
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/BRNZ.svg"
```
JeremyParish69, line 14282: Please provide an image with a square aspect ratio. (same width as height)

osmosis/assetlist.json
```diff
@@ -14261,6 +14261,26 @@
...
+    {
+      "description": "Memecoin for The International Brane Wave",
+      "denom_units": [
+        {
+          "denom": "factory/osmo13gu58hzw3e9aqpj25h67m7snwcjuccd7v4p55w/brnz",
+          "exponent": 0
+        },
+        {
+          "denom": "brnz",
+          "exponent": 0
```
JeremyParish69, line 14274: Please do not define two separate denom units with the same exponent value (in this case 0). Instead, please add 'brnz' as an alias of factory/osmo13gu58hzw3e9aqpj25h67m7snwcjuccd7v4p55w/brnz.

osmosis/assetlist.json
```diff
@@ -14261,6 +14261,26 @@
...
+        }
+      ],
+      "base": "factory/osmo13gu58hzw3e9aqpj25h67m7snwcjuccd7v4p55w/brnz",
+      "name": "BRNZ",
+      "display": "brnz",
```
JeremyParish69, line 14279: It is best to have display refer to the denom of a unit rather than an alias, so please change display value to factory/osmo13gu58hzw3e9aqpj25h67m7snwcjuccd7v4p55w/brnz.
Note that interface will only be displaying the name and symbol "BRNZ"--'display' is only used for informing which exponent to use.

osmosis/assetlist.json
```diff
@@ -14261,6 +14261,26 @@
...
+          "denom": "factory/osmo13gu58hzw3e9aqpj25h67m7snwcjuccd7v4p55w/brnz",
+          "exponent": 0
+        },
+        {
+          "denom": "brnz",
+          "exponent": 0
+        }
+      ],
+      "base": "factory/osmo13gu58hzw3e9aqpj25h67m7snwcjuccd7v4p55w/brnz",
+      "name": "BRNZ",
```
JeremyParish69, line 14278: It is recommended to have the name be as the token should be pronounced in conversation. Should users spell out BRNZ, or should they say 'Brainz' (which is just my guess of what this is supposed to be called)?

triccs: Pronounced brainz, spelled anyway they want

osmosis/assetlist.json
triccs: Added new image, more coin-like, should be 2222x2222

osmosis/assetlist.json
triccs: Goal is for the ticker to look like branez or burnz, just changed the name tho

PR #4230, opened 2024-03-28
takumi.png

NEXXGENAI:
takumi


JeremyParish69:
Not being used by any asset?
Also, it's problematic to use images of real people, especially faces, and should be avoided.

NEXXGENAI:
> Not being used by any asset? Also, it's problematic to use images of real people, especially faces, and should be avoided.

Hey Jeremy, It's part of a meme sticker set on Telegram.  It's not being used by any other asset currently.  Is the use of a face for a logo forbidden?  I chose this specifically because of the memes involved, so would like to run with it if possible.

Thanks,
Paul

JeremyParish69:
Image additions need to be in the same PR as updating the JSON file that refers to it.

PR #4240, opened 2024-03-29
Add cvn-osmosis IBC channel and asset

bitnet:


_IBC/cvn-osmosis.json
```diff
@@ -0,0 +1,32 @@
+{
+    "$schema": "../ibc_data.schema.json",
+    "chain_1": {
+      "chain_name": "conscious",
```
JeremyParish69, line 4: The file name "cvn-osmosis.json" should be renamed to use the chain_name, i.e., "conscious-osmosis.json"

PR #4241, opened 2024-03-29
Update IBC for cosmoshub-irisnet

oncloudit:
the channel-0 in IRIShub was created early when the chain supported IBC ics20, and be running by IRISnet officially. have to update the previous in the repo as the id of client and connection are different.

### DIFF
--- a/_IBC/cosmoshub-irisnet.json
+++ b/_IBC/cosmoshub-irisnet.json
@@ -6,7 +6,7 @@
     "connection_id": "connection-187"
   },
   "chain_2": {
-    "chain_name": "irishub",
+    "chain_name": "irisnet",
     "client_id": "07-tendermint-1",
     "connection_id": "connection-1"
   },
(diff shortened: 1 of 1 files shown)


JeremyParish69:
Please provide evidence of both chains voting to change the canoncial channel (or that the previous channel was never used and has no tokens stuck on the counterparty chain).

oncloudit:
thank for your reply and explanation, we should register the IBC channel more early.
if there're no other ways to pass the replacement above or modify the policy of the Cosmos Chain Registry, we'll prepare to proposal an text gov about updating the registered channel in this repo, hope them will be passed on both CosmoHub and IRISHub chain!

JeremyParish69:
Well, @reversesigh or @dylanschultzie might have opinions on updating the policy, but I'd say I can make major changes to policy like that unilaterally, and in the meantime I think we'll need a signaling proposal on both chains.

JeremyParish69:
Closing unless there is anything to add

PR #4242, opened 2024-03-29
Change client, connection, channel id

cifer-ai:


JeremyParish69:
For each ibc connection update, please provide evidence of both chains voting to change the canoncial channel (or that the previous channel was never used and has no tokens stuck on the counterparty chain).

cifer-ai:
As the previous config is the misconfiguration from our side, we misunderstood and create these IDs for local nodes so they are never really used.
Not sure how to check the previous channel status, if there is any way to check them please share with us.

What I can provide is relayer channels and some transaction details which appear on Mintscan.

---

**cifer-cosmoshub**
Previous channel : channel-0 for cifer, channel-0 for cosmoshub
New channel : channel-1 for cifer, channel-822 for cosmoshub

As we can see from Cosmos Mintscan relayer that cifer-1 chain only has channel-821 and channel-822 on Cosmos side. While there is no channel-0 on Cosmos.

Screenshot 2567-03-30 at 04 45 05
Screenshot 2567-03-30 at 04 44 41

You may see the IBC Channel Open Confirm transaction details here:
https://www.mintscan.io/osmosis/tx/3D9CA53982F13191E53DF6983400717738DA35327CB36AFBC76913708BA4A1D8?height=14562837

---

**cifer-osmosis**
Previous channel : channel-0 for cifer, channel-0 for osmosis
New channel : channel-2 for cifer, channel-35741 for osmosis

Somehow we cannot see cifer-1 chain from Osmosis Mintscan relayer yet, but as I search for channel-0 on Osmosis, this channel belongs to Cosmoshub.

Screenshot 2567-03-30 at 04 44 41

You may see the IBC Channel Open Confirm transaction details here:
https://www.mintscan.io/cosmos/tx/0E8E9942CEFCEDCCB8907D08FB871E6442C48365D9A5BE892022364EA763298C?height=19764822

JeremyParish69:
Thanks for providing context. Now that I think about how the registered channel being 'channel-0' on the Cosmos Hub and Osmosis sides is impossible, clearly the previous registration is invalid and never should have been registered in the first place. We can proceed.

cifer-ai:
Thank you very much. 🙏🏻

PR #4243, opened 2024-03-29
Add IBC (ImBeCiles) to Osmosis Assets

imbecile-dev:


osmosis/assetlist.json
```diff
@@ -14280,6 +14280,30 @@
...
+      },
+      "coingecko_id": "",
+      "keywords": [
+        "memecoin",
+      ],
+      "socials": {
+        "webiste": "https://www.ibcmeme.wtf",
+        "twitter": "https://twitter.com/IBCmemecoin"
+      }
+    },
```
JeremyParish69, line 14307: Please remove the extra comma

osmosis/assetlist.json
```diff
@@ -14280,6 +14280,30 @@
       "logo_URIs": {
           "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/BRNZ.svg"
       }
+    },
+    {
+      "description": "OnE mEmEcOiN tO cOnNeCt oL ImBeCiles - aNd in Da Cosmos BiNd DeM",
+      "denom_units": [
+        {
+          "denom": "factory/osmo1kqdw6pvn0xww6tyfv2sqvkkencdz0qw406x54r/IBC",
```
JeremyParish69, line 14288: The base denom unit must always be exponent 0, so this definition must be incorrect. Please define two denom units, one for the base (exponent 0), and one for desired display (likely exponent 6), and reference that display denom unit via the 'display' field.

imbecile-dev: Updated this like so: 
image

Does this new "display" denom need to be registered onchain in someway? As I cant find any metadata update function available in osmosis tokenfactory module

sunnya97: There is a metadata update msg in the osmosis tokenfactory module

https://github.com/osmosis-labs/osmosis/blob/main/proto/osmosis/tokenfactory/v1beta1/tx.proto#L119

PR #4260, opened 2024-03-31
add NODERS rpc,api,grpc,jsonrpc

aleksandrnoders:


lambda/chain.json
```diff
@@ -96,12 +104,20 @@
...
       }
     ],
     "evm-http-jsonrpc": [
       {
         "address": "https://jsonrpc-lambda.d-stake.xyz",
         "provider": "D-stake"
+      },
+      {
+        "address": "https://lambda-jsonrpc.noders.services",
+        "provider": [NODERS]TEAM"
```
reversesigh, line 120: Missing starting comma for provider name here

aleksandrnoders: Corrected

stratos/chain.json
```diff
@@ -112,6 +120,9 @@
       {
         "address": "stratos.grpc.nodersteam.com:9090",
         "provider": "[NODERS]TEAM"
+      {
```
reversesigh, line 123: Close the set above this. Are you also running two grpc end points or was this intended to be a replacement?

aleksandrnoders: also fixed :), these are two different nodes, maybe we will remove the one above later

stratos/chain.json
```diff
@@ -88,11 +88,11 @@
       {
         "address": "https://rpc-stratos.whispernode.com:443",
         "provider": "WhisperNode 🤐"
+      },
       {
         "address": "https://stratos-rpc.noders.services:443",
         "provider": "[NODERS]TEAM"
       }
```
reversesigh, line 95: @JeremyParish69 I'm a little stumped, this change looks correct but when I view the file it's still showing me the old commit.

PR #4268, opened 2024-04-02
Correct active noble-injective IBC path

boojamya:


JeremyParish69:
Is there any website that can show that the previous connection wasn't used and that there are no abandoned tokens on either chain via the old connection?

boojamya:
Yes, 
This is the old path. It is unknown according to mintscan and has no pending packets: https://www.mintscan.io/noble/relayers/channel-10/injective/channel-146

This is the new path:
https://www.mintscan.io/noble/relayers/channel-31/injective/channel-148

PR #4269, opened 2024-04-02
Update deprecated FURY.legacy CW20 asset

squiggly-line:
image[
Coingecko listing and migration notice](https://www.coingecko.com/en/coins/fanfury).

A Cw20 version of FURY was initially minted on Juno network but was deprecated later after migration to the native Furya Chain.
The entry resides in [Cosmos registry](https://github.com/cosmos/chain-registry/blob/master/furya/assetlist.json).

Corrections have to be made to reflect the current version of the token to avoid confusion.

This is probably the step one. Any further direction regarding this would also be appreciated.

### DIFF
--- a/juno/assetlist.json
+++ b/juno/assetlist.json
@@ -671,8 +671,8 @@
         }
       ],
       "base": "cw20:juno1cltgm8v842gu54srmejewghnd6uqa26lzkpa635wzra9m9xuudkqa2gtcz",
-      "name": "Fanfury",
-      "display": "fury.legacy",
+      "name": "FURY.legacy",
+      "display": "fury",
       "symbol": "FURY.legacy",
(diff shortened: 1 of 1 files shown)


juno/assetlist.json
```diff
@@ -672,8 +672,8 @@
       ],
       "base": "cw20:juno1cltgm8v842gu54srmejewghnd6uqa26lzkpa635wzra9m9xuudkqa2gtcz",
       "name": "Fanfury",
-      "display": "fury",
-      "symbol": "FURY",
+      "display": "fury.legacy",
```
JeremyParish69, line 675: Don't update the display line. It's only used for determining the exponent--it's never actually displayed. I'd recommend instead changing the Name field.

PR #4278, opened 2024-04-03
Add gASH 

jijicodes:


migaloo/assetlist.json
```diff
@@ -382,6 +382,31 @@
...
+        },
+        {
+          "denom": "gASH",
+          "exponent": 6
+        }
+      ],
+      "base": "factory/migaloo1r9x8fz4alekzr78k42rpmr9unpa7egsldpqeynmwl2nfvzexue9sn8l5rg/gash",
+      "name": "gASH",
+      "display": "gASH",
+      "symbol": "gASH",
```
JeremyParish69, line 401: why is the g lowercase? Unless there's a trace to another asset, this should be uppercase.

jijicodes: > why is the g lowercase? Unless there's a trace to another asset, this should be uppercase.

@JeremyParish69 gASH is a variant of ASH token which is in the same asset list. I was confirmed with Ash Dao team that this is the correct capitalization. 
Just like this:
https://github.com/terra-money/station-assets/pull/270/commits/5df47e986cd240922485ad18e6b8157b0d02200e

JeremyParish69: In what way is it a variant? What's the relationship?

jijicodes: > In what way is it a variant? What's the relationship?

ASH is minted when burning WHALE and gASH is minted when burning GUPPY. both ASH and gASH are done Ash DAO's furnace dapp on Migaloo that I'm working on.

JeremyParish69: Since gASH is not a variant or derivative of ASH, I don't see the justification for the lowercase letters. I'm still thinking this must be all uppercase. The only lowercase letters accepted in symbols are when one assets derives from another, such as 'stATOM' for liquid staked ATOM.

jijicodes: > Since gASH is not a variant or derivative of ASH, I don't see the justification for the lowercase letters. I'm still thinking this must be all uppercase. The only lowercase letters accepted in symbols are when one assets derives from another, such as 'stATOM' for liquid staked ATOM.

@JeremyParish69 Okay, i changed them to be all uppercase. Thank you.

JeremyParish69: It was really only symbol that needed to be uppercase. It is ok to keep everything else to have lowercase characters.

PR #4283, opened 2024-04-04
Update of chain.json for Persistence by Antrix Validator

antrix-validators:
Added end points for Antrix Validators


JeremyParish69:
Looks like this isn't your first time here. Please use appropriate PR titles that describe the change.

antrix-validators:
Added RPC end point to persistence chain for Antrix Validators

PR #4290, opened 2024-04-04
add migaloo-terra channels

fragwuerdig:


_IBC/terra-migaloo.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "terra",
```
JeremyParish69, line 4: please reverse the order of chain_1 and chain_2 (because terra comes later, alphabetically). And the file name, too

fragwuerdig: Done

fragwuerdig:
Please, advice on how to add a second/different client/connection pair for the same pair of chains.

JeremyParish69:
> Please, advice on how to add a second/different client/connection pair for the same set of chains.

You would add additional objects inside the 'channels' array in the appropriate file.

PR #4291, opened 2024-04-04
add rakoff logo to terra classic

fragwuerdig:


JeremyParish69:
Please define the token in terra's assetlist here too

PR #4293, opened 2024-04-04
add luna2 on migaloo

fragwuerdig:


migaloo/assetlist.json
```diff
@@ -382,6 +382,52 @@
...
+      "name": "Luna",
+      "display": "luna",
+      "symbol": "LUNA",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "terra2",
+            "base_denom": "uluna",
+            "channel_id": "channel-86"
```
JeremyParish69, line 408: Pending registration of this IBC connection. The transfer<>transfer channel should be first in the array of channels, then the wasm connetion after.

fragwuerdig: There is no Transfer-Transfer channel on this connection...

fragwuerdig: wait - this literally _is_ a transfer-transfer channel. i don't understand what the problem is right now... can you please elaborate? please note: this is migaloo-terra2 and _NOT_ migaloo-terra

JeremyParish69: My bad, this is already registered. I had terra in mind

PR #4295, opened 2024-04-05
Add shido chain

NodeNavigator:
Add shido chain


shido/assetlist.json
```diff
@@ -0,0 +1,34 @@
...
+  "assets": [
+    {
+      "description": "The native EVM and Wasm, governance and staking token of the Shido Chain",
+      "denom_units": [
+        {
+          "denom": "shido",
+          "exponent": 18
+        }
+      ],
+      "base": "shido",
```
JeremyParish69, line 13: base unit must be 0 exponent. Please define another denom unit with 0 exponent above the unit 'shido'.

shido/chain.json
```diff
@@ -0,0 +1,159 @@
...
+  "staking": {
+    "staking_tokens": [
+      {
+        "denom": "shido"
+      }
+    ]
+  },
+  "codebase": {
+    "git_repo": "https://github.com/ShidoGlobal/shidochain",
+    "recommended_version": "v1.0.0",
```
JeremyParish69, line 39: Please add a version object into the `versions` array.

shido/chain.json
```diff
@@ -0,0 +1,159 @@
...
+    "genesis": {
+      "genesis_url": "https://raw.githubusercontent.com/ShidoGlobal/mainnetShidoNodeSync/main/genesis.json"
+    },
+    "versions": []
+  },
+  "logo_URIs": {
+    "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/shido/images/shido.png",
+    "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/shido/images/shido.svg"
+  },
+  "description": "Developers use Shid as the Ethereum and Wasm Chain to deploy applications of the future. Get all the functionalities of Ethereum adn Wasm with the power of IBC and Interchain composability.",
```
JeremyParish69, line 64: is "Shid" a typo?
and "adn"

MaverickCryptoz (approved):

shido/assetlist.json
JeremyParish69: There is still ambiguity about which unit one is referring to because the denom 'shido' for each unit is the same. Please set the denom for the base 0 unit to whatever the denomination onchain is. And then the denom value for the exponent 18 unit must be something different.

e.g., 

```
"denom_units": [
        {
          "denom": "ashido",
          "exponent": 0
        },
        {
          "denom": "shido",
          "exponent": 18
        }
      ],
```
 but don't just copy that example, actually make sure that the 0 exponent denom is what's using onchain.
 Once determined, that is the denom value to be used under `staking` and `fees` in the chain.json file.

PR #4296, opened 2024-04-05
Update coingecko c4e

pablokulpa:


chain4energy/assetlist.json
```diff
@@ -21,7 +21,7 @@
       "logo_URIs": {
         "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/chain4energy/images/c4e.png"
       },
-      "coingecko_id": "",
+      "coingecko_id": "chain4energy",
```
JeremyParish69, line 24: The cgid isn't in CoinGecko's API.
https://api.coingecko.com/api/v3/coins/list

pablokulpa: https://www.coingecko.com/en/coins/chain4energy

ankurdotb (approved):
This ID does show up on CoinGecko API now. Although, it's still a preview token with no price. Not sure if that makes a difference per se to the chain registry though.

PR #4312, opened 2024-04-07
Update chain.json bitsong by 🔥STAVR🔥

obajay:


reversesigh:
@obajay I've accepted a few of these already but just realized you're adding yourself in the middle of each set. If it's a new addition, it needs to be added to the bottom of the list. Can you please fix this and the remaining ones and then we'll accept?

obajay:
> @obajay I've accepted a few of these already but just realized you're adding yourself in the middle of each set. If it's a new addition, it needs to be added to the bottom of the list. Can you please fix this and the remaining ones and then we'll accept?

yes, no problem, I'll do it.
It was just more convenient for me to add in the middle so as not to make a mistake with punctuation

PR #4326, opened 2024-04-07
Add Nodeshub Endpoint For Bitcanna

NodesHub:


reversesigh:
@NodesHub If an endpoint or explorer is a new addition, it needs to be added to the bottom of the list. Please fix this and then we can merge.

PR #4375, opened 2024-04-11
Shido register

NodeNavigator:
Add shido IBC connection register


shido/assetlist.json
```diff
@@ -0,0 +1,38 @@
+{
+  "$schema": "../assetlist.schema.json",
+  "chain_name": "shido",
+  "assets": [
+    {
+      "description": "The native EVM and Wasm, governance and staking token of the Shido Chain",
+      "denom_units": [
+        {
+          "denom": "shido",
```
sunnya97, line 9: Sigh, this is pretty confusing that at the chain level, it is `shido` instead of `ashido` which is the standard for 10^-18 tokens.  But I suppose it is too late.

sunnya97 (approved):

PR #4385, opened 2024-04-12
add eth ibc assets 

JafarAz:
Added some assets (and icons) that are being transferred from Ethereum via IBC.


composable/assetlist.json
```diff
@@ -229,6 +229,326 @@
...
+      "base": "ibc/4C14B42547F6428A4E84F501D21CE5DF6C40C537635A8B0E6A5ED8FB4919BCE7",
+      "name": "Ethereum",
+      "display": "eth",
+      "symbol": "ETH",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "ethereum",
+            "base_denom": "eth-wei"
```
JeremyParish69, line 255: None of the base_denoms specified are registered as `base` in ethereum assetlist

composable/assetlist.json
```diff
@@ -229,6 +229,326 @@
...
+        }
+      ],
+      "type_asset": "ics20",
+      "base": "ibc/4C14B42547F6428A4E84F501D21CE5DF6C40C537635A8B0E6A5ED8FB4919BCE7",
+      "name": "Ethereum",
+      "display": "eth",
+      "symbol": "ETH",
+      "traces": [
+        {
+          "type": "ibc",
```
JeremyParish69, line 252: for ibc trace type, the schema of trace is incorrect. The `chain` object is missing and there is no `provider`

composable/assetlist.json
```diff
@@ -229,6 +229,326 @@
...
+        {
+          "denom": "fxs",
+          "exponent": 18
+        }
+      ],
+      "type_asset": "ics20",
+      "base": "ibc/4404E14E1FD538EF6B0972221213A0F1BE747797252AA97F8C0D547BB36BEBD7",
+      "name": "Frax Shares",
+      "display": "fxs",
+      "symbol": "FXS",
```
JeremyParish69, line 329: Not registered in ethereum

composable/assetlist.json
```diff
@@ -229,6 +229,326 @@
...
+        {
+          "denom": "frxeth",
+          "exponent": 18
+        }
+      ],
+      "type_asset": "ics20",
+      "base": "ibc/45DA1EEDE2480677BFC5B8E7A1332F92DB2EAD71D04D98343E9D2C8ADA8FAB32",
+      "name": "Frax Ether",
+      "display": "frxeth",
+      "symbol": "frxETH",
```
JeremyParish69, line 409: Not registered in ethereum

composable/assetlist.json
JafarAz: There are a few that already exist, I will submit the rest of them now

JafarAz: is it possible to register two `base` fields under one asset?

JeremyParish69: no, there is only 1, being it's contract

composable/assetlist.json
```diff
@@ -229,6 +229,350 @@
...
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "ethereum",
+            "base_denom": "wei",
+            "channel_id": ""
+          },
+          "chain": {
+            "channel_id": "",
```
JeremyParish69, line 259: Please register the ibc connection to the /IBC/ directory, then fill in the `channel_id`s and `path` here (for each new asset) to match.

composable/assetlist.json
```diff
@@ -356,7 +358,7 @@
         {
           "image_sync": {
             "chain_name": "ethereum",
-            "base_denom": "0x3432B6A60D23Ca0dFCa7761B7ab56459D9C964D0"
+            "base_denom": "0x3432b6a60d23ca0dfca7761b7ab56459d9c964d0"
```
JeremyParish69, line 361: The letter casing must match, and I don't see why this is any more (or less) correct than what's currently registered

JafarAz: the original submission included upper-case characters, following [etherscan](https://etherscan.io/token/0x3432b6a60d23ca0dfca7761b7ab56459d9c964d0) on this one

composable/assetlist.json
```diff
@@ -297,12 +297,13 @@
...
+            "channel_id": "channel-2"
           },
           "chain": {
-            "channel_id": "",
-            "path": ""
-          }
+            "channel_id": "channel-52",
+            "path": "transfer/channel-52/0x6b175474e89094c44da98b954eedeac495271d0f"
+          },
+          "provider": "Picasso"
```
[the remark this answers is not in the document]
JafarAz: removed

composable/assetlist.json
JeremyParish69: The thing with etherscan, is when you click "copy", it includes the upper-case characters, but if you manually select the address from the page or url, it will be lowercase only. This is why it's difficult to determine which is correct, even on Etherscan, which is inconsistent.

JeremyParish69:
Please remove the upload of the asset images, since they have already been registered with images in another PR.

composable/assetlist.json
JafarAz: if it must match, then we should go with lowercase. From what I've read online, lowercase is the norm with eth addresses but there can be some cases where upper case letters are found. I normally go with the format in the following section of Etherscan
Screenshot 2024-04-19 at 23 50 40

composable/assetlist.json
```diff
@@ -229,6 +229,359 @@
...
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "ethereum",
+            "base_denom": "wei",
+            "channel_id": "channel-2"
+          },
+          "chain": {
+            "channel_id": "channel-52",
+            "path": "transfer/channel-52/eth-wei"
+          },
```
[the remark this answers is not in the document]
JafarAz: my bad, fixed

composable/assetlist.json
JeremyParish69: In that case the registration for FXS must be updated as well

JafarAz: updated

composable/assetlist.json
```diff
@@ -229,6 +229,359 @@
...
+      "base": "ibc/F9D075D4079FC56A9C49B601E54A45292C319D8B0E8CC0F8439041130AA7166C",
+      "name": "Ethereum",
+      "display": "eth",
+      "symbol": "ETH",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "ethereum",
+            "base_denom": "eth-wei",
```
JeremyParish69, line 255: On ethereum, the base denom is just "wei"

JafarAz: we never ended up changing this, can I add it as an alias to the eth folder?

JeremyParish69: it must refer to the "base" property, not just any alias on that denom, which is why this must be "wei"

composable/assetlist.json
```diff
@@ -229,6 +229,359 @@
       "logo_URIs": {
         "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/tinkernet/images/tnkr.svg"
       }
+    },
+    {
+      "description": "The native token of Ethereum, bridged via IBC.",
+      "denom_units": [
+        {
+          "denom": "wei",
```
JeremyParish69, line 237: for all ibc/ assets, the denom for the exp: 0 unit must be the ibc denom, with anything else as a n alias.
e.g., 
```
{
  "denom": "ibc/F9D075D4079FC56A9C49B601E54A45292C319D8B0E8CC0F8439041130AA7166C",
  "exponent": 0,
  "aliases": [
    "wei"
  ]
},
```

PR #4400, opened 2024-04-14
Update chain.json

omerlavanet:
remove http link

PR #4401, opened 2024-04-14
HTTPS for Lava REST API

staking-explorer:


reversesigh:
Which is correct? https://github.com/cosmos/chain-registry/pull/4400

omerlavanet:
Appreciate the correction, adding secure is preferable to removing the endpoint and I assumed it was insecure

staking-explorer:
Added SSL for Lava REST API endpoint, please merge with https.

PR #4421, opened 2024-04-17
Wardenprotocol buenavista testnet

jlehtimaki:
This PR will move old alfama testnetwork to secondary place and add new buenavista testnetwork for Warden Protocol.


testnets/wardenprotocoltestnet2/chain.json
```diff
@@ -0,0 +1,119 @@
+{
+  "$schema": "../../chain.schema.json",
+  "chain_name": "wardenprotocoltestnet2",
+  "status": "live",
```
JeremyParish69, line 4: will this remain live?

testnets/wardenprotocoltestnet2/chain.json
```diff
@@ -0,0 +1,119 @@
+{
+  "$schema": "../../chain.schema.json",
+  "chain_name": "wardenprotocoltestnet2",
```
JeremyParish69, line 3: Just double checking on the desired chain name; typically archiving a network iterates a counter to 1 first, instead of 2. Are we sure we don't want to start with archiving as chain_name: 'wardenprotocoltestnet1'?
Whichever it will be, the new assetlist must also have it's chain_name property updated, and so far it was missed.

testnets/wardenprotocoltestnet2/chain.json
jlehtimaki: Yes, this will remain live for now.

testnets/wardenprotocoltestnet2/chain.json
jlehtimaki: Hey @JeremyParish69 ,
Yes I think you are right, I've now removed Alfama since it's not active anymore.
We will start following the numbering convention from here onwards, e.g. next one will have the "2" in it.
Thank you.

PR #4480, opened 2024-04-24
Sei Testnet (atlantic-2): Update Global min gas price

codebycarson:
In accordance with the latest Sei governance proposal and to match the other Cosmos chain registries this PR updates the Global Min Gas Price to 0.08 as outlined in the Sei Chain.

https://github.com/sei-protocol/chain-registry/pull/58/files

### DIFF
--- a/testnets/seitestnet2/chain.json
+++ b/testnets/seitestnet2/chain.json
@@ -16,7 +16,7 @@
     "fee_tokens": [
       {
         "denom": "usei",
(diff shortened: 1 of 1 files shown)


JeremyParish69:
This change looks fine, but just as a heads up we aught to switch 'seitestnet' to 'seitestnet1', and 'seitestnet2' to 'seitestnet'. Becasuse the name without the incrementing suffix should always be the chain_name of the currently used chain.

PR #4481, opened 2024-04-24
Add Noble USDC to Neutron and update Arena token

ismellike:
Renames Axelar USDC on Neutron to axlusdc
Adds Noble USDC to Neutron's assets
Removes Arena token from Juno's assets and images
Adds Arena's images to Neutron

Future PR:
Will add Arena token to Neutron assets
Will add image_sync to junotestnet's Arena token when released on Neutron


neutron/assetlist.json
```diff
@@ -94,14 +94,14 @@
...
-          "denom": "usdc",
+          "denom": "axlusdc",
           "exponent": 6
         }
       ],
       "type_asset": "ics20",
       "base": "ibc/F082B65C88E4B6D5EF1DB243CDA1D331D002759E938A0F5CD3FFDC5D53B3E349",
       "name": "IBC usdc",
-      "display": "usdc",
+      "display": "axlusdc",
```
JeremyParish69, line 104: This doesn't rename the asset. All this property does is inform the number of decimals to show by pointing for one of the denom_units.
To rename the asset to be able to differentiate it from Noble's USDC, I'd recommend updating the `name` and `symbol` to something like "USD Coin (Axelar)" and "axlUSDC", resp.

testnets/junotestnet/assetlist.json
```diff
@@ -84,17 +84,13 @@
         }
       ],
       "logo_URIs": {
-        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/juno/images/arena.png",
-        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/juno/images/arena.svg"
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/arena.png",
```
JeremyParish69, line 87: This image you're pointing to doesn't exist. I'd recommend having this PR only add USDC to neutron. Then in the next PR, when the new token is created, we can focus on deprecating the Juno version (not deleting it), and registering the new ARENA.

JeremyParish69: Good to know. And you're right the images are just being moved, not deleted. I would still ask you use image_sync when you can point it to neutron's

PR #4500, opened 2024-04-26
composable name and prefix change

JafarAz:
This PR is being requested in advance to a mainnet upgrade which we are planning to schedule for next week. 

We want to make it so the chain registry has the following changes:

1. changing the "chain name" from "composable" to "picasso."
2. Aligning the "pretty_name" to "Picasso"
3. "website": "https://picasso.xyz/"
4. adjusting the "bech32_prefix" from "centauri" to "pica"
5. changing the "daemon_name" from "centaurid" to "picad"
6. changing the chain name of "picasso" to "picassokusama" to show the difference

I also edited the IBC files that were referencing composable.

We are keeping the chain id as centauri-1 to avoid state-breaking changes. chain id won't ever be user-facing so it shouldn't be too confusing.

The upgrade has successfully taken place on our testnet. Please lmk if these changes are gtg and ready to be requested in the chain registry. **Please do not merge it until we have completed the chain upgrade.**


JafarAz:
is CI failing because the order of the chains in the json need to match the order of the chains in the name of the json?

picasso/chain.json
```diff
@@ -32,7 +32,7 @@
     ]
   },
   "codebase": {
-    "git_repo": "https://github.com/notional-labs/composable-centauri",
+    "git_repo": "https://github.com/composable/composable-cosmos",
     "recommended_version": "v6.5.3",
```
hoank101, line 36: v6.6.1 please

JafarAz: Not sure if this will be the same release for mainnet, will add once we confirm it

JeremyParish69:
1. changing the "chain name" from "composable" to "picasso."
I don't think we can do this. chain_name's changes are problematic and has never happened before. There are other places that refer to the chain_name that would break.

picasso/chain.json
```diff
@@ -1,13 +1,13 @@
...
   "chain_id": "centauri-1",
-  "website": "https://www.composable.finance/",
-  "pretty_name": "Composable",
+  "website": "https://www.picasso.xyz/",
+  "pretty_name": "Picasso",
   "status": "live",
   "network_type": "mainnet",
-  "bech32_prefix": "centauri",
-  "daemon_name": "centaurid",
+  "bech32_prefix": "pica",
```
JeremyParish69, line 9: This prefix should be registered

composable/chain.json
```diff
@@ -2,12 +2,12 @@
...
   "chain_id": "centauri-1",
-  "website": "https://www.composable.finance/",
-  "pretty_name": "Composable",
+  "website": "https://www.picasso.xyz/",
+  "pretty_name": "Picasso",
   "status": "live",
   "network_type": "mainnet",
-  "bech32_prefix": "centauri",
-  "daemon_name": "centaurid",
+  "bech32_prefix": "pica",
```
JeremyParish69, line 9: Please register this prefix https://github.com/satoshilabs/slips/blob/master/slip-0173.md

JafarAz: registered, please lmk if the PR needs further changes, the chain upgrade is scheduled for wednesday

PR #4504, opened 2024-04-28
Added native USDt to assetlist on Terra

Niels277:
See title

### DIFF
--- a/terra2/assetlist.json
+++ b/terra2/assetlist.json
@@ -959,6 +959,50 @@
       ]
     },
     {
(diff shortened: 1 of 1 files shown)


terra2/assetlist.json
```diff
@@ -958,6 +958,50 @@
         }
       ]
     },
+    {
+      "description": "Tether USDt on Terra",
```
JeremyParish69, line 962: please relocate this asset to the bottom of the list

Niels277: done

PR #4506, opened 2024-04-28
Update quicksilver-2 to v1.5.5

joe-bowman:
https://www.mintscan.io/quicksilver/proposals/37
https://www.mintscan.io/quicksilver/block/6926000


joe-bowman:
Yes, the hotfix was a fix to a single GRPC query that was affecting the front end. It is 100% consensus compatible.

PR #4510, opened 2024-04-30
adding onex chain

laloquidity:


onex/assetlist.json
```diff
@@ -0,0 +1,36 @@
+{
+  "$schema": "../assetlist.schema.json",
+  "chain_name": "ONEX",
```
JeremyParish69, line 3: must be lowercase letters

onex/chain.json
```diff
@@ -0,0 +1,123 @@
...
+        "address": "a.node.mainnet.onomy.io:19091",
+        "provider": "pendulum"
+      }
+    ]
+  },
+  "explorers": [
+    {
+      "kind": "Dexplorer - must enter rpc link provided above",
+      "url": "https://dexplorer.cakralabs.site",
+      "tx_page": "https://dexplorer.cakralabs.site/txs/${txHash}",
```
JeremyParish69, line 116: Will this tx URL not work, then, if the user has to find an RPC to connect to? I'm not sure other interfaces will be able to make use of this

onex/chain.json
```diff
@@ -0,0 +1,123 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "onex",
+  "status": "live",
+  "website": "https://app.onomy.io/",
+  "network_type": "mainnet",
+  "pretty_name": "ONEX",
+  "chain_id": "onex-mainnet-1",
+  "bech32_prefix": "onomy",
```
JeremyParish69, line 9: This bech32 prefix is already taken by Onomy

laloquidity: the bech32 prefix is set to onomy. community voted to keep prefixes the same for the onomy ecosystem so that all tokens in the ecosystem can be sent to the same address. ONEX is a consumer chain to the Onomy parent chain.

PR #4517, opened 2024-04-30
Update DYDX versioning post v4.x

2dpetkov:


dydx/chain.json
```diff
@@ -59,7 +59,7 @@
     },
     "versions": [
       {
-        "name": "v2",
+        "name": "v2.0",
```
reversesigh, line 62: "name" has to match the upgrade_handler - in this case it's fine since the chain started with v2 - but for later modifications you made (i.e. v3.0.0 to v3.0) are not valid as the upgrade_handler is v3.0.0.

dydx/chain.json
```diff
@@ -94,10 +94,10 @@
           "type": "cometbft",
           "version": "dydxprotocol/cometbft v0.37.3-0.20230908230338-65f7a2f25c18"
         },
-        "next_version_name": "protocol/v4.0.0"
+        "next_version_name": "v4.0"
```
reversesigh, line 98: Good catch here - but just name is "v4.0.0" to match the upgrade_handler

dydx/chain.json
2dpetkov: got it!

dydx/chain.json
```diff
@@ -94,25 +95,25 @@
           "type": "cometbft",
           "version": "dydxprotocol/cometbft v0.37.3-0.20230908230338-65f7a2f25c18"
         },
-        "next_version_name": "protocol/v4.0.0"
+        "next_version_name": "v4.0.0"
```
dylanschultzie, line 98: Great catch

PR #4520, opened 2024-05-01
Update TIA chain.json

faneaatiku:
Updated explorer links that are using ping.pub to use type ping.pub

### DIFF
--- a/celestia/chain.json
+++ b/celestia/chain.json
@@ -457,7 +457,7 @@
       "account_page": "https://mintscan.io/celestia/address/${accountAddress}"
     },
     {
-      "kind": "NodeStake",
+      "kind": "ping.pub",
       "url": "https://explorer.nodestake.top/celestia/",
       "tx_page": "https://explorer.nodestake.top/celestia/txs/${txHash}",
       "account_page": "https://explorer.nodestake.top/celestia/account/${accountAddress}"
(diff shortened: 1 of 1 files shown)


reversesigh:
"kind" is a little misleading here, but the "kind" should refer to the host of the explorer rather than what is used as the base.

faneaatiku:
It's obviously not my choice what "kind" is but as a DApp developer, when i first encountered "kind" key here I thought about "type". You can reject the PR if i misunderstood. 

But in this case I want to suggest adding a new key called "gov_proposal_page" (like "tx_page" and "account_page"), so developers like me can use any explorer when they want to redirect the user to a proposal page. I had to pick ping.pub since it's the most popular and I can guess the URL to the gov proposal page. I honestly thought it's a no brainer what "kind" means. :)

faneaatiku:
As you can see when I submitted this file https://github.com/cosmos/chain-registry/blob/master/beezee/chain.json 1 year ago "ping.pub" was accepted as "kind". Now I have to find a fallback when i don't find "ping.pub" kind for a chain LOL

reversesigh:
> It's obviously not my choice what "kind" is but as a DApp developer, when i first encountered "kind" key here I thought about "type". You can reject the PR if i misunderstood.
> 
> But in this case I want to suggest adding a new key called "gov_proposal_page" (like "tx_page" and "account_page"), so developers like me can use any explorer when they want to redirect the user to a proposal page. I had to pick ping.pub since it's the most popular and I can guess the URL to the gov proposal page. I honestly thought it's a no brainer what "kind" means. :)

@JeremyParish69 do you have thoughts on adding these for explorer entries? Makes sense to me.

JeremyParish69:
I'm we should keep it to reflect 'host' as much as possible and move away from 'kind'. If I had the time I'd push for renaming the property for the whole registry

reversesigh:
> I'm we should keep it to reflect 'host' as much as possible and move away from 'kind'. If I had the time I'd push for renaming the property for the whole registry

I'm more so speaking on his point adding "gov_proposal_page" to explorers.

JeremyParish69:
There is now a proposal_page for explorers.
Closing PR due to inactivity.

PR #4428, opened 2024-04-18
fix: update ibc details for dhealth-osmosis

nsonanh:
Hi, my apology for the changes again, this is our final update.

Some information:
- We are an on-reviewed asset with no pool created yet.
- There's currently only one un-relayed transaction which was from our side: https://explorer.dhealth.com/dhealth/transactions/73D1C0525D470CF930B176894E3B32DB103E2D6CB5E646F9183BD9EEF98D2DCC

PR #4534, opened 2024-05-03
fix: update dhealth-osmosis ibc data

nsonanh:
Some information:

Osmosis assetlists ref: https://github.com/osmosis-labs/assetlists/pull/1521
We are an on-reviewed asset with no pool created yet.

### DIFF
--- a/_IBC/dhealth-osmosis.json
+++ b/_IBC/dhealth-osmosis.json
@@ -2,22 +2,22 @@
   "$schema": "../ibc_data.schema.json",
   "chain_1": {
     "chain_name": "dhealth",
-    "client_id": "07-tendermint-2",
-    "connection_id": "connection-7"
+    "client_id": "07-tendermint-4",
+    "connection_id": "connection-9"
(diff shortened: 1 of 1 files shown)


JeremyParish69:
There have already been tokens transferred. What's wrong with reviving the old channel?

nsonanh:
We had a quite short unbonding time, which could not prevent the Client from expiring due to our relayer maintenance or chain upgrade. As a result our previous channel was expired.
The unbonding time has been increased in our latest proposal: https://ping.pub/dhealth/gov/3 and it'll be very unlikely that we will need to make a pull request again.
Since we have been an unverified asset, with no pool created yet and no one has used the relayer at all, we decided that creating a new channel would be better, also to prevent confusion from the community.

> Note the conversation about it not changing again: https://github.com/cosmos/chain-registry/pull/4428#issuecomment-2065314802

We understand that this is not according to our comment last time, but it'll be very likely the last time we change our ibc data.

PR #4536, opened 2024-05-03
Add Blockval services for Comdex & Andromeda

allomart1:


andromeda/chain.json
```diff
@@ -145,6 +145,12 @@
         "id": "28876b3094518bef97a1250ef641c26b7d4a658d",
         "address": "138.201.21.197:39656",
         "provider": "StakeTown"
+      },
+      ,
```
JeremyParish69, line 149: this comma breaks the schema

allomart1: Thanks, fixed

PR #4554, opened 2024-05-06
Upgrade Neutron to v3.0.5

reversesigh:
https://governance.neutron.org/proposals/37

https://www.mintscan.io/neutron/block/10525000


reversesigh:
For right now, would it be acceptable to remove graphql from both chain.json and the schema until we make a decision on how to handle?

ankurdotb (approved):

PR #4563, opened 2024-05-07
Add Gluon (fka UnUniFi)

Senna46:
UnUniFi has been rebranded to Gluon.
UnUniFi will remain as is, and will be updated this in the future.


JeremyParish69:
Looking at the version history, we can see this is an update of ununify, so I think we should not create anew gluon directory, but just update the branding for ununify.

Senna46:
@JeremyParish69  Currently, there is a mainnet for UnUniFi and a testnet for Gluon. 
We would therefore like to register both; UnUniFi will be removed later.

JeremyParish69:
> @JeremyParish69 Currently, there is a mainnet for UnUniFi and a testnet for Gluon. We would therefore like to register both; UnUniFi will be removed later.

Is gluon a testnet for ununify?

gluon/chain.json
```diff
@@ -0,0 +1,455 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "gluon",
+  "status": "live",
+  "network_type": "mainnet",
```
JeremyParish69, line 5: You have network type here is mainnet. Is this a mainnet or a testnet?

JeremyParish69: or is there two, simultaneously running mainnets?

JeremyParish69:
Is there still interest in proceeding with this @Senna46 ?

PR #4578, opened 2024-05-08
feat: wss + nibiru

cgilbe27:
Add WSS to template and to nibiru


reversesigh:
For right now, would it be acceptable to remove graphql from both chain.json and the schema until we make a decision on how to handle?

cgilbe27:
Fixed @reversesigh

PR #4580, opened 2024-05-08
Update Nim Network

Cordtus:
Added git_repo and genesis_url to codebase on Nim chain.json

### DIFF
--- a/nim/chain.json
+++ b/nim/chain.json
@@ -40,6 +40,10 @@
     }
   ],
   "codebase": {
+    "git_repo": "https://github.com/Nim-Network-Foundation/mainnet",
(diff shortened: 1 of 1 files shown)


nim/chain.json
```diff
@@ -40,6 +40,10 @@
     }
   ],
   "codebase": {
+    "git_repo": "https://github.com/Nim-Network-Foundation/mainnet",
+    "genesis": {
+      "genesis_url": "https://github.com/Nim-Network-Foundation/mainnet/blob/main/genesis.json"
```
reversesigh, line 45: This should link to the raw version instead

PR #4588, opened 2024-05-10
Add AVIAONE endpoints for Dymension

AviaOne:


dymension/chain.json
```diff
@@ -94,6 +94,11 @@
         "id": "86bd5cb6e762f673f1706e5889e039d5406b4b90",
         "address": "seed.dymension.node75.org:10956",
         "provider": "Pro-Nodes75"
+      },
+	        {
```
reversesigh, line 98: Can you fix the formatting here and the other spots below?

AviaOne (commented):
is it ok now ?

PR #4643, opened 2024-05-16
Correct CoinGecko IDs for USDC.grv and USDT.axl

blake-regalia:
This PR simply corrects the CoinGecko IDs for two assets, which were accidentally set to the slug URLs rather than the API IDs.


JeremyParish69:
Thanks for the correction. We'll be more mindful of this going forward.

PR #4654, opened 2024-05-17
Fix typo for dyson chain

tdnguyenND:


ankurdotb (approved):

JeremyParish69:
It is common practice to name the image file as the asset symbol, which is DYS, so I see no typo. Renaming files can mildly break apps temporarily, but not impossible to fix. Are we sure we want this renamed?

PR #4656, opened 2024-05-17
Update rpc/api for apple juice validator

druhtru:


ankurdotb (approved):
`:443` suffix with HTTPS is a bit redundant but not incorrect

druhtru:
> `:443` suffix with HTTPS is a bit redundant but not incorrect

I know this, but sometimes users face the problem of missing ports when using rpc, so I decided to specify them

PR #4660, opened 2024-05-17
Fix Broken Osmosis::versions.json

pyramation:
fixing broken JSON, how did this pass?


jonator (approved):
Yeah agreed. An action seems trivial to set up for this, and long overdue

PR #4688, opened 2024-05-23
Creates mantrachain dir, chain.json, replaces testnet logos with mainnet

Cordtus:
Created `mantrachain` dir (no chain.json, only preliminary assets.json ahead of mainnet launch)
Added multiple logo variants
Removed images sub-dir from `mantrachaintesnet`, replaced URLs in asset/chain json with mainnet versions


Cordtus:
@JeremyParish69 please review

PR #4707, opened 2024-05-24
Update oraichain logos for better darkmode display

tucq88:
Replace old ones for better darkmode display


tucq88:
GM @JeremyParish69 , 

Wanna provide a bit more context: Because many apps (e.g ibc.fun) is using both two images, so I keep both even they look the same to make sure no change is required from apps 

Secondly, by splitting different names, it can be updated later without create any breaking change, we agree now they look similar. So I believe keeping like this is better. 

Any further suggestions ? Thank you.

JeremyParish69:
> Any further suggestions ? Thank you.

Thanks for the explanation, but I disagree with this plan. If apps are using the Chain Registry for images, they should be referring to the registry (and not breaking) as opposed to hardcoding the URL and breaking.

Secondly, when updating the png image, the svg image must be changed in the same way. We cannot have orai.png looking different than orai.svg.

I recommend not replacing the old images, but just adding the new image, and putting that at the top of the `images` array.

JeremyParish69:
@tucq88 I'd like to propose this change instead, which sets the first and primary image to be the one you've added

https://github.com/cosmos/chain-registry/pull/4788 
What do you think?

JeremyParish69:
bump?

tucq88:
gm @JeremyParish69 . I think it's a great idea. Let's proceed it on your PR.

PR #4711, opened 2024-05-24
Add LP tokens to assetlist on Terra

Niels277:
3 LP tokens: Atom-Luna , Luna-USDC and Luna-USDT . Ura dex (hence Ura icon)


OKEAMAH (commented):
Commit

PR #4717, opened 2024-05-27
Added ashHUAHUA

jijicodes:


chihuahua/assetlist.json
```diff
@@ -234,6 +234,34 @@
...
+          "denom": "ashHUAHUA",
+          "exponent": 6
+        }
+      ],
+      "type_asset": "sdk.coin",
+      "address": "chihuahua1hplyuj2hzxd75q8686g9vm3uzrrny9ggvt8aza2csupgdp98vg2sp0e3h0",
+      "base": "factory/chihuahua1hplyuj2hzxd75q8686g9vm3uzrrny9ggvt8aza2csupgdp98vg2sp0e3h0/uhuahua.ash",
+      "name": "Burned HUAHUA",
+      "display": "ashHUAHUA",
+      "symbol": "ashHUAHUA",
```
JeremyParish69, line 256: This symbol has lowercase letters, meaning it derives from HUAHUA in some way. How so? Please define the relationship using the `traces` property.

jijicodes: @JeremyParish69 
ashHUAHUA is minted when burning HUAHUA token. This is done by Ash DAO's furnace dapp- 
https://chihuahua.burn.community/

Just like ashLAb - https://github.com/cosmos/chain-registry/pull/4516

JeremyParish69: Thanks. I'm still not clear exactly what the purpose is. Can it be redeemed to turn ashHUAHUA back into HUAHUA? Or is it more like a 'receipt' of having burned HUAHUA? Does it earn the user rewards or anything?
I'm trying to figure out whether both of these should be defined as a 'synthetic' version of the original asset, but that would only make sense if it's meant to follow the same price action, which I'm not sure is the case here.

JeremyParish69: For now I can't think of an appropriate example, so I'm not sure the current usage of traces will work with this. For now I think we can proceed without it.

PR #4723, opened 2024-05-28
Update GEM Icon in chain-registry per Passed Gov Prop

Studio-Mojo:


Studio-Mojo:
gem

This is the corrected image size to be use for gem.png 

Thank you

JeremyParish69:
The original image has lowercase letters, but the upload titles it with uppercase, which will cause issues. Please rename the image to 'gem.png' (in lowercase).

Studio-Mojo:
@JeremyParish69  Thanks for that catch.  
Also first time working with this process, so lots of learning on my end happening in real time. 

I delete the file and uploaded the proper sized image with proper lowercase.

PR #4732, opened 2024-05-29
add grpc, change api/rpc apple juice validator for shido

druhtru:


druhtru:
[JeremyParish69](https://github.com/JeremyParish69) can u merge this

PR #4736, opened 2024-05-29
Add GetBlock RPC to Cosmoshub

JoGetBlock:
Add GetBlock's public endpoints to the list


JeremyParish69:
Next time please use an appropriate PR title

PR #4747, opened 2024-05-31
Added coingecko_id for XRP token in COREUM assets list

rixhy-cyd:


coreum/assetlist.json
```diff
@@ -61,6 +61,7 @@
       "name": "Ripple",
       "display": "xrp",
       "symbol": "XRP",
+      "coingecko_id": "ripple",
```
JeremyParish69, line 64: Bridged tokens aren't allowed to reuse a coingecko_id. Only the original asset and additional-mintages can reuse a coingecko id.

PR #4765, opened 2024-06-03
Update agoric/chain.json for agoric-upgrade-15

gibson042:
https://ping.pub/agoric/gov/74

### DIFF
--- a/agoric/chain.json
+++ b/agoric/chain.json
@@ -44,16 +44,7 @@
     ],
     "genesis": {
       "genesis_url": "https://main.agoric.net/genesis.json"
-    },
-    "versions": [
-      {
-        "name": "agoric-upgrade-8",
-        "recommended_version": "agoric-upgrade-8",
-        "compatible_versions": [
-          "agoric-upgrade-8"
-        ]
-      }
-    ]
+    }
   },
   "logo_URIs": {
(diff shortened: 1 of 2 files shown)


agoric/chain.json
```diff
@@ -38,19 +38,24 @@
...
+      "agoric-upgrade-15"
     ],
     "genesis": {
       "genesis_url": "https://main.agoric.net/genesis.json"
     },
     "versions": [
       {
-        "name": "agoric-upgrade-8",
-        "recommended_version": "agoric-upgrade-8",
+        "name": "agoric-upgrade-15",
```
JeremyParish69, line 51: Please retain upgrade-8 in the version history

gibson042: Is `versions` expected to be complete? The agoric chain has had [many releases](https://github.com/Agoric/agoric-sdk/releases), and we just realized in the latest one that we've been lax in documenting them here.

JeremyParish69: At the Chain Registry we try to at least retain the history of versions that have ever been registered, but, yes, ideally versions contains a complete history. If it's quite long it can be relocated to the 'versions.json' file.

gibson042: OK, I split out the versions.json file and we will backfill it after this PR lands.

PR #4766, opened 2024-06-03
Update archway-composable channel

PolkachuIntern:


JeremyParish69:
Note: I merged this without question because I know that this connection was initially registered only a couple hours ago. Otherwise updating the connection would be more difficult.

PR #4768, opened 2024-06-03
Shitmos chain registry pr 1

flarnrules:
Updates chain registry for Migaloo chain for asset that has been IBC'd to Migaloo from Osmosis: Shitmos. Also includes images for token logos (png and svg).


migaloo/assetlist.json
```diff
@@ -647,6 +647,54 @@
...
+        {
+          "image_sync": {
+            "chain_name": "osmosis",
+            "base_denom": "factory/osmo1q77cw0mmlluxu0wr29fcdd0tdnh78gzhkvhe4n6ulal9qvrtu43qtd0nh8/shitmos"
+          },
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/migaloo/images/shitmos.png",
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/migaloo/images/shitmos.svg"
+        }
+      ],
+      "logo_URIs": {
```
JeremyParish69, line 694: Please remove the legacy `logo_URIs` property

migaloo/assetlist.json
```diff
@@ -647,6 +647,54 @@
...
+          }
+        }
+      ],
+      "images": [
+        {
+          "image_sync": {
+            "chain_name": "osmosis",
+            "base_denom": "factory/osmo1q77cw0mmlluxu0wr29fcdd0tdnh78gzhkvhe4n6ulal9qvrtu43qtd0nh8/shitmos"
+          },
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/migaloo/images/shitmos.png",
```
JeremyParish69, line 690: The logo images for assets should be kept in the /images/ directory corresponding to their original chain (in this case, in osmosis' /images directory). There is already a square image in osmosis' /images directory, https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/SHITMOS.png, so we don't need really these. But if you'd like to add a round version (because the original is square), you can do so by adding another object into the `images` array for the original osmosis asset.

migaloo/assetlist.json
flarnrules: I removed that object per your instructions, and pushed updated code.

migaloo/assetlist.json
flarnrules: I moved the images to the osmosis subdirectory, and changed the path to the images to properly refer to those images in the osmosis subdirectory.

PR #4769, opened 2024-06-04
Upgrade SGE Network to v1.7.0

reversesigh:
https://explorer.tcnetwork.io/sge/proposal/9

https://explorer.tcnetwork.io/sge/block/3976350


reversesigh:
Converting to draft as upgrade height reached before prop ended, requiring a new prop and height

PR #4788, opened 2024-06-05
Add new ORAI Token Logo Image

JeremyParish69:
Add new ORAI Token Logo Image
solid white background, more suitable for interfaces in dark mode, but also work with light mode


tucq88:
This is a great idea than my PR #4707 
Thanks @JeremyParish69 for suggestion. Let's proceed it this way.

PR #4793, opened 2024-06-06
Upgrade to v1.0.2 Swisstronik_testnet

kii-awesome:
- change denom to `aswtr`


testnets/swisstroniktestnet/chain.json
```diff
@@ -15,24 +15,27 @@
     "fees": {
         "fee_tokens": [
             {
-                "denom": "uswtr",
+                "denom": "aswtr",
```
reversesigh, line 18: You'll need to update assetlist.json to reflect this change as well.

PR #4799, opened 2024-06-07
deprecate BERLIN

sirnicolaz:
We are going to transition to a new type of token on the EVMOS chain which will allow us to have a native IBC BERLIN token in the future. Reason why we are deprecating the current BERLIN token into a legacy token.

### DIFF
--- a/evmos/assetlist.json
+++ b/evmos/assetlist.json
@@ -111,21 +111,21 @@
       }
     },
     {
-      "description": "The token of Teledisko DAO.",
+      "description": "The legacy token of Teledisko DAO.",
       "denom_units": [
         {
           "denom": "erc20/0x1cFc8f1FE8D5668BAFF2724547EcDbd6f013a280",
           "exponent": 0
         },
         {
-          "denom": "berlin",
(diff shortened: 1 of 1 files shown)


evmos/assetlist.json
```diff
@@ -111,21 +111,21 @@
...
           "exponent": 18
         }
       ],
       "base": "erc20/0x1cFc8f1FE8D5668BAFF2724547EcDbd6f013a280",
-      "name": "Teledisko DAO",
-      "display": "berlin",
-      "symbol": "BERLIN",
+      "name": "Teledisko DAO - Legacy",
+      "display": "berlin-legacy",
+      "symbol": "BERLIN-legacy",
```
JeremyParish69, line 128: Thoughts on BERLIN.legacy instead?

sirnicolaz: uhm would have been an option, but I'd have now to change again the metadata on chainlist... is this a deal breaker?

JeremyParish69: I'm not too sure, there isn't a super strict standard at the moment.

PR #4801, opened 2024-06-08
Update stale entries - Furya

sudopunk:
This Revision contains:

1. FURY asset and Furya chain have different images as per the chain's branding. Update to differentiate both.
2. FURY asset icons were named by their denom notation that caused issues with fetching the icon in some wallets. Change that to the actual asset name.
3. Update explorer link with newer version of the build with the new chain domain URL.
4. Update recommended chain binary version to v2.0.0 as chain has been upgraded.
5. Update IBC connection and channel info for Furya-Osmosis relayer

Update:
Furya-Terra2 IBC info.


sudopunk:
@JeremyParish69 We are hoping to see the merge happen soon as it is critical to our upcoming campaign plans.
Please let me know if you need anymore clarification during your review.

furya/chain.json
```diff
@@ -29,9 +29,9 @@
   },
   "codebase": {
     "git_repo": "https://github.com/furysport/furya-chain",
-    "recommended_version": "v1.0.0",
+    "recommended_version": "v2.0.0",
```
JeremyParish69, line 32: Please add the new version object in the `versions` array as well (while keeping v1 before it)

_IBC/furya-osmosis.json
```diff
@@ -2,22 +2,22 @@
...
   "channels": [
     {
       "chain_1": {
-        "channel_id": "channel-3",
+        "channel_id": "channel-8",
         "port_id": "transfer"
       },
       "chain_2": {
-        "channel_id": "channel-8690",
+        "channel_id": "channel-74222",
```
JeremyParish69, line 20: There are already existing IBC connections that have been registered for a long time. Why are new connections being registered? Can we prove that both of these connections have no abandoned tokens on the counterparty chain?

furya/chain.json
sudopunk: As per your instruction, I have the objects updated in the latest commit.
I've followed the same convention as the other chains. Please let me know if this requires further revision.

_IBC/furya-osmosis.json
sudopunk: Creating liquidity pools on Osmosis requires having a reliable channel open to the Osmosis chain. We find the new relayers to be reliable. The one for Terra2 will also be currently managed by the same provider.
I have asked for a confirmation from the providers about the any abandoned assets and a proper halting of the older relayer. I'll keep you posted.

JeremyParish69: Can these reliable relayers not just operate on the existing channels?

qf3l3k: Previous channels expired, so new ones had to be created anyway. At the moment it is just under discussion with the team if we should raise governance to replace old with new or just use new channels.

JeremyParish69: I have osmosisd installed, and it looks like there is a total of 1.000009 FURY on Osmosis, which is much less than 1 penny, and so can be considered a negligible 'testing amount'.
I'm not sure how much OSMO is on Furya. If we can show that's it's also a tiny amount then we should be able to update.

JeremyParish69: It has also been determined that no Osmosis assets are on Furya, so I think we are OK to proceed with new channels. Is this still desired?

sudopunk: Yes, please. Let's proceed with the new channels.

PR #4804, opened 2024-06-09
add back logo uri objects for shitmos so skip protocol works

flarnrules:
Previously pushed assetlist.json code for shitmos on Migaloo chain, and was asked to remove the "logo_URIs" object because it had been deprecated. I was confused because I could see that many assets use that object. I complied, but this PR reverses that change. The removal of the logo URI object from the osmosis side and migaloo side has made shitmos no longer compatible with skip protocol's IBC Fun, which is crucial infrastructure for shitmos users.

See screenshot of conversation with Skip Protocol dev:
Screenshot 2024-06-08 214625


JeremyParish69:
Apologies for the issues and thanks for sharing context. Skip is significant enough that we can justify adding back in for now. But I intend to align with them on expected properties.

PR #4810, opened 2024-06-11
Add ampgash

0xphilipp:


migaloo/assetlist.json
```diff
@@ -695,6 +695,31 @@
         "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/shitmos.png",
         "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/osmosis/images/shitmos.svg"
       }
+    },
+    {
+      "description": "ERIS Liquid Alliance Staked GASH",
```
JeremyParish69, line 700: Is there already a token called GASH somewhere? I think this should be defined with `traces` pointing its origin as GASH liquid-staking

0xphilipp: Yes there is, do you have an example of how the traces should be defined?

0xphilipp: Would this be correct?

```
[
        {
          "type": "liquid-stake",
          "chain": {
            "contract": "migaloo1nsskhvvh0msm7d5ke2kfg24a8d4jecsnxd28s27h0uz5kf9ap60shlqmcl"
          },
          "provider": "ERIS Protocol",
          "counterparty": {
            "chain_name": "migaloo",
            "base_denom": "factory/migaloo1r9x8fz4alekzr78k42rpmr9unpa7egsldpqeynmwl2nfvzexue9sn8l5rg/gash",
            "contract": "migaloo1r9x8fz4alekzr78k42rpmr9unpa7egsldpqeynmwl2nfvzexue9sn8l5rg"
          }
        }
      ]
```

JeremyParish69: Yes, I believe so

PR #4815, opened 2024-06-12
Add DTEAM's peer, rest API endpoint, RPC endpoint to chain4energy chain.json

doodleoo3:


chain4energy/chain.json
```diff
@@ -294,6 +294,11 @@
       }
     ],
     "persistent_peers": [
+      {
+        "id": "00eb92db69a4848f8ac1c84e7b8b2d9b01b95dff",
+        "address": "peer.c4e.mainnet.dteam.tech:30656",
+        "provider": "DTEAM"
+      }
```
reversesigh, line 301: Add a comma here

doodleoo3: done

chain4energy/chain.json
```diff
@@ -678,6 +695,11 @@
       "url": "https://explorer.apeironnodes.com/chain4energy",
       "tx_page": "https://explorer.apeironnodes.com/chain4energy/transactions/${txHash}"
     },
+    {
+      "kind": "explorer",
```
JeremyParish69, line 699: I think kind here should be the host (DTEAM)

doodleoo3: yep

just changed it, thanks

PR #4820, opened 2024-06-13
Update "$schema" Property for Root-level schema Definitions

JeremyParish69:
Try absolute $schema Path


pyramation:
This won't work

BEFORE:

Screen Shot 2024-06-13 at 12 54 15 PM
Screen Shot 2024-06-13 at 12 54 02 PM

AFTER:

Screen Shot 2024-06-13 at 12 54 40 PM

PR #4822, opened 2024-06-13
Register Upcoming Assets

JeremyParish69:
Initia's Socials, Prysm, Namada, Coinhall, Movement


coinhall/images/hall.png
cordt-sei, line 1: Their team mentioned the other image was the proper one for the token. This is more of the "organization" logo.

hall

PR #4823, opened 2024-06-13
Add Quark qcAQLA

AmitPr:
Add's Quark's recently launch qcAQLA derivative to Kujira.


kujira/assetlist.json
```diff
@@ -390,6 +390,31 @@
...
+        },
+        {
+          "denom": "qcAQLA",
+          "exponent": 6
+        }
+      ],
+      "base": "factory/kujira1d9fmsgvu3yvvppvqech2s3dhrqdm5ukngcf82ghcl70edfpwcrlsu6yr0u/urcpt",
+      "name": "qcAQLA",
+      "display": "qcAQLA",
+      "symbol": "qcAQLA",
```
JeremyParish69, line 408: since it's a derivative, please define the `traces`.

AmitPr: Hey, can you give an example of a non-IBC transition that would fit the trace?

Previous Quark assets (and indeed, all derivatives on Kujira) don't have this information.

JeremyParish69: I guess it depends on the type of derivative. If it's a burn receipt, or leveraged version, or an index token, then there's no single 1:1 asset to refer to. But in cases of being transferred version or synthetic version of a single asset, then there should be traces. It's hard to say when I don't know what the relationship of qcAQLA to AQLA is.

AmitPr: It’s a LSD (a la stATOM). I suppose I’d trace back to the original AQLA?

Do you have an example of a non-IBC trace?

AmitPr: Added a trace.

kujira/assetlist.json
```diff
@@ -390,6 +390,44 @@
...
+        {
+          "denom": "qcAQLA",
+          "exponent": 6
+        }
+      ],
+      "traces": [
+        {
+          "type": "liquid-stake",
+          "counterparty": {
+            "base_denom": "factory/kujira1xe0awk5planmtsmjel5xtx2hzhqdw5p8z66yqd/uaqla",
```
JeremyParish69, line 409: I don't see this. Can we register AQLA factory/kujira1xe0awk5planmtsmjel5xtx2hzhqdw5p8z66yqd/uaqla
as well?

AmitPr: It's not my token, but I've added the info.

PR #4829, opened 2024-06-14
Add Minimum Length to Strings in JSON Schemas

JeremyParish69:
Add Minimum Length to Strings in JSON Schemas
to prevent properties with empty strings


pyramation:
LGTM!

PR #4847, opened 2024-06-18
Add DTEAM's Dymension mainnet peer, seed, rest API endpoint, RPC endpoint, explorer to chain.json

doodleoo3:


dymension/chain.json
```diff
@@ -495,6 +521,12 @@
     ]
   },
   "explorers": [
+    {
```
JeremyParish69, line 524: Please insert these additions to the bottom of their respective arrays

doodleoo3: done

PR #4848, opened 2024-06-18
Remove Individual Schema Checks on PR

JeremyParish69:
Remove Individual Schema Checks on PR
assetlist, chain, versions, keys, but NOT ibc, because that has additional checks


pyramation:
LGTM!

PR #4851, opened 2024-06-19
add yum

natecalc:


axelar/assetlist.json
```diff
@@ -1380,6 +1366,50 @@
...
+      "type_asset": "ics20",
+      "base": "ibc/507BE7E33F06026652F519AD4D36716251F2D34DF04514A905D3B19A7D8130F7",
+      "name": "Axelar Wrapped YUM",
+      "display": "yum.axl",
+      "symbol": "YUM.axl",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "axelar",
```
JeremyParish69, line 1391: this trace doesn't make sense. It cant be IBC transferred to the same chain it came from. Where does it originate?

natecalc: Hey @JeremyParish69 updated as originally from Arbitrum. Let me know if this is all good. Thanks (and sorry ive been travelling the past few days)

axelar/assetlist.json
```diff
@@ -1380,6 +1366,50 @@
...
+      "type_asset": "ics20",
+      "base": "ibc/507BE7E33F06026652F519AD4D36716251F2D34DF04514A905D3B19A7D8130F7",
+      "name": "Axelar Wrapped YUM",
+      "display": "yum.axl",
+      "symbol": "YUM.axl",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "arbitrum",
```
JeremyParish69, line 1391: is there an IBC connection between axelar and arbitrum? I didn't know arbitrum had IBC... If this exists, it would need to be registered under /_IBC/

axelar/assetlist.json
JeremyParish69: The asset should then be registered in Arbitrum's assetlist and the image uploaded to /arbitrum/images/. Not sure why it's added to /ethereum/images/...

axelar/assetlist.json
natecalc: Hey @JeremyParish69 sorry about this (first time doing it). So, our token is on Arbitrum, but also we bridged it with axelar. So we actualkly have YUM (on arb) , and YUM.axl (on axelar). 

Any guidance on which one to use?

PR #4854, opened 2024-06-19
Add stISLM token Stride assetlist and IBC connection

VorobevSA:


stride/assetlist.json
```diff
@@ -598,6 +598,42 @@
...
+          "provider": "Stride"
+        }
+      ],
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/stride/images/islm.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/stride/images/islm.svg"
+      },
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/stride/images/islm.png",
```
JeremyParish69, line 633: This would be the wrong image to use. The islm image already exists under haqq, so we don't need the duplication. But also the stISLM has a different image

PR #4855, opened 2024-06-19
Update chain.json - testnet inj gas should be identical to mainnet inh

PFC-developer:
testnet inj gas settings should be identical to mainnet


achilleas-kal:
Confirming the change. Thank you @PFC-developer !

PR #4871, opened 2024-06-21
feat: change status of blockspacerace to killed

jcstein:
replaces #4220


JeremyParish69:
Please see this comment on how to keep 'celestiatestnet' be the primary testnet
https://github.com/cosmos/chain-registry/pull/4220#issuecomment-2021506182

PR #4872, opened 2024-06-22
adding bluechip chain to registry

Bluechip23:
New push for bluechip with updates via recommendations from first push.


bluechip/chain.json
```diff
@@ -0,0 +1,74 @@
...
+          "id": "3x4pynlp86prhcmtns742kgsgu7pjtzjpk8d93",
+          "address": "http://67.28.8.88:26659/",
+          "provider": "CryptoDungeon"
+        }
+      ]
+    },
+    "images": [
+      {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/bluechip/images/bluechip.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/bluechip/images/bluechip.svg"
```
JeremyParish69, line 71: I think the svg might be invalid

bluechip/chain.json
```diff
@@ -0,0 +1,74 @@
+{
+    "$schema": "../chain.schema.json",
+    "chain_name": "bluechip",
+    "status": "live",
+    "website": "https://www.bluechip.link",
+    "network_type": "mainnet",
+    "pretty_name": "BlueChip",
+    "chain_id": "bluechip_1",
+    "bech32_prefix": "bcp",
```
JeremyParish69, line 9: Pending registration: https://github.com/satoshilabs/slips/blob/master/slip-0173.md

Bluechip23: I have made a pull request on the registry

JeremyParish69: I see it registered as "bluechip", not as "bcp". Please make sure they are both the same and match what the chain uses.

Bluechip23: I misread the column. I have updated my information on SLIP table.

bluechip/chain.json
JeremyParish69: Please remove the svg. It looks low quality and unlike the png.

JeremyParish69: The svg is not scalable vectors, it's just an embedded image. Please remove unless there is an actual scalable vector file.

Bluechip23: I have deleted it. I will have to find a resource to help make an svg file. Is this something required for initial submittal?

JeremyParish69: I think the last thing is to remove iamges[0].svg from this chain file and from the assetlist.

bluechip/assetlist.json
```diff
@@ -0,0 +1,28 @@
...
+            "exponent": 6
+          }
+        ],
+        "base": "ubluechip",
+        "name": "blue chip",
+        "display": "bcp",
+        "symbol": "BCP",
+        "images": [
+          {
+            "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/bluechip/images/bluechip.png",
```
JeremyParish69, line 23: this comma will need to be removed

PR #4873, opened 2024-06-23
feat: add yum

joshlopes:


JeremyParish69:
Pending registration of YUM at Arbitrum and Axelar
https://github.com/cosmos/chain-registry/pull/4851

joshlopes:
Do we need the registration first ? maybe we can move the registration changes to the other PR and just register this token? So the token that are in the wallets are recognised.

JeremyParish69:
> Do we need the registration first ? maybe we can move the registration changes to the other PR and just register this token? So the token that are in the wallets are recognised.

Yes, this must be registered to the Axelar asset first, and should also be registered at the Arbitrum assetlist as well.

joshlopes:
@JeremyParish69 do you need anything else from me?

kujira/assetlist.json
```diff
@@ -2362,6 +2362,49 @@
...
+        },
+        {
+          "denom": "YUM.axl",
+          "exponent": 18
+        }
+      ],
+      "type_asset": "ics20",
+      "base": "ibc/507BE7E33F06026652F519AD4D36716251F2D34DF04514A905D3B19A7D8130F7",
+      "name": "Axelar Wrapped YUM",
+      "display": "yum",
```
JeremyParish69, line 2380: Error: Display denomination yum is not defined as a denom_unit.
Then denom with exponent 18 is defined as YUM.axl

kujira/assetlist.json
```diff
@@ -2362,6 +2362,49 @@
...
+        {
+          "image_sync": {
+            "chain_name": "ethereum",
+            "base_denom": "0xcE682c89C63d2850Cb2ca898E44D6c7c30d897a6"
+          },
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/ethereum/images/yum.png"
+        }
+      ],
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/ethereum/images/yum.png"
```
JeremyParish69, line 2406: This image has been upload elsewhere, so we no longer need the image file you're adding in this PR

JeremyParish69:
Stale

PR #4880, opened 2024-06-26
feat: add USDY on Noble testnet

johnletey:


testnets/nobletestnet/assetlist.json
```diff
@@ -81,48 +81,48 @@
...
+      "display": "usdy",
+      "name": "Ondo US Dollar Yield",
+      "symbol": "USDY",
       "traces": [
         {
           "type": "test-mintage",
           "counterparty": {
-            "chain_name": "noble",
-            "base_denom": "uusdlr"
+            "chain_name": "nobletestnet",
```
JeremyParish69, line 106: this is a self-referential pointer, which doesn't make sense. It should point to the mainnet version, but if it doesn't exist yet, then just skip the trace.

johnletey:
Thanks for the info @JeremyParish69, updated!

PR #4903, opened 2024-07-02
Add Airchains Testnet Configuration and Asset Information

kritarth1107:
This pull request adds the necessary configuration and asset information for the Airchains Testnet. The following files have been added:

- `testnets/airchainstestnet/assetlist.json`: Contains the asset list for the Airchains Testnet.
- `testnets/airchainstestnet/chain.json`: Contains the chain configuration for the Airchains Testnet.
- `testnets/airchainstestnet/images/airchains-logo.png`: PNG format of the Airchains logo.
- `testnets/airchainstestnet/images/airchains-mono.svg`: Monochrome SVG format of the Airchains logo.
- `testnets/airchainstestnet/images/airchains.png`: Additional PNG format of the Airchains logo.
- `testnets/airchainstestnet/images/airchains.svg`: Additional SVG format of the Airchains logo.


testnets/airchainstestnet/assetlist.json
```diff
@@ -0,0 +1,33 @@
...
+      "name": "Airchains",
+      "display": "amf",
+      "symbol": "AMF",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/airchainstestnet/images/airchains-mono.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/airchainstestnet/images/airchains-mono.svg"
+      },
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/airchainstestnet/images/airchains.png",
```
JeremyParish69, line 27: Please do not add logotypes or combined logos. Only logomarks/icons and token logos.

testnets/airchainstestnet/assetlist.json
```diff
@@ -0,0 +1,33 @@
...
+          "denom": "amf",
+          "exponent": 6
+        }
+      ],
+      "base": "amf",
+      "name": "Airchains",
+      "display": "amf",
+      "symbol": "AMF",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/airchainstestnet/images/airchains-mono.png",
```
JeremyParish69, line 22: this is not the correct file name (-mono). The PNG is -logo. I'd suggest renaming the visually-identical image files to be the same. Then, but sure the include all images in the `images` array. The logo_URIs is an optional legacy property.

testnets/airchainstestnet/assetlist.json
```diff
@@ -0,0 +1,33 @@
...
+  "assets": [
+    {
+      "description": "The native staking token of Airchains Testnet.",
+      "denom_units": [
+        {
+          "denom": "amf",
+          "exponent": 0
+        },
+        {
+          "denom": "amf",
```
JeremyParish69, line 13: Ambiguous denoms, does amf mean 0 exponent or 6 exponent? They must be different.

kritarth1107: Regarding the ambiguous `denom` values:

Since we have this configuration set up in the chain itself, that's why it is written this way. The chain's minimal denom and denom are both set to be `amf`, hence the duplication. If this creates an issue, can we just add the minimal denom here and leave out the normal denom? Would that work for the requirements?

kritarth1107:
Hey @JeremyParish69 
have made the changes based on your feedback:

1. Removed all logotypes and combined logos, ensuring only logomarks/icons and token logos are included.
2. Corrected the file names for consistency:
   - `airchains-logo.svg`
   - `airchains-logo-mono.png`
   - `airchains-logo.svg`
3. Clarified the `denom` values by differentiating between the minimal denom and the normal denom. Since the chain's minimal denom and denom are both set to `amf`, we have retained this configuration. If there are any issues, please let us know if it's acceptable to include only the minimal denomination.

Please review the changes and let us know if further adjustments are needed.

Thank you!

testnets/airchainstestnet/assetlist.json
JeremyParish69: It is problematic to have both be the same, and there is a strict requirement that they be different.
It is quite rare that there is no second denom, so I'd just like to ask confirmation that the asset, when represented by the token symbol AMF will never be divisible on any interface?
In most cases, the spelling of non-base denoms are arbitrary, so if an AMF is meant to be divisible into 1 million units, then I'd recommend adding the exp:6 denom back in, but changing the string to something like "AMF" (upper case).

testnets/airchainstestnet/chain.json
```diff
@@ -0,0 +1,368 @@
...
+        "recommended_version": "v1.0.0",
+        "compatible_versions": [
+          "v1.0.0"
+        ]
+      }
+    ]
+  },
+  "images": [
+    {
+      "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/airchainstestnet/images/airchains.png",
```
JeremyParish69, line 51: there is no file of this name. I assume this should include the "-logo"...

kritarth1107:
Hey @JeremyParish69
implemented the necessary changes according to the feedback you provided
Please review the changes and let me know if further adjustments are needed.

testnets/airchainstestnet/chain.json
kritarth1107: Yes, you are absolutely correct. I forgot to include that information. I have now updated it. Please review it and let me know if it is accurate.

testnets/airchainstestnet/assetlist.json
JeremyParish69: I don't think this will affect transactions, because all commands to a chain generally only use the base denom ('amf'). The display denom (AMF) is just used to display the asset on application interfaces.
However, merely being included in the denom_untis array isn't enough to tell interfaces to use AMF over amf; there's one last change expected here, and that is to change the "display" value to "AMF" (to uppercase, to match the new denom unit).

kritarth1107: Done

PR #4916, opened 2024-07-04
update channel of shido-noble IBC

NodeNavigator:
update IBC channel information between shido-noble

### DIFF
--- a/_IBC/noble-shido.json
+++ b/_IBC/noble-shido.json
@@ -2,22 +2,22 @@
   "$schema": "../ibc_data.schema.json",
   "chain_1": {
     "chain_name": "noble",
-    "client_id": "07-tendermint-105",
-    "connection_id": "connection-97"
+    "client_id": "07-tendermint-106",
+    "connection_id": "connection-99"
   },
   "chain_2": {
     "chain_name": "shido",
-    "client_id": "07-tendermint-8",
-    "connection_id": "connection-6"
+    "client_id": "07-tendermint-9",
+    "connection_id": "connection-11"
   },
   "channels": [
     {
       "chain_1": {
-        "channel_id": "channel-84",
+        "channel_id": "channel-87",
         "port_id": "transfer"
       },
       "chain_2": {
-        "channel_id": "channel-4",
+        "channel_id": "channel-5",
         "port_id": "transfer"
       },
       "ordering": "unordered",


JeremyParish69:
Channels cannot be updated once registered--they should be revived.
Is there proof that no tokens are left abandoned on the counterparties via the current/old channels?

NodeNavigator:
@JeremyParish69  Actually, the previous noble client expired, so we created a new client for it. There are no funds in the client we just created; it was just for testing purposes. This updated client is the actual one, so please merge it into the main. If anything else is required, let me know. Thanks.

JeremyParish69:
> @JeremyParish69 Actually, the previous noble client expired, so we created a new client for it. There are no funds in the client we just created; it was just for testing purposes. This updated client is the actual one, so please merge it into the main. If anything else is required, let me know. Thanks.

Hi, yes I ask for evidence that shows that no funds were transferred in the previous connection.

NodeNavigator:
> > @JeremyParish69 Actually, the previous noble client expired, so we created a new client for it. There are no funds in the client we just created; it was just for testing purposes. This updated client is the actual one, so please merge it into the main. If anything else is required, let me know. Thanks.
> 
> Hi, yes I ask for evidence that shows that no funds were transferred in the previous connection.

@JeremyParish69  Actually, there is only 11 USD (11000000) there, but the client account has expired, so I'm not able to transfer those funds. Therefore, I don't know how I can provide proof since the client account has expired. But here is the swagger link of that shido noble IBC token. For reference you can see this. 

https://swagger.shidoscan.com/cosmos/bank/v1beta1/supply/by_denom?denom=ibc/BFF0D3805B50D93E2FA5C0B2DDF7E0B30A631076CD80BC12A48C0E95404B4A41

JeremyParish69:
> @JeremyParish69 Actually, there is only 11 USD (11000000) there, but the client account has expired, so I'm not able to transfer those funds. Therefore, I don't know how I can provide proof since the client account has expired. But here is the swagger link of that shido noble IBC token. For reference you can see this.
> 
> https://swagger.shidoscan.com/cosmos/bank/v1beta1/supply/by_denom?denom=ibc/BFF0D3805B50D93E2FA5C0B2DDF7E0B30A631076CD80BC12A48C0E95404B4A41

Thank you this is helpful info.
In this case it would be most standard to require that the client be reenabled via governance, or else have governance on both chains vote to allow the registered connection to be changes. Granted, 11 USD is not too much, but a testing amount should be less than 1$.

NodeNavigator:
> > @JeremyParish69 Actually, there is only 11 USD (11000000) there, but the client account has expired, so I'm not able to transfer those funds. Therefore, I don't know how I can provide proof since the client account has expired. But here is the swagger link of that shido noble IBC token. For reference you can see this.
> > https://swagger.shidoscan.com/cosmos/bank/v1beta1/supply/by_denom?denom=ibc/BFF0D3805B50D93E2FA5C0B2DDF7E0B30A631076CD80BC12A48C0E95404B4A41
> 
> Thank you this is helpful info. In this case it would be most standard to require that the client be reenabled via governance, or else have governance on both chains vote to allow the registered connection to be changes. Granted, 11 USD is not too much, but a testing amount should be less than 1$.

@JeremyParish69  Can you merge this request? Actually, from now on, all real transactions will happen in this channel.

PR #4924, opened 2024-07-08
Update archway osmosis testnet IBC channels with active channels

isakfm:
- Update connections
- Update channels
- Update port id
- Replace wrong base_denom aarch with aconst

Reason: 
The archway testnet channels on cosmos chain registry not active anymore, so I update to "active" ones

### DIFF
--- a/testnets/_IBC/archwaytestnet-osmosistestnet.json
+++ b/testnets/_IBC/archwaytestnet-osmosistestnet.json
@@ -2,26 +2,56 @@
   "$schema": "../ibc_data.schema.json",
   "chain_1": {
     "chain_name": "archwaytestnet",
-    "client_id": "07-tendermint-77",
-    "connection_id": "connection-73"
+    "client_id": "07-tendermint-121",
+    "connection_id": "connection-120"
   },
   "chain_2": {
     "chain_name": "osmosistestnet",
-    "client_id": "07-tendermint-1195",
-    "connection_id": "connection-1101"
+    "client_id": "07-tendermint-3459",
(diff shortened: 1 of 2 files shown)


JeremyParish69:
Can you please re-add the "unnecessary data" and change it back to aarch (not aconst). It's a singly linked list that is supposed to be referencing mainnet ARCH. We also cannot register the coingecko_id for the testnet asset.

testnets/_IBC/archwaytestnet-osmosistestnet.json
```diff
@@ -2,29 +2,59 @@
...
       },
       "ordering": "unordered",
       "version": "ics20-1",
       "tags": {
         "status": "live"
       }
+    },
+    {
+      "chain_1": {
+        "channel_id": "*",
```
JeremyParish69, line 31: These connections with no property channel IDs or ports are concerning.
I'm not comfortable approving it because I don't understand it

isakfm: Hi @JeremyParish69 The reason of this is we are developing smart contract that will create channel for ICA and ICQ, so the channel id will be dynamic and the port id on archway side will be dynamic also. 
Actually Archway chain team is fine with that.

isakfm:
> Can you please re-add the "unnecessary data" and change it back to aarch (not aconst). It's a singly linked list that is supposed to be referencing mainnet ARCH. We also cannot register the coingecko_id for the testnet asset.

Hi @JeremyParish69 I already reverted it

PR #4925, opened 2024-07-08
Update TTT Endpoint Umee

caotre:


ankurdotb (approved):
Removes unreachable RPC endpoint from Umee

PR #4926, opened 2024-07-08
Remove TTT Endpoint Shentu

caotre:


ankurdotb (approved):
Removes unreachable RPC endpoint from Shentu

PR #4927, opened 2024-07-08
Update TTT Endpoint Cosmoshub

caotre:


ankurdotb (approved):
Removes unreachable RPC endpoint on Cosmos Hub

PR #4928, opened 2024-07-08
Remove TTT Endpoint Aura

caotre:


ankurdotb (approved):
Removes unreachable RPC endpoint on Aura

PR #4944, opened 2024-07-09
Add neutaro-osmosis ibc path

Neutaro:
Add neutaro-osmosis ibc path.

### DIFF
--- a/_IBC/neutaro-osmosis.json
+++ b/_IBC/neutaro-osmosis.json
@@ -1,7 +1,7 @@
 {
     "$schema": "../ibc_data.schema.json",
     "chain_1": {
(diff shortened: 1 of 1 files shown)


_IBC/neutaro-osmosis.json
```diff
@@ -0,0 +1,32 @@
+{
+    "$schema": "../ibc_data.schema.json",
+    "chain_1": {
+      "chain_name": "Neutaro",
```
JeremyParish69, line 4: this should have a lowercase 'n'

Neutaro: Hiho,
changed it.

PR #4945, opened 2024-07-09
Update INIT logo

alecande11:


ankurdotb (approved):
I'm not 100% certain whether the slightly different image sizes are okay, but given that the aspect ratio is the same and these PNGs/SVGs, I don't see a big issue (especially since the size is so close to the original values).

PR #4957, opened 2024-07-10
Adds $MAND to osmosis

prtk418:


JeremyParish69:
Please register $MAND [here ](https://github.com/cosmos/chain-registry/blob/master/dymension/assetlist.json)in Dymension's assetlist, similar to how NIM is registered there as an ICS20, too.

PR #4982, opened 2024-07-15
Update Neutron Assetlist add BNB

luisqa-neutron:
add BNB to Neutron chain registry


JeremyParish69:
Can we retain the original 2-space tab formatting please?

PR #4984, opened 2024-07-16
Add non-cosmos chain Schema and Ethereum

JeremyParish69:
Add non-cosmos chain Schema and Ethereum
Removes requirement for chain_id (which might be able to be retained as a requirement... TBD) and removes requirement for bech32_prefix (required for cosmos chains)


sunnya97:
Maybe add a non-EVM chain as well, such as Solana, to make sure there's no issues with this schema being generalized enough

JeremyParish69:
Added a chain_type property instead, so we do not need a separate schema file

PR #5004, opened 2024-07-21
Vidulum gave up on their own chain and became a BZE token

faneaatiku:
- added VDL to beezee asset list
- upgraded BZE to latest version
- removed vidulum as a native chain


faneaatiku:
@Corey-Code as VDL founder please confirm the removal of Vidulum as native chain

PR #5005, opened 2024-07-21
added VDL as token on beezee network and upgraded beezee

faneaatiku:


beezee/assetlist.json
```diff
@@ -32,6 +32,90 @@
           }
         }
       ]
+    },
+    {
+      "description": "Vidulum App Token",
```
JeremyParish69, line 37: Does the VDL on Vidulum network become a legacy-mintage?

faneaatiku: yes. We will remove the entire old vdl network from chain registry in a different PR in the following days.

JeremyParish69: I'd ask that you please update the VDL in the vidulum/assetlist.json in this PR to trace to this new VDL as a "legacy-mintage" and update the image (of vidulum::VDL) to synchronize with this asset (and no longer point to /vidulum/images/vdl.png, svg).

JeremyParish69: Also, when ready to "remove" the old vidulum chain, instead just change its status as "killed", and it will eventually be moved to an archive location, but retained in the registry.

PR #5014, opened 2024-07-22
adding SYN for streamswap

BHIIKTOR:


injective/assetlist.json
```diff
@@ -174,6 +174,34 @@
         }
       ]
     },
+    {
```
JeremyParish69, line 177: Please relocate to the bottom of the list of assets.

injective/assetlist.json
```diff
@@ -174,6 +174,34 @@
...
+      "description": "The most degenerate NFT on Injective. Gravedigger collection for $bINJ. Giving power back to the community.",
```
JeremyParish69, line 178: Want to double check that this is the right description?

BHIIKTOR: the description is correct, thank you

PR #5017, opened 2024-07-22
Add ashSYN to Injective

DBcosmos:
Add ashSYN (SYN burn derivative) to Injective asset list

### DIFF
--- a/zetachain/chain.json
+++ b/zetachain/chain.json
@@ -34,16 +34,16 @@
   },
   "codebase": {
     "git_repo": "https://github.com/zeta-chain/node",
-    "recommended_version": "v14.0.0",
+    "recommended_version": "v18.0.0",
     "compatible_versions": [
-      "v17.0.0"
+      "v18.0.0"
     ],
-    "cosmos_sdk_version": "v0.47.10",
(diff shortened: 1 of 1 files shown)


migaloo/assetlist.json
```diff
@@ -790,6 +790,52 @@
...
+          }
+        }
+      ],
+      "images": [
+        {
+          "image_sync": {
+            "chain_name": "injective",
+            "base_denom": "factory/inj1a6xdezq7a94qwamec6n6cnup02nvewvjtz6h6e/SYN"
+          },
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/injective/images/syn.ash.png"
```
JeremyParish69, line 833: It looks like this asset is SYN, but the image you're pointing to here is syn.ash, which seems incorrect to me. Should just just syn.png?

migaloo/assetlist.json
```diff
@@ -790,6 +790,52 @@
...
+        }
+      ],
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/injective/images/shitmos.png"
```
JeremyParish69, line 837: This is the wrong image

PR #5035, opened 2024-07-23
fix(coingecko-id): Noble USDC

nick134-bit:
missing id on entries


JeremyParish69:
We do not duplicate coin-gecko IDs on IBC assets--only for additional-mintages.

PR #5051, opened 2024-07-25
Add stBAND to Stride

brianbodollo:
Adds info for Stride's `stBAND`.

Let us know if you need anything else!


stride/assetlist.json
```diff
@@ -686,6 +684,45 @@
...
+      ],
+      "base": "stuband",
+      "name": "Stride Staked BAND",
+      "display": "stBAND",
+      "symbol": "stBAND",
+      "traces": [
+        {
+          "type": "liquid-stake",
+          "counterparty": {
+            "chain_name": "laozi-mainnet",
```
JeremyParish69, line 707: Please correct the chain_name to "bandchain"

brianbodollo: good catch! this should be fixed now

PR #5061, opened 2024-07-27
Update assetlist.json for AutoKujira $AUTO token

araiczyk-wbi:
Update assetlist.json for AutoKujira (https://autokujira.app/) $AUTO token


araiczyk-wbi:
my bad, we're not registered in CG yet. We must wait for this to happen or
we can remove that line?

JeremyParish69:
> my bad, we're not registered in CG yet. We must wait for this to happen or we can remove that line?

Just remove the line for now.

PR #5062, opened 2024-07-27
Add $AUTO token logos

araiczyk-wbi:
Add AutoKujira logos


JeremyParish69:
Can I recommend (not required) omitting the SVG? It's not terrible, but it has the classic converted-to-SVG look and doesn't really add value IMO

araiczyk-wbi:
Yeah, I thought both were required.

Issue #5088, opened 2024-07-30
duplicate json key `address` in osmosis asset list


turbocrime:
https://github.com/cosmos/chain-registry/blob/5b38d257b8a5efebacd6a43fdc333f83e3228a4f/osmosis/assetlist.json#L21447

https://github.com/cosmos/chain-registry/blob/5b38d257b8a5efebacd6a43fdc333f83e3228a4f/osmosis/assetlist.json#L21462

PR #5089, opened 2024-07-30
deduplicate osmosis assetlist tron usdt address field

turbocrime:
fixes https://github.com/cosmos/chain-registry/issues/5088

please confirm the correct address field is deleted


turbocrime:
cc @JeremyParish69

PR #5097, opened 2024-07-31
Improve Chain Codebase Objects' Versioning

JeremyParish69:
Improve Chain Codebase Objects' Versioning
added (or improved): language, consensus, cosmwasm, ibc, sdk.
Each can now have a repo, version, and tag
and many have a type (language, consensus, sdk)


pyramation:
wow this is great! Would we keep the old stuff around for migration?

e.g. cosmos_sdk_version for example

PR #5099, opened 2024-07-31
Add Cosmo token to Osmosis

raphaellafar:
Add Cosmo token to Osmosis.


JeremyParish69:
Please use appropriate PR title

PR #5111, opened 2024-08-01
Update sge to v1.7.2

dreamloss:


sge/chain.json
```diff
@@ -187,6 +187,22 @@
           "type": "cometbft",
           "version": "v0.37.5"
         },
+        "next_version_name": "v1.7.2"
+      },
+            {
```
reversesigh, line 192: Fix formatting here please.

PR #5143, opened 2024-08-09
feat: devnet 3 with evm + fix wss for lower envs

cgilbe27:
Adds devnet 3 with evm + fixes wss for lower envs


testnets/nibirudevnet3/assetlist.json
```diff
@@ -0,0 +1,33 @@
...
+        {
+          "denom": "nibi",
+          "exponent": 6
+        }
+      ],
+      "base": "unibi",
+      "name": "Nibiru",
+      "display": "nibi",
+      "symbol": "NIBI",
+      "logo_URIs": {
```
Cordtus, line 21: I think logo_URIs is sort of soft-deprecated

cgilbe27: Should it just be removed?

JeremyParish69: it's fine to keep for now

testnets/nibirudevnet3evm/assetlist.json
```diff
@@ -0,0 +1,33 @@
+{
+  "$schema": "../assetlist.schema.json",
+  "chain_name": "nibirudevnet3evm",
```
JeremyParish69, line 3: This naming doesn't follow current usage at the Chain Registry. Can we use 'nibirudevnet4' instead (it doesn't matter that it doesn't match the chain_id)?

cgilbe27:
@JeremyParish69 addressed

PR #5150, opened 2024-08-12
updated MARS tokens according to migration

linkielink:
- The MARS token was migrated away from `mars-1` to be a native factory token on `neutron-1` and have its voting rights ported over to DAO DAO; this PR will reflect the change of denom and also adjust the token logo to be white (without background color) for the old (MARS.ibc & MARS.old) token


neutron/assetlist.json
```diff
@@ -1242,6 +1242,42 @@
...
+          "exponent": 0,
+          "aliases": ["umars"]
+        },
+        {
+          "denom": "MARS",
+          "exponent": 6
+        }
+      ],
+      "base": "factory/neutron1ndu2wvkrxtane8se2tr48gv7nsm46y5gcqjhux/MARS",
+      "name": "Mars Protocol token",
```
JeremyParish69, line 1260: Because Mars is being replaced, please add a "traces" to the original Mars asset using type: "legacy-mintage", Provider: "Mars Protocol". This will signal that it has canonically been replaced.

osmosis/assetlist.json
```diff
@@ -21632,6 +21578,108 @@
...
+          "denom": "MARS.old",
+          "exponent": 6
+        }
+      ],
+      "type_asset": "ics20",
+      "base": "ibc/573FCD90FACEE750F55A8864EF7D38265F07E5A9273FA0E8DAFD39951332B580",
+      "name": "Mars Hub",
+      "display": "MARS.old",
+      "symbol": "MARS.old",
+      "coingecko_id": "mars-protocol-a7fcbcfb-fd61-4017-92f0-7ee9f9cc6da3",
```
JeremyParish69, line 21600: Please remove the coingecko ID from the Osmosis assetlist

osmosis/assetlist.json
```diff
@@ -21632,6 +21578,108 @@
...
+          "denom": "MARS",
+          "exponent": 6
+        }
+      ],
+      "type_asset": "ics20",
+      "base": "ibc/B67DF59507B3755EEDE0866C449445BD54B4DA82CCEBA89D775E53DC35664255",
+      "name": "Mars Hub",
+      "display": "MARS",
+      "symbol": "MARS",
+      "coingecko_id": "mars-protocol-a7fcbcfb-fd61-4017-92f0-7ee9f9cc6da3",
```
JeremyParish69, line 21651: Please remove the coingecko ID

neutron/assetlist.json
JeremyParish69: actually it would be the mars/assetlist.json::umars asset that needs the legacy-mintage, because that is the one become legacy. This asset should not have traces because it's the new 'canonical' asset.

mars/assetlist.json
```diff
@@ -18,6 +18,16 @@
       "name": "Mars",
       "display": "mars",
       "symbol": "MARS",
+      "traces": [
+        {
+          "type": "legacy-mintage",
```
JeremyParish69, line 23: Still not quite right because of the self-referencing pointer. It's in the right location now, but should be referencing neutron's MARS (to answer the question: this is a legacy mintage of what other token?)

neutron/assetlist.json
```diff
@@ -1242,6 +1242,42 @@
           }
         }
       ]
+    },
+    {
+      "description": "Mars Protocol is a cross-collateralized Money Market Protocol on Neutron and Osmosis.",
```
JeremyParish69, line 1247: Can we also add an `extended_description` here as well? It has become a requirement for 'verified' status on Osmosis Zone (which I assume is desired due to the modification of the osmosis/assetlist.json)

PR #5184, opened 2024-08-19
Upgrade quicksilver to v1.6.3

dreamloss:


JeremyParish69:
> @JeremyParish69 should `cosmos_sdk_version` still be flagged? I'm not familiar with the style I'm seeing here with sdk->type etc.,

I'd say not required, but if you like.

We want to move toward a new structure for representing key components (like sdk, language, ibc, etc.) with a more accommodating structure (including: type, version, repo, tag) instead of just assuming the same type of these components for all chains via the property name (e.g., the property name `ibc_go_version' assumes that IBC can only be implemented in Go, which isn't correct (could be in Rust, etc.)).

At the moment, the legacy properties (like `cosmos_sdk_version`) are still retained where they already exist because apps might be interfacing directly with the CCR Github main branch, and we don't want apps to suddenly break without warning. However, for new version data, I'd say the new model ( "sdk_version": { "type": "cosmos" } ) is preferred, but you may include the legacy property if you like.

Soon, requirements will be added to ensure that the replacement property is used wherever the legacy property might be used.  Aiming for October to completely remove `cosmos_sdk_version`

JeremyParish69 (approved):

PR #5197, opened 2024-08-21
feat: update arabica-11

rootulp:
This Celestia testnet got upgraded to https://github.com/celestiaorg/celestia-app/releases/tag/v2.0.0 recently so I updated the list of versions to reference it. Also the existing explorer link was broken so I replaced it with Celenium.

Tagging @jcstein for review

### DIFF
--- a/testnets/celestiatestnet2/chain.json
+++ b/testnets/celestiatestnet2/chain.json
@@ -26,16 +26,42 @@
   },
   "codebase": {
     "git_repo": "https://github.com/celestiaorg/celestia-app",
-    "recommended_version": "v2.0.0",
-    "compatible_versions": ["v2.0.0", "v2.1.0"],
+    "recommended_version": "v2.1.2",
+    "compatible_versions": [
+      "v2.0.0",
+      "v2.1.0",
+      "v2.1.1",
+      "v2.1.2"
+    ],
     "genesis": {
       "genesis_url": "https://raw.githubusercontent.com/celestiaorg/networks/master/arabica-11/genesis.json"
     },
     "versions": [
       {
-        "name": "v2.0.0",
-        "recommended_version": "v2.0.0",
-        "compatible_versions": ["v2.0.0", "v2.1.0"]
+        "name": "v1",
+        "recommended_version": "v1.14.0",
+        "compatible_versions": [
+          "v1.3.0",
+          "v1.6.0",
+          "v1.7.0",
+          "v1.9.0",
+          "v1.10.0",
+          "v1.10.1",
+          "v1.11.0",
+          "v1.12.0",
+          "v1.13.0",
+          "v1.14.0"
+        ]
+      },
+      {
+        "name": "v2",
+        "recommended_version": "v2.1.2",
+        "compatible_versions": [
+          "v2.0.0",
+          "v2.1.0",
+          "v2.1.1",
+          "v2.1.2"
+        ]
       }
     ]
   },
--- a/testnets/celestiatestnet2/chain.json
+++ b/testnets/celestiatestnet2/chain.json
@@ -26,30 +26,16 @@
   },
   "codebase": {
     "git_repo": "https://github.com/celestiaorg/celestia-app",
-    "recommended_version": "v1.11.0",
-    "compatible_versions": [
-      "v1.3.0",
-      "v1.6.0",
-      "v1.7.0",
-      "v1.9.0",
-      "v1.10.1",
-      "v1.11.0"
-    ],
+    "recommended_version": "v2.0.0",
+    "compatible_versions": [],
     "genesis": {
       "genesis_url": "https://raw.githubusercontent.com/celestiaorg/networks/master/arabica-11/genesis.json"
     },
     "versions": [
       {
-        "name": "v1.11.0",
-        "recommended_version": "v1.11.0",
-        "compatible_versions": [
-          "v1.3.0",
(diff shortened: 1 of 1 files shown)


testnets/celestiatestnet2/chain.json
```diff
@@ -26,30 +26,16 @@
...
-    "compatible_versions": [
-      "v1.3.0",
-      "v1.6.0",
-      "v1.7.0",
-      "v1.9.0",
-      "v1.10.1",
-      "v1.11.0"
-    ],
+    "recommended_version": "v2.0.0",
+    "compatible_versions": [],
```
jcstein, line 30:
```suggestion
    "compatible_versions": ["v2.0.0"],
```

jcstein (approved):
LGTM with one suggestion. One question for @JeremyParish69 and other reviewers, on whether or not we need both `codebase.compatible_versions` and `versions.compatible_versions` ?

testnets/celestiatestnet2/chain.json
```diff
@@ -26,30 +26,16 @@
...
-    "compatible_versions": [
-      "v1.3.0",
-      "v1.6.0",
-      "v1.7.0",
-      "v1.9.0",
-      "v1.10.1",
-      "v1.11.0"
-    ],
+    "recommended_version": "v2.0.0",
+    "compatible_versions": ["v2.1.0"],
```
jcstein, line 30:
```suggestion
    "compatible_versions": ["v2.0.0", "v2.1.0"],
```

testnets/celestiatestnet2/chain.json
```diff
@@ -26,30 +26,16 @@
...
-          "v1.3.0",
-          "v1.6.0",
-          "v1.7.0",
-          "v1.9.0",
-          "v1.10.1",
-          "v1.11.0"
-        ]
+        "name": "v2.0.0",
+        "recommended_version": "v2.0.0",
+        "compatible_versions": ["v2.1.0"]
```
jcstein, line 38:
```suggestion
        "compatible_versions": ["v2.0.0", "v2.1.0"]
```

jcstein: adding these in case it is recommended by other reviewers

JeremyParish69:
> LGTM with one suggestion. One question for @JeremyParish69 and other reviewers, on whether or not we need both `codebase.compatible_versions` and `versions.compatible_versions` ?

Yes, the versions array, if anything, is even more important than the codebase's version data. The version data that shown in the codebase section is just aggregating the latest known data from versions[].
I recommend adding it in both sections.

jcstein:
thank you @JeremyParish69. let's add both suggestions then @rootulp then this should be good to go

rootulp:
Thanks! I think this can merge

jcstein (approved):

testnets/celestiatestnet2/chain.json
```diff
@@ -26,30 +26,16 @@
...
       {
-        "name": "v1.11.0",
-        "recommended_version": "v1.11.0",
-        "compatible_versions": [
-          "v1.3.0",
-          "v1.6.0",
-          "v1.7.0",
-          "v1.9.0",
-          "v1.10.1",
-          "v1.11.0"
```
JeremyParish69, line 51: Note that, within 'versions', it is best to retain the version history (so v2 would be added as a new object below this v1 object). Perhaps I won't require it for a testnet, but this is the intention.

jcstein: Even with a network upgrade? https://blog.celestia.org/lemongrass-celestias-first/

JeremyParish69: oh, is it a hardfork?

jcstein: yes, and will be referring to it as a "network upgrade" going forward (see https://github.com/celestiaorg/docs/issues/1676 for context)

JeremyParish69: that terminology would be confusing here because a network upgrade generally means the version number increases but the genesis file stays the same. Shouldn't this hardfork (or "Network upgrade" as you call it) has a new genesis file too?

jcstein: the genesis file does stay the same. the chain-id also stays the same. you just can't run a consensus node with v1.x.x software, because it is not compatible with v2.x.x software, hence dropping the array of v1.x.x. versions. does that help?

JeremyParish69: I don't know about running a consensus node, but this sounds like a standard version upgrade (and NOT a new network), in which case I think it makes sense to retain the version history. Note that, even though v1 would be included in the version history, only the versions listed under compatible_versions (and the recommended_version) are actually compatible with the current blocks.

jcstein: ahh, that makes a ton of sense. thanks for clarifying @JeremyParish69 !

jcstein: where should V1 be included in version history?

rootulp: Thanks for all the discussion on this! I updated this file with the historical v1.x releases based on how cosmoshub/chain.json did it. Please LMK if anything looks incorrect.

jcstein: LGTM, thank you @rootulp and @JeremyParish69 !

PR #5203, opened 2024-08-22
Update Neutron Version v.4.2.0

pr0n00gler:


reversesigh:
I believe this was a halt upgrade, and for cosmovisor the binary belongs in the v4.0.1 folder. Instead of creating a new version for v4.2.1, it should become the recommended for the v4.0.0 series.

neutron/chain.json
```diff
@@ -248,20 +250,55 @@
...
+          "type": "cometbft",
+          "version": "v0.38.11"
+        },
+        "cosmwasm_version": "neutron-org/wasmd v0.51.2-neutron",
+        "cosmwasm_enabled": true,
+        "ibc_go_version": "v8.2.1",
+        "next_version_name": "",
+        "sdk": {
+          "type": "cosmos",
+          "version": "v0.50.8",
```
JeremyParish69, line 276: missing
`"repo": "https://github.com/neutron-org/cosmos-sdk",`

neutron/chain.json
```diff
@@ -248,20 +250,55 @@
...
+          "tag": "v0.50.8-neutron"
+        },
+        "cosmwasm": {
+          "version": "v0.51.2-neutron",
```
JeremyParish69, line 280:
```
"version": "v0.51.2",
"tag": "v0.51.2-neutron",
```

neutron/chain.json
```diff
@@ -248,20 +250,55 @@
...
+      "version": "v0.50.8",
+      "tag": "v0.50.8-neutron"
     },
     "ibc": {
       "type": "go",
       "version": "v8.2.1"
     },
     "cosmwasm": {
-      "version": "v0.51.0",
+      "version": "v0.51.2-neutron",
```
JeremyParish69, line 301:
```
"version": "v0.51.2",
"tag": "v0.51.2-neutron",
```

neutron/chain.json
```diff
@@ -248,20 +253,58 @@
           "type": "go",
           "version": "v8.2.1"
         }
+      },
+      {
+        "name": "v4.2.1",
```
JeremyParish69, line 258: Seems like there should be a single version object for all v4.2 versions

JeremyParish69:
@reversesigh It seems like the various v4.2 versions could be grouped together, but different versions of components are captured in the more precise version objects for v4.2.x. Should this be forced together into a single object since all v4.2.x versions are all compatible (I believe)?

reversesigh:
> @reversesigh It seems like the various v4.2 versions could be grouped together, but different versions of components are captured in the more precise version objects for v4.2.x. Should this be forced together into a single object since all v4.2.x versions are all compatible (I believe)?

Correct. v4.2.1 should not be its own version, and should be the recommended binary for v4.2.x. v4.2.1 is in the upgrade folder for v4.2.0 in cosmovisor.

neutron/chain.json
pr0n00gler: But it should be applied only at specific height, how can i specify it in a single version object?

JeremyParish69: the block height number to use should be when v4.2.0 started. The versions of components (language, sdk, ibc, cosmwasm, etc.) and the binaries should reflect the latest well-running/recommended version. Since v4.2.0, v4.2.1, and v4.2.2 are all compatible versions, it should be inconsequential to use the earliest version or the earlier version of components.

JeremyParish69:
Appears abandoned and now there are conflicts.

PR #5204, opened 2024-08-22
Upgrade injective to v1.13.2

dreamloss:


injective/chain.json
```diff
@@ -33,13 +33,13 @@
     ]
   },
   "codebase": {
-    "git_repo": "https://github.com/InjectiveLabs/injective-chain-releases",
-    "recommended_version": "v1.13.0-1722157491",
+    "git_repo": "https://github.com/InjectiveFoundation/injective-core/",
+    "recommended_version": "v1.13.2-1723753267",
     "compatible_versions": [
-      "v1.13.0-1722157491"
+      "v1.13.2-1723753267"
```
reversesigh, line 39: Since the source build tag is only v1.13.2 and we are now using the core repo, remove the `-1723753267` from tags (but leave it as part of the binary links)

PR #5208, opened 2024-08-24
Add tucana testnet

garymeso:


testnets/tucanatestnet/chain.json
```diff
@@ -0,0 +1,107 @@
...
+      }
+    ]
+  },
+  "codebase": {
+    "git_repo": "https://github.com/TucanaProtocol/Tucana",
+    "recommended_version": "v8.1.0",
+    "compatible_versions": [
+      "v8.1.0"
+    ],
+    "cosmos_sdk_version": "v0.50.8",
```
JeremyParish69, line 43: this cosmos sdk version number at this higher level disagrees with the number given in the versions array. This should match the latest version in the versions array.
Same issue with ibc go version

garymeso: Thank you for pointing that out, fixed. Pls review again thank you ! @JeremyParish69

PR #5223, opened 2024-08-28
Add dATOM token to Neutron assets list

albertandrejev:


neutron/assetlist.json
```diff
@@ -1279,6 +1279,36 @@
...
+        },
+        {
+          "denom": "dATOM",
+          "exponent": 6
+        }
+      ],
+      "base": "factory/neutron1k6hr0f83e7un2wjf29cspk7j69jrnskk65k3ek2nj9dztrlzpj6q00rtsa/udatom",
+      "name": "dATOM",
+      "display": "dATOM",
+      "symbol": "dATOM",
```
JeremyParish69, line 1299: this is a liquid staking token? Could you please include the `trace` path to ATOM?

JeremyParish69: The trace for IBC-transferred ATOM is actually already defined earlier in the list. I think in this case we're looking for something like: 
```
{
  "type": "liquid-stake",
  "counterparty": {
    "chain_name": "cosmoshub",
    "base_denom": "uatom"
  },
  "provider": "Drop Protocol"
}
```

PR #5230, opened 2024-08-29
Add dATOM and derivatives to Pryzm assets list

madrezaz:
This PR adds dATOM token and the p/y derivatives to the list of PRYZM assets.

Note that this PR uses the dATOM logo introduced in #5223, so it should be merged after that.


pryzm/assetlist.json
```diff
@@ -543,6 +543,52 @@
...
+          "denom": "dATOM",
+          "exponent": 6
+        }
+      ],
+      "type_asset": "ics20",
+      "base": "ibc/EA6E1E8BA2EB9F681C4BD12C8C81A46530A62934F2BD561B120A00F46946CE87",
+      "name": "dATOM",
+      "display": "dATOM",
+      "symbol": "dATOM",
+      "coingecko_id": "drop-staked-atom",
```
JeremyParish69, line 564: Please remove the coingecko ID (it is defined in the neutron assetlist)

PR #5255, opened 2024-09-03
Upgrade axelar to v1.0

dreamloss:


axelar/chain.json
```diff
@@ -34,20 +34,20 @@
...
-    "recommended_version": "v0.35.5",
+    "recommended_version": "v1.0.2",
     "compatible_versions": [
-      "v0.35.5"
+      "v1.0.2"
     ],
     "binaries": {
-      "linux/amd64": "https://github.com/axelarnetwork/axelar-core/releases/download/v0.35.5/axelard-linux-amd64-v0.35.5",
-      "darwin/amd64": "https://github.com/axelarnetwork/axelar-core/releases/download/v0.35.5/axelard-darwin-amd64-v0.35.5"
+      "linux/amd64": "https://axelar-releases.s3.us-east-2.amazonaws.com/axelard/v1.0.2/axelard-linux-amd64-v1.0.2.zip",
```
reversesigh, line 42: The github links should probably be used since the binaries are still available there. Probably best to use alternative sources when github is not available.

https://github.com/axelarnetwork/axelar-core/releases/tag/v1.0.2

reversesigh:
@dreamloss thank you for changing. Since Axelar is offering ARM64 binaries as well, can you add those too?

PR #5281, opened 2024-09-07
chore:update persistent peers and change the IBC version

NodeNavigator:
update persistent peers and change the IBC version

### DIFF
--- a/shido/chain.json
+++ b/shido/chain.json
@@ -37,9 +37,9 @@
   },
   "codebase": {
     "git_repo": "https://github.com/ShidoGlobal/shidochain",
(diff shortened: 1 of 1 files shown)


shido/chain.json
```diff
@@ -57,18 +57,18 @@
     },
     "versions": [
       {
-        "name": "v1.0.0",
-        "tag": "v1.0.0",
-        "recommended_version": "v1.0.0",
+        "name": "v2.0.0",
```
JeremyParish69, line 60: Please retain v1, and add v2 as a new version object later in the `versions` array

NodeNavigator: Done

Issue #4760, opened 2024-06-02
thorchain/assetlist.json not found when import chain-registry


zznam:
Request URL: https://raw.githubusercontent.com/cosmos/chain-registry/master/thorchain/assetlist.json

This route call on our project and it return 404

![image](https://github.com/cosmos/chain-registry/assets/23559800/8c0337f8-50d5-4940-bb4c-91b978442b2f)

JeremyParish69:
Thorchain doesn't have any assetlist.json file.
Althea was missing the $schema property which will be fixed soon.

amiryadid-fb:
@JeremyParish69 shouldn't Thorchain have one?

JeremyParish69:
> @JeremyParish69 shouldn't Thorchain have one?

It would be nice if it did, but nobody has bothered registering it so far

PR #5290, opened 2024-09-09
add thorchain assetlist

doggystylez:
fix problem with dependents as illustrated in #4760


thorchain/assetlist.json
```diff
@@ -0,0 +1,28 @@
...
+          "exponent": 0
+        },
+        {
+          "denom": "RUNE",
+          "exponent": 8
+        }
+      ],
+      "base": "rune",
+      "name": "THORChain RUNE",
+      "display": "rune",
```
JeremyParish69, line 19: I believe the display denom would be 'RUNE', not 'rune'

PR #5303, opened 2024-09-09
update chain_id for thorchain

markfromdenmarkk:


reversesigh:
Was this originally a typo or did THORChain fork to a new chain-id?

reversesigh:
> > Was this originally a typo or did THORChain fork to a new chain-id?
> 
> With the recent hardfork, the chainid was changed to thorchain-1

Got it. If you look over the folders andromeda and andormeda1 in this repo, you can get an idea of how forks are handled.

The old testnet will become a folder named thorchain1, and you'll make a new thorchain folder for the current chain ID (with a field for `"pre_fork_chain_name"` defining `thorchain1` folder as the previous chain.

JeremyParish69:
What is the name of the current version? The old version name is still in there.
I believe this should have a different genesis file, but the genesis URL looks broken

reversesigh:
Hey @markfromdenmarkk apologies for the mix-up here. Since there is no assetlist and not much in terms of chain version history in the original folder - you're actually fine to wipe thorchain1 and remove the pre_fork_name from the current chain.json if you wish. The fork method I mentioned above is really only relevant when that information already exists in the registry. If you do that though, since it's technically a new chain ID, it would be best to start the version history with the initial version used for the fork (and as Jeremy wrote above, the genesis will still need updated too)

reversesigh:
@markfromdenmarkk these changes look good - thank you! The last thing is resolving the codebase and version history which Jeremy also mentioned above. 

If you look at the schema (https://github.com/cosmos/chain-registry/blob/master/chain.schema.json) and look over codebase, there's things we need like the recommended version, compatible versions, repo link (@JeremyParish69 do we support gitlab links?), tag, etc., not all the fields need populated, but some basic ones related to the repo location and version tag at least are very helpful! 

For the version history, just start with the first tag for the new fork. If the current version name in the version history is accurate for the current tag, you can leave that name but just add the information you've added to codebase. You can refer to other chains like Cosmos Hub and look over their chain history/codebase to get an idea of what I mean for everything.

JeremyParish69:
@reversesigh (@JeremyParish69 do we support gitlab links?)
I haven't come across it, so not atm, but shouldn't be too hard to adapt for.

JeremyParish69 (approved):

PR #5318, opened 2024-09-11
Add Vault - The native token of the Underworld

GGkingken:
Vault is the native token of the Underworld used in DAO DAO

Please let me know if there is any issues , thank you!


GGkingken (commented):
Ok fixed it i think

PR #5324, opened 2024-09-12
Add BitBadges

trevormil:
This pull requests adds support for the BitBadges blockchain which just went live.


bitbadges/chain.json
```diff
@@ -0,0 +1,100 @@
...
+      "kind": "ping.pub",
+      "url": "https://explorer.bitbadges.io/BitBadges%20Mainnet",
+      "tx_page": "https://explorer.bitbadges.io/BitBadges%20Mainnet/txs/${txHash}",
+      "account_page": "https://explorer.bitbadges.io/BitBadBitBadges%20Mainnetges/account/${accountAddress}"
+    },
+  ],
+  "images": [
+    {
+      "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/bitbadges/images/bitbadgeslogo.png",
+      "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/bitbadges/images/bitbadgeslogo.svg"
```
JeremyParish69, line 97: the png and the svg are not visually idendical, and therefore cannot be placed inside the image object. You may have two image objects in the array.

bitbadges/assetlist.json
```diff
@@ -0,0 +1,70 @@
...
+      ],
+      "base": "stake", 
+      "name": "BitBadges Staking",
+      "display": "stake",
+      "symbol": "STAKE",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/bitbadges/images/bitbadgeslogo.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/bitbadges/images/bitbadgeslogo.svg"
+      },
+      "coingecko_id": "",
```
JeremyParish69, line 57: Please remove fields with empty strings

bitbadges/chain.json
```diff
@@ -0,0 +1,100 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "bitbadges",
+  "status": "live",
+  "network_type": "mainnet",
+  "website": "https://bitbadges.io/",
+  "pretty_name": "BitBadges",
+  "chain_id": "bitbadges-1",
+  "bech32_prefix": "cosmos",
```
JeremyParish69, line 9: The bech32 prefix is taken.... I think it's not allowed to be the same.

bitbadges/chain.json
```diff
@@ -0,0 +1,100 @@
...
+      }
+    ]
+  },
+  "codebase": {
+    "git_repo": "https://github.com/bitbadges/bitbadgeschain/",
+    "recommended_version": "v1.0-mainnet",
+    "compatible_versions": [
+      "v1.0-mainnet"
+    ],
+    "versions": ["v1.0-mainnet"],
```
JeremyParish69, line 37: The versions array contains objects, not strings. Please redefine v1.0-mainnet as a version object

trevormil:
I resolved the previous comments made except for the bech32 Cosmos prefix one.

I actually didn't realize the unique bech32 prefix was a requirement, so I will have to work towards changing it in the next chain update. Would you want me to remove the field / leave empty and edit when updated? Keep it as "cosmos" for now? Or, should I just revisit this PR when the fix goes live?

bitbadges/chain.json
JeremyParish69: This was not resolved

bitbadges/assetlist.json
```diff
@@ -0,0 +1,73 @@
...
+      ],
+      "base": "badge",
+      "name": "BitBadges Credits",
+      "display": "badge",
+      "symbol": "BADGE",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/bitbadges/images/bitbadgeslogo.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/bitbadges/images/bitbadgeslogo.svg"
+      },
+      "coingecko_id": "",
```
JeremyParish69, line 25: Please remove fields with empty strings

bitbadges/chain.json
```diff
@@ -0,0 +1,114 @@
...
+    "compatible_versions": [
+      "v1.0-mainnet"
+    ],
+    "versions": [{
+      "name": "v1.0-mainnet",
+      "binaries": {
+        "linux/amd64": "https://github.com/BitBadges/bitbadgeschain/releases/download/v1.0-mainnet/bitbadgeschain-linux-amd64",
+        "linux/arm64": "https://github.com/BitBadges/bitbadgeschain/releases/download/v1.0-mainnet/bitbadgeschain-linux-arm64"
+      },
+      "cosmos_sdk_version": "v0.50.8",
```
JeremyParish69, line 44: the "sdk" and "ibc" properties are also required. They describe the same version data.

bitbadges/chain.json
```diff
@@ -0,0 +1,114 @@
...
+        "type": "cometbft",
+        "version": "github.com/cometbft/cometbft v0.38.9"
+      }
+    }],
+    "binaries": {
+      "linux/amd64": "https://github.com/BitBadges/bitbadgeschain/releases/download/v1.0-mainnet/bitbadgeschain-linux-amd64",
+      "linux/arm64": "https://github.com/BitBadges/bitbadgeschain/releases/download/v1.0-mainnet/bitbadgeschain-linux-arm64"
+    },
+    "cosmos_sdk_version": "v0.50.8",
+    "ibc_go_version": "v8.3.2",
```
JeremyParish69, line 56: the "sdk" and "ibc" properties are also required. They describe the same version data.

bitbadges/chain.json
JeremyParish69: I recommend using the version without text, is it will not be readable >99% of the time.

JeremyParish69:
> @JeremyParish69
> 
> I resolved the previous comments made except for the bech32 Cosmos prefix one.
> 
> I actually didn't realize the unique bech32 prefix was a requirement, so I will have to work towards changing it in the next chain update. Would you want me to remove the field / leave empty and edit when updated? Keep it as "cosmos" for now? Or, should I just revisit this PR when the fix goes live?

I suppose we can proceed for now

bitbadges/chain.json
JeremyParish69: there are still several instances of bitbadges svg and png in the same object. If they are going to remain appearing different, then they should be given different names, and both should be registered as an images for something (either the chain or an asset) at least once.

trevormil: @JeremyParish69 Do I need an SVG, or am I okay to just have the PNG? The PNG is the main official logo. 

In the most recent version, I removed all SVG references.

JeremyParish69: Generally an SVG is better than PNG because it is scalable.
Also, I find it odd that this is perhaps the only token logo with text in it--that is highly unusual because it will not be legible in most UIs. I thought the SVG without text would be more appropriate....
Regardless, I suppose we can proceed as is.

trevormil: Yea, you are right. Thinking about it now, that is a little unconventional, but the logo with the text is kinda how we have been displaying it everywhere.

PR #5332, opened 2024-09-13
Add Stargaze VLT vault.png new image

GGkingken:
The previous image size was 250x250 - i changed it to 500x500 

Let me know if this is ok to change , thanks!


JeremyParish69:
This is just above 200 KB. It's best to try to keep it smaller than that. I would instead recommend focusing on removing the padding so the visuals take up the entire canvas.

GGkingken:
Changed the padding

JeremyParish69:
Please resolve conflicts

PR #5333, opened 2024-09-13
Update v2 upgrade height for Celestia

reversesigh:


reversesigh:
@rootulp @jcstein 

Tagging you both - this is what you'd like me and others to do for any Celestia changes, correct? Merged your previous PR but went ahead and added a few changes for review and to sorta dry run how you'd like us to handle PRs.

jcstein:
yes this works for us! thank you for tagging us :) feel free to request reviews too if it helps

celestia/chain.json
```diff
@@ -83,13 +83,15 @@
...
+        "height": 2371495,
         "recommended_version": "v2.1.2",
         "compatible_versions": [
           "v2.0.0",
           "v2.1.0",
           "v2.1.1",
           "v2.1.2"
-        ]
+        ],
+        "next_version_name": ""
```
jcstein, line 94: lgtm as long as this can be empty. it will be v3, but i think we can make that change later. wdyt @rootulp ?

jcstein (approved):

rootulp (approved):

celestia/chain.json
reversesigh: @jcstein we tend to set a dummy blank "next_version_name", so if someone goes to update they might catch that it needs filled! this is good to go! Thanks for catching and inquiring though!

jcstein: anytime! thanks for following up 😄

PR #5335, opened 2024-09-13
bluechip-1 to bluechip-2

Bluechip23:
We originally had changed out chain-Id but after doing so, we ran into multiple issues with syncing to other nodes and IBC issues. We upgraded our chain at created a new one. Everything else about the chain is the same.

### DIFF
--- a/bluechip/chain.json
+++ b/bluechip/chain.json
@@ -55,7 +55,7 @@
         "cosmwasm_enabled": true,
         "sdk": {
           "type": "cosmos",
-          "version": "0.46.7"
+          "version": "0.47.11"
         },
         "cosmwasm": {
           "enabled": true
@@ -64,7 +64,7 @@
     ],
     "sdk": {
       "type": "cosmos",
-      "version": "0.46.7"
+      "version": "0.47.11"
     },
     "cosmwasm": {
       "enabled": true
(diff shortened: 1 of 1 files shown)


Bluechip23:
@JeremyParish69 To be clear, we have launched a new chain due to errors we were having with IBC and peer to peer communication issues after chainid change. I pushed new data to this chain becuase I am not 100% sure how the registry works. If I need to delete this chain from registry completely and create new push I am happy to do this. The new chain is the exact same outside the changes I have made to this push. Sorry for any inconvenience

JeremyParish69:
Normally we would have the old version (bluechip-1) be archived and the new version would take its place in this directory, but since there doesn't seem to be much integration and no ibc connections registered, we may be able to update a few values in place as you have.
If this is a new chain, does it need a genesis with addresses and balances?

Bluechip23:
I am not entirely sure what you are asking. This is a new chain that has a new genesis file of addresses and balances. 
The link to the genesis is the same link, and I updated the validators address in the push to make sure it was no longer the old one.

JeremyParish69:
> I am not entirely sure what you are asking. This is a new chain that has a new genesis file of addresses and balances. The link to the genesis is the same link, and I updated the validators address in the push to make sure it was no longer the old one.

I was thinking that the link to the genesis file must be different and should have at least hundreds of balances in it, but I see it was updated 2 weeks ago, and I guess there just hasn't been any usage of the old chain at all because I see only 2 balance in now.

PR #5338, opened 2024-09-16
update injectice Drugs json and Images

DBcosmos:


injective/assetlist.json
```diff
@@ -1068,6 +1068,31 @@
...
+    },
+    {
+      "description": "Distributing happiness, is a serious business",
+      "denom_units": [
+        {
+          "denom": "factory/inj1sklcy2px26jj73ffs2f7fmxw77zsts66prrqxr/DRUGS",
+          "exponent": 0
+        },
+        {
+          "denom": "ashSYN",
```
JeremyParish69, line 1091: What's with this denom? "ashSYN" doesn't seem to match with "DRUGS".

injective/assetlist.json
```diff
@@ -1068,6 +1068,31 @@
       "type_asset": "cw20",
       "address": "inj19vy83ne9tzta2yqynj8yg7dq9ghca6yqn9hyej",
       "base": "cw20:inj19vy83ne9tzta2yqynj8yg7dq9ghca6yqn9hyej",
+      "name": "cw20 DRUGS",
+      "display": "DRUGS",
+      "symbol": "DRUGS",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/injective/images/cw20.drugs.png"
+      },
+      "images": [
```
JeremyParish69, line 1077: Please remove the jpg, those are not accepted at the Chain Registry

injective/assetlist.json
```diff
@@ -1068,6 +1068,31 @@
       "type_asset": "cw20",
       "address": "inj19vy83ne9tzta2yqynj8yg7dq9ghca6yqn9hyej",
       "base": "cw20:inj19vy83ne9tzta2yqynj8yg7dq9ghca6yqn9hyej",
+      "name": "cw20 DRUGS",
+      "display": "DRUGS",
+      "symbol": "DRUGS",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/injective/images/cw20.drugs.png"
```
JeremyParish69, line 1075: I'll request that the (.) dot not be used in the file name, because it looks like drugs is the extension, rather than part of a given name.

injective/assetlist.json
```diff
@@ -1068,6 +1068,31 @@
       "type_asset": "cw20",
       "address": "inj19vy83ne9tzta2yqynj8yg7dq9ghca6yqn9hyej",
       "base": "cw20:inj19vy83ne9tzta2yqynj8yg7dq9ghca6yqn9hyej",
+      "name": "cw20 DRUGS",
```
JeremyParish69, line 1071: Is this a legacy mintage of DRUGS? I.e., is this created by the same team, but being replace with this new one?

DBcosmos: Yes kind of. The old cw20 dev is stepping aside and a new team is taking over and minting the factory version. cw20 holders can swap for the new token 1:1 :)

JeremyParish69: In that case I would that the old 'cw20' version be indicated as a legacy mintage using traces: [ { type: legacy-mintage} ]

PR #5349, opened 2024-09-18
Deleting everything

GGkingken:
Deleting it all , to add it again to resolve the conflict if that is possible , thanks


JeremyParish69:
Please fix the change so checks pass, and also update the PR title to be more precise.
Also, what is the context behind this? Generally, we don't delete content but let stuff be archived over time, although this being so recently added might be OK to delete depending on the reason.

GGkingken:
All i want is to change the image to another one, but it seems i have to
study ancient greek while jumping through several hoops while being on fire
and trying to balance a stick with a plate on top of my head to do this ,
you can ignore this pull request , i will just live with the smaller image

Op wo 18 sep 2024 om 19:44 schreef JeremyParish69 ***@***.***

PR #5358, opened 2024-09-19
WIP: Move Asset to a Separate Schema File

JeremyParish69:
Move Asset to a Separate Schema File
Various Assetlists will reference asset but will different its own requirements, depending on the type of assetlist.


JeremyParish69:
Pending the permission of the new schema in the referenced validation package(s).

PR #5374, opened 2024-09-25
add pepe token & hmstr token in oraichain

haunv3:


oraichain/assetlist.json
```diff
@@ -718,6 +716,36 @@
...
+        },
+        {
+          "denom": "pepe",
+          "exponent": 6
+        }
+      ],
+      "base": "factory/orai1wuvhex9xqs3r539mvc6mtm7n20fcj3qr2m0y9khx6n5vtlngfzes3k0rq9/extPEPE",
+      "name": "Pepe",
+      "display": "pepe",
+      "symbol": "PEPE",
```
JeremyParish69, line 735: How did Ethereum PEPE get here? Is it bridged or a synthetic?

oraichain/assetlist.json
```diff
@@ -718,6 +716,36 @@
...
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/ethereum/images/pepe.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/ethereum/images/pepe.svg"
+      },
+      "images": [
+        {
```
JeremyParish69, line 741: Please use image_sync, referring back to the ethereum pepe

oraichain/assetlist.json
haunv3: They are bridged over oraibridge (obridge-sdk) which i have not registered on chain-registry.

JeremyParish69: I'm thinking it might be as simple as adding a trace to pepe, similar to what was added for Oraichain's TON:
https://github.com/oraichain/chain-registry-fork/blob/feat/add-pepe-oraichain/oraichain/assetlist.json#L686-L693

JeremyParish69 (changes requested):
The noted issues were not resolved

PR #5377, opened 2024-09-25
update current version of pion-1 and deps

sotnikov-s:


testnets/neutrontestnet/chain.json
```diff
@@ -25,16 +25,16 @@
   },
   "codebase": {
     "git_repo": "https://github.com/neutron-org/neutron",
-    "recommended_version": "v4.0.0-rc3",
+    "recommended_version": "v4.2.2-testnet",
```
JeremyParish69, line 28: Be sure to update the `versions` array

testnets/neutrontestnet/chain.json
```diff
@@ -25,16 +25,16 @@
...
+      "v4.2.2-testnet"
     ],
     "cosmos_sdk_version": "0.50",
     "consensus": {
       "type": "cometbft",
-      "version": "0.38.7"
+      "version": "0.38.11"
     },
-    "cosmwasm_version": "0.52",
+    "cosmwasm_version": "0.51",
```
JeremyParish69, line 37: Will need to update the `cosmwasm` property as well

testnets/neutrontestnet/chain.json
sotnikov-s: done

testnets/neutrontestnet/chain.json
sotnikov-s: done as well. please take another look 🙏  it must be all

PR #5382, opened 2024-09-26
add explorer & rpc & api for blockx network

nodeiistt:
add explorer & rpc & api for blockx network


JeremyParish69:
Can you fix the formatting please

PR #5398, opened 2024-09-29
Adding CryptoGopniks tokens to neutron and stargaze

MedvedCrypto:


neutron/assetlist.json
```diff
@@ -1354,6 +1354,37 @@
...
+    "display": "GOP",
+    "symbol": "GOP",
+    "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/gop.png"
+    },
+    "images": [
+        {
+            "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/gop.png",
+            "theme": {
+                "primary_color_hex": "#808080"
```
JeremyParish69, line 1384: Shouldn't this be more of a black color rather than a medium grey?

stargaze/assetlist.json
```diff
@@ -566,6 +566,56 @@
           "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/stargaze/images/vault.png"
         }
       ]
+    },
+    {
+      "description": "Liquid token STARS from NFT lending platform GopLend",
```
JeremyParish69, line 571: Sorry, I'm not quite understanding how this is derived. This is a liquid form of what position, exactly?

MedvedCrypto: Thanks for your attention, yes, this is a liquid form of lended STARS, i.e. like stSTARS from Stride (just not for staking)

This token is issued in exchange for the $STARS token when a user makes a deposit on the platform

MedvedCrypto:
I've made a small edit to the description, it should be a little clearer now

stargaze/assetlist.json
JeremyParish69: Thanks for explaining. I was wondering if this aught to describe how it derives from STARS using traces[], but there is currently no trace type to describe liquid-lending (I presume this works kinda like bonds).

PR #5405, opened 2024-10-01
Upgrade Carbon to v2.51.1

tehkohrz:


carbon/chain.json
```diff
@@ -699,6 +699,21 @@
           "linux/amd64": "https://github.com/Switcheo/carbon-bootstrap/releases/download/v2.50.0/carbond-mainnet.linux-amd64.tar.gz",
           "linux/arm64": "https://github.com/Switcheo/carbon-bootstrap/releases/download/v2.50.0/carbond-mainnet.linux-arm64.tar.gz"
         },
+        "next_version_name": "v2.51.0"
+      },
+      {
+        "name": "v2.51.0",
```
JeremyParish69, line 705: You'll also need to update the version described outside the versions array. It's still set to v2.48

PR #5427, opened 2024-10-05
Add CI action to test endpoints from whitelisted chains and providers

pablin-10:
This PR adds a github action to do a simple smoke test of the RPC and LCD endpoints offered in the chain registry.

This is how it looks like:
https://github.com/pablin-10/chain-registry/actions/runs/11265464331

image
image

By now only tests providers whitelisted in the test itself:

```
# Set this to False to ignore the whitelist and process all providers
use_whitelist = True

# Whitelist for specific chains and providers
whitelist = {
    "chains": [
        "axelar",
        "celestia",
        "composable",
        "cosmoshub",
        "dydx",
        "dymension",
        "evmos",
        "injective",
        "neutron",
        "noble",
        "osmosis",
        "stargaze",
        "stride"
    ],
    "providers": [
        "Osmosis Foundation", 
        "Polkachu",
        "CryptoCrew",
        "forbole",
        "Imperator.co",
        "WhisperNode 🤐",
        "chainlayer",
        "Numia",
        "Enigma",
        "kjnodes",
        "Stake&Relax 🦥",
        "Allnodes ⚡️ Nodes & Staking",
        "Lava",
        "Golden Ratio Staking",
        "Stargaze Foundation",
    ]
} if use_whitelist else {'chains': [], 'providers': []}
```

We may configure it to run every day or every week since it may take a lot of time to run for all the chains and providers if whitelist disabled.


_tests/test_endpoints.py
```diff
@@ -0,0 +1,92 @@
...
+
+def generate_endpoint_tests():
+    test_cases = []
+    logging.info(f"Current working directory: {os.getcwd()}")
+    files_found = glob.glob('**/chain.json', recursive=True)
+    for filename in files_found:
+        with open(filename) as f:
+            data = json.load(f)
+            chain_name = data.get('chain_name', 'unknown')
+            if 'apis' in data:
```
p0mvn, line 68: Might be helpful to print some warnings if this condition is false in case the structure changes and we start skipping the entries.

Applies to other if statements below too

_tests/test_endpoints.py
```diff
@@ -0,0 +1,92 @@
...
+                            address = api['address']
+                            if api_type == 'rpc':
+                                address += '/status'
+                            elif api_type == 'rest':
+                                address += '/cosmos/base/tendermint/v1beta1/syncing'
+                            test_cases.append(EndpointTest(chain=chain_name, endpoint=api_type, provider=api['provider'], address=address))
+    return test_cases
+
+test_cases = generate_endpoint_tests()
+if test_cases:
```
p0mvn, line 81: Why do we need this check? Is it ever expected to not pass?

_tests/test_endpoints.py
```diff
@@ -0,0 +1,92 @@
...
+        "cosmoshub",
+        "dydx",
+        "dymension",
+        "evmos",
+        "injective",
+        "neutron",
+        "noble",
+        "osmosis",
+        "stargaze",
+        "stride"
```
p0mvn, line 33: If we want to ensure that our CI passes only when all providers are processed, we could also keep a tally of the number of successful chains validated, checking that it matches the length of the whitelisted chains at the end.

p0mvn (approved):
Nice work! I only had minor defense-in-depth recommendations but overall LGTM!

_tests/test_endpoints.py
pablin-10: Added

p0mvn (approved):
Visualization is pretty sweet

PR #5431, opened 2024-10-08
Update/pepe hmstr oraichain

haunv3:


oraichain/assetlist.json
```diff
@@ -720,8 +718,100 @@
...
+          "counterparty": {
+            "chain_name": "ton",
+            "base_denom": "nanoHmstr"
+          },
+          "provider": "Oraichain Labs Ton Bridge"
+        }
+      ],
+      "base": "factory/orai1wuvhex9xqs3r539mvc6mtm7n20fcj3qr2m0y9khx6n5vtlngfzes3k0rq9/HMSTR",
+      "name": "HMSTR",
+      "display": "hmstr",
```
JeremyParish69, line 793: this denom (case sensitive) isn't defined in denom_units. Either update this to "HMSTR" to make the defined unit, or change the exponent:9 denom unit to "hmstr" so it matches this.

oraichain/assetlist.json
```diff
@@ -720,8 +718,100 @@
...
+      "symbol": "HMSTR",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/oraichain/images/hmstr.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/oraichain/images/hmstr.svg"
+      },
+      "images": [
+        {
+          "image_sync": {
+            "chain_name": "ton",
+            "base_denom": "nanoHmstr"
```
JeremyParish69, line 803: This asset isn't defined, so there's nowhere to pull the image from. Can we register this asset as well under TON's assetlist?

haunv3: This is an asset on the TON network. You can register it. 👍

oraichain/assetlist.json
```diff
@@ -720,8 +718,100 @@
...
+          },
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/oraichain/images/hmstr.png",
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/oraichain/images/hmstr.svg",
```
JeremyParish69, line 806: This SVG file size is way too large. Please only include files <250kB.

JeremyParish69 (changes requested):
Display denom is not correct. Image file size is too large. Referenced asset is not registered.

JeremyParish69:
Appears abandoned

PR #5435, opened 2024-10-09
[nibiru] Add liquid staked NIBI for Nibiru

Unique-Divine:
- Closes https://github.com/NibiruChain/nibiru/issues/2067


nibiru/assetlist.json
```diff
@@ -25,7 +25,7 @@
       ],
       "base": "unibi",
       "name": "Nibiru",
-      "display": "nibi",
+      "display": "NIBI",
```
JeremyParish69, line 28: This unit is not defined and will not pass tests. It was fine before, and I assume you mistake what the 'display' property means. "Symbol" is shown to users, while "display" is only used to determine the number decimal points for the asset amounts.

nibiru/assetlist.json
```diff
@@ -43,6 +43,42 @@
...
+        },
+        {
+          "denom": "stNIBI",
+          "exponent": 6
+        }
+      ],
+      "base": "tf/nibi1udqqx30cw8nwjxtl4l28ym9hhrp933zlq8dqxfjzcdhvl8y24zcqpzmh8m/ampNIBI",
+      "name": "Liquid Staked Nibiru (Eris)",
+      "display": "stNIBI",
+      "symbol": "stNIBI",
```
JeremyParish69, line 66: Please define the trace for this, showing how it relates to NIBI (using trace type: liquid-stake)

Unique-Divine: Updated in 5f8a4e8ee8cfceca0a8d914b5623e3876f0175ff

nibiru/assetlist.json
Unique-Divine: I see, so the display has to match one of the denom units, while symbol is the actual string that people see in most wallets

Fixed in 5f8a4e8ee8cfceca0a8d914b5623e3876f0175ff

PR #5445, opened 2024-10-11
Set agoric to version agoric-upgrade-17

gibson042:
https://ping.pub/agoric/gov/78

### DIFF
--- a/agoric/versions.json
+++ b/agoric/versions.json
@@ -37,6 +37,9 @@
         "tag": "v0.34.30-alpha.agoric.1"
       },
       "cosmwasm_enabled": false,
+      "cosmwasm": {
+        "enabled": false
+      },
       "previous_version_name": "agoric-upgrade-15",
       "sdk": {
         "type": "cosmos",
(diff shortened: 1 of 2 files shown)


agoric/versions.json
```diff
@@ -55,6 +55,44 @@
...
+      ],
+      "cosmos_sdk_version": "agoric-labs/cosmos-sdk v0.46.16-alpha.agoric.2.4",
+      "consensus": {
+        "type": "cometbft",
+        "version": "v0.34.30",
+        "repo": "https://github.com/agoric-labs/cometbft",
+        "tag": "v0.34.30-alpha.agoric.1"
+      },
+      "cosmwasm_enabled": false,
+      "previous_version_name": "agoric-upgrade-16",
```
JeremyParish69, line 76: because the legacy `cosmwasm_enabled` is here, we'll also require the replacement `cosmwasm` object.

gibson042: Hmm, I don't think I understand... are you asking for `"cosmwasm": { "enabled": false }`? And if so, would that be equivalent to just omitting both `cosmwasm_enabled` and `cosmwasm`?

JeremyParish69: yes, and pretty much yes

gibson042: Done.

PR #5462, opened 2024-10-16
quasar token legacy

vladosda0:
made some changes to QSR token infos due to deprecating the asset and migrating it to EVM


osmosis/assetlist.json
```diff
@@ -11568,9 +11568,9 @@
       ],
       "type_asset": "ics20",
       "base": "ibc/1B708808D372E959CD4839C594960309283424C775F4A038AAEBE7F83A988477",
-      "name": "Quasar",
-      "display": "qsr",
-      "symbol": "QSR",
+      "name": "Quasar (legacy)",
+      "display": "qsr.legacy",
```
JeremyParish69, line 11572: This 'display' unit is not defined anywhere, and therefore the exponent is unknown. I recommend changing this back to what it was.

quasar/assetlist.json
```diff
@@ -18,9 +18,9 @@
       ],
       "type_asset": "sdk.coin",
       "base": "uqsr",
-      "name": "Quasar",
-      "display": "qsr",
-      "symbol": "QSR",
+      "name": "Quasar (legacy)",
+      "display": "qsr.legacy",
```
JeremyParish69, line 22: This 'display' unit is not defined anywhere, and therefore the exponent is unknown. I recommend changing this back to what it was.

osmosis/assetlist.json
vladosda0: Fixed the "display" unit in PR as per feedback

quasar/assetlist.json
vladosda0: Fixed the "display" unit in PR as per feedback

PR #5463, opened 2024-10-16
Upgrade sei to v5.9.0

dreamloss:


sei/chain.json
```diff
@@ -519,12 +519,45 @@
...
+      },
+      {
+        "name": "v5.9.0",
+        "proposal": 85,
+        "height": 107000672,
+        "recommended_version": "v5.9.0",
+        "compatible_versions": [
+          "v5.9.0"
+        ],
+        "cosmos_sdk_version": "sei-cosmos v0.3.37-0.20240923023912-aa7a702d42cc",
```
reversesigh, line 531: I could be wrong but I think the check is failing because `sei-cosos v0.3.37-0.20.....` doesn't match `v0.3.3-0.20....`. @JeremyParish69 is my thinking on this correct? @dreamloss try to fix and re-run and see if that makes the check pass. The same might be true for wasm?

dreamloss: sei go.mod for v5.9.0 lists `sei-cosmos v0.3.37-0.20240923023912-aa7a702d42cc`, do you mean to change that to v0.3.3?

reversesigh: I mean to change `sei-cosmos v0.3.37-0...` to `v0.3.37-0.20240923023912-aa7a702d42cc` and see if the check passes. Same logic for wasmd.

sei/chain.json
```diff
@@ -519,12 +519,45 @@
...
+          "type": "sei-tendermint",
+          "version": "v0.3.8"
+        },
+        "cosmwasm_version": "sei-wasmd v0.2.4",
+        "cosmwasm_enabled": true,
+        "cosmwasm_path": "$HOME/.sei/wasm",
+        "next_version_name": "",
+        "sdk": {
+          "type": "cosmos",
+          "version": "v0.3.37-0.20240923023912-aa7a702d42cc"
```
JeremyParish69, line 542: This is not an accepted version. Instead
version: v0.3.37
tag: v0.3.37-0.20240923023912-aa7a702d42cc
repo: https://github.com/[org-name]/sei-cosmos

sei/chain.json
```diff
@@ -519,12 +519,45 @@
...
+        },
+        "cosmwasm": {
+          "version": "v0.2.4",
+          "repo": "https://github.com/sei-wasmd",
```
JeremyParish69, line 546: I don't think this repo path is valid. What is the org name? I that have to be part of the repo URL

sei/chain.json
```diff
@@ -519,12 +519,45 @@
...
+          "path": "$HOME/.sei/wasm",
+          "enabled": true
+        },
+        "ibc": {
+          "type": "go",
+          "repo": "https://github.com/sei-ibc-go/v3",
```
JeremyParish69, line 552: Doesn't appear to be valid...

sei/chain.json
JeremyParish69: I think this value is OK, but further down where the replacement 'sdk' property is isn't correct.

sei/chain.json
```diff
@@ -212,13 +212,13 @@
...
           "version": "v0.0.4",
-          "repo": "https://github.com/sei-wasmd",
+          "repo": "https://github.com/sei-protocol/sei-wasmd",
           "path": "$HOME/.sei/wasm",
           "enabled": true
         },
         "ibc": {
           "type": "go",
-          "repo": "https://github.com/sei-ibc-go/v3",
+          "repo": "https://github.com/sei-protocol/sei-ibc-go/v3",
```
JeremyParish69, line 221: We'll still need to remove the /v3 at the end of the repo property (for all cases of 'repo')

JeremyParish69:
The tag is OK to keep, it just has to be recorded properly if you were to include it. Otherwise, is this ready to merge?

PR #5493, opened 2024-10-21
change "bech32" prefix for bitbadges

trevormil:
It has come to my attention via X that using the bech32 prefix "cosmos" has broken some services that depend on this registry. BitBadges is showing up where the Cosmos Hub is expected or causing other problems due to the collision.

To avoid confusion and breaking anything, I would like to remove this line from the registry. I noted that "bech32_prefix" is not required in the schema file, so for the time being, I would like to remove it. 

I think this is a better solution because:
-Collisions are avoided.
-If not having the "bech32_prefix" causes issues, this would only occur for BitBadges and we will handle accordingly (rather than affecting other projects).
-Anyone can still just view the BitBadges documentation to get our Cosmos prefix.

I still plan to migrate to a personal prefix, but that may require a complete chain restart or some other difficult feat of engineering.

### DIFF
--- a/bitbadges/chain.json
+++ b/bitbadges/chain.json
@@ -8,6 +8,7 @@
   "pretty_name": "BitBadges",
   "chain_id": "bitbadges-1",
   "daemon_name": "bitbadgeschaind",
+  "bech32_prefix": "bb",
   "node_home": "$HOME/.bitbadgeschaind",
   "slip44": 118,
   "fees": {
--- a/bitbadges/chain.json
+++ b/bitbadges/chain.json
@@ -31,26 +31,26 @@
   },
   "codebase": {
     "git_repo": "https://github.com/bitbadges/bitbadgeschain/",
-    "recommended_version": "v1.0-mainnet",
+    "recommended_version": "v1.0-bb-mainnet",
     "compatible_versions": [
-      "v1.0-mainnet"
+      "v1.0-bb-mainnet"
     ],
     "versions": [
       {
-        "name": "v1.0-mainnet",
+        "name": "v1.0-bb-mainnet",
         "binaries": {
-          "linux/amd64": "https://github.com/BitBadges/bitbadgeschain/releases/download/v1.0-mainnet/bitbadgeschain-linux-amd64",
-          "linux/arm64": "https://github.com/BitBadges/bitbadgeschain/releases/download/v1.0-mainnet/bitbadgeschain-linux-arm64"
+          "linux/amd64": "https://github.com/BitBadges/bitbadgeschain/releases/download/v1.0-bb-mainnet/bitbadgeschain-linux-amd64",
+          "linux/arm64": "https://github.com/BitBadges/bitbadgeschain/releases/download/v1.0-bb-mainnet/bitbadgeschain-linux-arm64"
         },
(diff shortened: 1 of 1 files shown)


trevormil:
Ok, after seeing the validation checks fail, I will just simply update it to "bb" which is the future prefix I plan to migrate to.

I will elevate this and work on handling the migration from "cosmos" to "bb" on my end as soon as possible. Again, sorry for any confusion or inconveniences this caused.

Cordtus:
Yeah typically this is something relating to or identifiable by your chain name or project name or something. `bb` is probably fitting, or `badge` or something like this..

What you were trying to do by ensuring it was the same wallet address across chains (if I am not mistaken) is something you actually don't need to do, this is not EVM.

When you create a wallet, the bech32 prefix is actually part of the derivation of the resulting wallet address. 
As long as you know one of them (and they were derived using the same path of course), you can determine any of the others. You can [test it for yourself here](https://cosmosrescue.com/converter).

trevormil:
@Cordtus Yea, I understand how they work. 

The reasoning isn't exactly as explained. The original idea was more that we didn't want to have our own addresses because we can be thought of as a hub and are supportive of every ecosystem. So, we support Solana, Ethereum, Bitcoin addresses, and in a similar manner, we wanted to make it seem like we support Cosmos ecosystem as well (thus using "cosmos" addresses). See https://docs.bitbadges.io/for-developers/accounts. 

You are right. We can just map "cosmos" addresses to "bb" addresses too. It was just a design decision we made that we will now have to change. I do not actually think it breaks anything on the blockchain level, but it adds confusion / breaks dependent apps on the registry level.

Cordtus:
ah I see I was understanding this part differently:

```
To enable interoperability between different blockchains, each individual L1 blockchain will have its native addresses mapped to an equivalent Cosmos bech32 address. We use the mapped cosmosAddress as the universal base address whenever needed. 
```

so yeah in this case you can actually choose whatever (standard, cointype 118-derived) wallet address you want here, whether it's osmosis, cosmos, stargaze, juno, etc..

bitbadges/chain.json
```diff
@@ -7,8 +7,8 @@
   "website": "https://bitbadges.io/",
   "pretty_name": "BitBadges",
   "chain_id": "bitbadges-1",
-  "bech32_prefix": "cosmos",
   "daemon_name": "bitbadgeschaind",
+  "bech32_prefix": "bb",
```
JeremyParish69, line 11: Can we separate the bech32_prefix change from the other updates (like those pertaining to version) in separate PRs?
I'm not so sure it's appropriate to register a bech32 prefix that is knowingly incorrect, and personally am leaning toward recommending keeping it as 'cosmos' until the chain has actually been updated to use a new prefix.

trevormil: The prefix has been changed to "bb" via a full chain restart as of a couple days ago (10/23 I believe). It wasn't ideal but probably best long term for all parties. 

As for version updates, it was a full chain restart (so new genesis and everything). This is why such details are overwritten rather than new versions.

JeremyParish69: Oh, I didn't realize that it was actually changed; I'm OK with this, then. Thanks for the info!

PR #5502, opened 2024-10-23
Register Synternet IBC information

foxis:
Registering Testnet and Mainnet IBC channels.


testnets/_IBC/synternettestnet-osmosistestnet.json
```diff
@@ -0,0 +1,49 @@
+{
+  "$schema": "../ibc_data.schema.json",
```
JeremyParish69, line 2: this relative path requires and an extra ../ (to go up two directories)
`"$schema": "../../ibc_data.schema.json",`

foxis: Thanks! Fixed.

_IBC/synternet-osmosis.json
```diff
@@ -0,0 +1,32 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "synternet",
```
JeremyParish69, line 4: I overlooked this before, but the chains have to be in alphabetical order (for chain_1 vs chain_2 and the file names)

foxis: Updated.

PR #5528, opened 2024-10-30
chore: Make nibiru name pretty and all chain envs

cgilbe27:


nibiru/chain.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "../chain.schema.json",
-  "chain_name": "nibiru",
+  "chain_name": "Nibiru",
```
JeremyParish69, line 3: This chain_name is not allowed. The pretty_chain_name field is for readability, but the chain_name field is for indexing. It is immutable and isn't allowed spaced or uppercased characters.

cgilbe27: When you say immutable, is it immutable if we are talking about lower chain envs?

cgilbe27: Addressed feedback, assuming chain names for lower envs can be changed

testnets/nibirudevnet/chain.json
```diff
@@ -1,18 +1,16 @@
 {
   "$schema": "../../chain.schema.json",
-  "chain_name": "nibirudevnet",
+  "chain_name": "nibiru-devnet-1",
```
JeremyParish69, line 3: Please do not modify any of the chain_name values. The chain_name must not be changed, and must exactly match the name of the directory it's in.

nibiru/chain.json
JeremyParish69: It's a self-imposed restriction for the Chain Registry that once a chain_name is determined for a project's chain, that it cannot be changed, because it is used as a unique identifier and is referenced externally, making changes to them very dangerous and likely to cause breakage.

JeremyParish69: the Pretty_name value is OK to change.

cgilbe27:
Superseded by: https://github.com/cosmos/chain-registry/pull/5535

PR #5565, opened 2024-11-10
feat: coingecko id for some osmosis wrapped assets

joshlopes:


JeremyParish69:
The coingecko ID only gets added to the original mintages/issuances--not to bridged variants

PR #5585, opened 2024-11-14
Update Dungeon to IBC registry

ApeironNodes:


ApeironNodes:
updated also chain.json of dungeonchain

PR #5592, opened 2024-11-16
Update chain.json after Odin mainnet upgrade

Satchitananda:


JeremyParish69:
Please fix the schemas and ensure the validation checks can pass

odin/chain.json
```diff
@@ -279,8 +303,8 @@
   },
   "explorers": [
     {
-      "kind": "odin web",
-      "url": "https://mainnet.odinprotocol.io/",
+      "kind": "Runa",
+      "url": "https://runa.odinprotocol.io/",
       "tx_page": "https://mainnet.odinprotocol.io/transactions/${txHash}"
```
JeremyParish69, line 308: should the tx_page URL be updated to runa at the same time?

Satchitananda: Agreed, yes. Will make a fix

PR #5604, opened 2024-11-19
Upgrade dydx chain to v7.0

pswies:
As per https://www.mintscan.io/dydx/proposals/173

@reversesigh pls see

### DIFF
--- a/dydx/chain.json
+++ b/dydx/chain.json
@@ -298,7 +298,7 @@
           "type": "go",
           "version": "v8.0.0"
         },
-        "next_version_name": "v7.0.1"
+        "next_version_name": "v7.0"
       },
       {
         "name": "v7.0",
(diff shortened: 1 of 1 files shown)


dydx/chain.json
```diff
@@ -298,6 +298,36 @@
           "type": "go",
           "version": "v8.0.0"
         },
+        "next_version_name": "v7.0.1"
+      },
+      {
+        "name": "v7.0",
```
ttl33, line 304: what does the `next_version_name` indicate? is that just supposed to point to the `name` of the next version which is `v7.0`? 

if that's the case, the `name` should be `v7.0` and the `next_version_name` should also be `v7.0`

https://github.com/cosmos/chain-registry/blob/1437c797123c2972476c2029350fddddd2f06ffc/chain.schema.json#L267-L270

pswies: good catch, @ttl33 . Fixed.

PR #5608, opened 2024-11-19
Update medasdigital Info

oxygene76:


JeremyParish69:
Might as well update the IBC channels at the same time, since I know you're trying to update Osmosis Assetlist at the same time (and that is required)

oxygene76:
Yes, changes added

PR #5612, opened 2024-11-21
add: amATOM to neutron mainnet asset list

chris-ricketts:


neutron/assetlist.json
```diff
@@ -1575,6 +1575,34 @@
...
+        },
+        {
+          "denom": "amATOM",
+          "exponent": 6
+        }
+      ],
+      "base": "factory/neutron1shwxlkpdjd8h5wdtrykypwd2v62z5glr95yp0etdcspkkjwm5meq82ndxs/amatom",
+      "name": "amATOM",
+      "display": "amATOM",
+      "symbol": "amATOM",
```
JeremyParish69, line 1594: Since the symbol isn't solely all CAPS, please include the traces[] to describe how it derives from ATOM

chris-ricketts: I am not sure that any of the currently available `traces` definitions in the JSON schema matches the relationship between `amATOM` and `ATOM`, which you can think of as an asset minted as part of a CDP-type arrangement where `ATOM` is the collateral asset and `amATOM` is the debt asset.

`synthetic` is probably the closest matching type and is the one I've added but the required fields of `chain_name` and `base_denom` seem redundant in this case.

JeremyParish69: Yes, I'd agree with using `synthetic`, and it makes sense to point the ibc-transferred ATOM on Neutron if that's what's deposited to create the position.

chris-ricketts: Understood. I've changed the `base_denom` to IBC transferred ATOM on Neutron.

PR #5623, opened 2024-11-25
Add coingecko ids for assets on Nolus

metodi96:


JeremyParish69:
CoinGecko IDs are currently only registered to the original mintage/issuance of an asset (and not repeated for ICS20 tokens), so I think this should not proceed. CoinGecko IDs can be inherited through the defined traces.

metodi96:
Makes sense. Closing this PR.

PR #5636, opened 2024-11-27
added Jade JSD

Deploydon:
Adds the entry and images for Jade.money $JSD on Neutron.


Deploydon:
Fixed the duplicate error. All checks now pass

PR #5646, opened 2024-12-02
Add Zenrock Mainnet

Marty-Q:
This PR adds Zenrock Mainnet (`diamond-1`) to the chain registry and updates images for testnet.


zenrock/chain.json
```diff
@@ -0,0 +1,114 @@
+{
+    "$schema": "../chain.schema.json",
+    "chain_name": "zenrock",
+    "status": "live",
+    "website": "https://zenrocklabs.io",
+    "network_type": "mainnet",
+    "chain_type": "cosmos",
+    "pretty_name": "Zenrock Mainnet",
+    "chain_id": "diamond-1",
+    "bech32_prefix": "zen",
```
JeremyParish69, line 10: `zen` is taken by Zen protocol?? https://github.com/satoshilabs/slips/blob/master/slip-0173.md
At line: https://github.com/satoshilabs/slips/blob/master/slip-0173.md?plain=1#L293

zenrock/assetlist.json
```diff
@@ -0,0 +1,40 @@
...
+        "symbol": "ROCK",
+        "logo_URIs": {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/zenrock/images/zenrock.png"
+        },
+        "keywords": [
+          "mpc"
+        ],
+        "images": [
+          {
+            "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/zenrock/images/zenrock.png"
```
JeremyParish69, line 31: Why keep so much padding on the png (which will make it appear very tiny) and register as separate images? Might as well remove the padding to make the image appear correctly and then it will visually match and they can be registered as two formats for the same image.

zenrock/chain.json
Marty-Q: Bech32 prefixes are not meant to be exclusive, no? 
We haven't seen it in the chain registry here either

JeremyParish69: This topic was recently discussed and there did turn out to be some issues when two chains shared a bech32 prefix, although that may have been limited to bech32 prefixes actually on the Chain Registry.
However we generally require that all chains register their prefix to that SLIP0173 page once it reaches mainnet, and its probably best to not have any collisions.

zenrock/assetlist.json
JeremyParish69: We should still register them as a single image, now

testnets/zenrocktestnet/assetlist.json
```diff
@@ -21,17 +21,17 @@
...
           "mpc"
         ],
         "images": [
           {
-            "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/zenrocktestnet/images/zenrock.png"
+            "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/zenrock/images/zenrock.png"
           },
           {
-            "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/zenrocktestnet/images/zenrock.svg"
+            "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/zenrock/images/zenrock.svg"
```
JeremyParish69, line 34: Can we combine here as well, please?

zenrock/chain.json
```diff
@@ -0,0 +1,114 @@
...
+      ],
+      "grpc": [
+        {
+            "address": "https://grpc.diamond.zenrocklabs.io/",
+          "provider": "zenrock"
+        }
+      ]
+    },
+    "logo_URIs": {
+      "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/zenrock/images/zenrock.png"
```
JeremyParish69, line 109: And add the svg here

zenrock/chain.json
Marty-Q: I understand your concern of having a collusion-free repo but some heads-up would be great. There is no information on this requirement and the dependency of the SLIP0173 page here on the repo. We can't change our prefix anymore as we are already on mainnet. 

Further, I saw that there are other projects with a shared prefix on this repo (like terra) and the other chain in question is not listed here and not an IBC-enabled chain like us.  

Can you further explain what ind of issues where occurring with two chains sharing a bech32 prefix?

JeremyParish69: I understand the frustration, and are looking to reach consensus on this ASAP.

I don't recall too well what the issue was, but some apps were using the bech32 prefix as a reliable index and would load metadata from the wrong chain (e.g., when users try to view Cosmos Hub data, the app would show data relating to some other chain). Apparently there were multiple apps with issues like this, but I'm not certain.

PR #5660, opened 2024-12-04
update archway mainnet

junaidk:
update archway mainnet data

### DIFF
--- a/archway/chain.json
+++ b/archway/chain.json
@@ -37,33 +37,33 @@
   },
   "codebase": {
     "git_repo": "https://github.com/archway-network/archway",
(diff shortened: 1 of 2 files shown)


archway/chain.json
```diff
@@ -37,33 +37,33 @@
   },
   "codebase": {
     "git_repo": "https://github.com/archway-network/archway",
-    "recommended_version": "v8.0.0",
+    "recommended_version": "v9.0.0",
```
JeremyParish69, line 40: Although the automated verification is yet to come, we'll be requiring that versions referenced here be defined in the versions.json file. Please add v8 and v9 there, as well.

PR #5661, opened 2024-12-04
Add Aaron Network

aaronetwork:


aaronetwork/chain.json
```diff
@@ -0,0 +1,84 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "aaronetwork",
```
JeremyParish69, line 3: I think this chain_name will lead to a lot of confusion. If keeping the 'network' part, then I'd suggest "aaronnetwork" instead.

aaronetwork/assetlist.json
```diff
@@ -0,0 +1,40 @@
...
+      },
+      "socials": {
+        "website": "https://aaronetwork.xyz",
+        "twitter": "https://x.com/aaronetwork"
+      },
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/aaronetwork/images/aaron.png",
+          "theme": {
+            "circle": true
```
JeremyParish69, line 33: this image is not circular. Please change to false

aaronetwork/chain.json
JeremyParish69: Btw, this suggestion is not a requirement. I see that the merging of the two 'n's is used in many places (chain_id, website, twitter, explorer url, etc.)

aaronetwork: Should chain_name be fused? Can we specify "aaron network" or only "aaronnetwork"?

JeremyParish69: Space characters are not allowed. This is just for identifying, and never shown to app users.

PR #5662, opened 2024-12-04
Add WHALE and Register migaloo-archway

calabiyau87:
This PR adds WHALE, Native token of the Migaloo blockchain, to multiple chain assetlists:

Terra2
Terra
Archway
Juno
Chihuahua
Injective

Additionally it also registers migaloo-archway connection in _IBC

### DIFF
--- a/_IBC/archway-migaloo.json
+++ b/_IBC/archway-migaloo.json
@@ -1,23 +1,23 @@
 {
   "$schema": "../ibc_data.schema.json",
   "chain_1": {
-    "chain_name": "migaloo",
-    "client_id": "07-tendermint-138",
-    "connection_id": "connection-108"
-  },
-  "chain_2": {
     "chain_name": "archway",
     "client_id": "07-tendermint-22",
     "connection_id": "connection-114"
   },
+  "chain_2": {
+    "chain_name": "migaloo",
+    "client_id": "07-tendermint-138",
+    "connection_id": "connection-108"
+  },
   "channels": [
     {
       "chain_1": {
-        "channel_id": "channel-141",
+        "channel_id": "channel-184",
         "port_id": "transfer"
       },
       "chain_2": {
-        "channel_id": "channel-184",
+        "channel_id": "channel-141",
         "port_id": "transfer"
       },
       "ordering": "unordered",
--- a/_IBC/archway-migaloo.json
+++ b/_IBC/archway-migaloo.json
@@ -2,7 +2,7 @@
   "$schema": "../ibc_data.schema.json",
   "chain_1": {
     "chain_name": "archway",
-    "client_id": "07-tendermint-22",
+    "client_id": "07-tendermint-119",
     "connection_id": "connection-114"
   },
   "chain_2": {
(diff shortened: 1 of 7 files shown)


_IBC/archway-migaloo.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "migaloo",
```
JeremyParish69, line 4: chain_1 must be `archway` in this case, not `migaloo` (to follow alphabetical order)

calabiyau87: @JeremyParish69 Corrected

calabiyau87:
I did a test transfer from Terra (LUNC) to Migaloo:
https://finder.terra.money/classic/tx/B349394425A91FFABB17E4942F496A264B9616857B75DB4D4D4C6451D20F7E99

The channels don't match what is already registerd in _IBC/migaloo-terra.json

Do I need to register another IBC path or update the existing?

JeremyParish69:
> I did a test transfer from Terra (LUNC) to Migaloo: https://finder.terra.money/classic/tx/B349394425A91FFABB17E4942F496A264B9616857B75DB4D4D4C6451D20F7E99
> 
> The channels don't match what is already registerd in _IBC/migaloo-terra.json
> 
> Do I need to register another IBC path or update the existing?

Only one transfer/transfer channel may be registered, but none exists currently, so this transfer/transfer channel may be added to the existing file (above the current channel, because that's has a wasm channel). It would be required to registered WHALE on Terra Classic.

terra/assetlist.json
```diff
@@ -6359,6 +6359,58 @@
...
+      "name": "Migaloo",
+      "display": "whale",
+      "symbol": "WHALE",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "migaloo",
+            "base_denom": "uwhale",
+            "channel_id": "channel-114"
```
JeremyParish69, line 6389: This looks to be incorrect

calabiyau87:
So what is the schema to add an additional channel because there is a new client id, channel, and connection id for the transfer/transfer between migaloo<>terra
Screenshot 2024-12-05 115234
Screenshot 2024-12-05 115309
Screenshot 2024-12-05 115353

JeremyParish69:
> So what is the schema to add an additional channel because there is a new client id, channel, and connection id for the transfer/transfer between migaloo<>terra

In line with my original think, previous cases where a wasm. IBC transfer port is registered as an additional channel, it shares the same clients and connection ids as the original transfer/transfer channel.

I'm not totally sure how to proceed, here. Either: 
- 1) this new channel (and the assets associated with) cannot be registered because it's not using the canonical IBC connection-ids and clients arleady registered for this chain pair--OR,
 - 2) if we can show that the transfer/transfer channel's connection and clients were created first, in use first, and have more transfer volume, then we can work to de-list the assets associated with the existing wasm. connection, and remove that wasm. channel from the registry, but that would be a big effort, since registering an IBC connection is generally considered permanent.

calabiyau87:
So no way of registering a migaloo-terra_2.json with all the details mentioned above?

JeremyParish69:
> So no way of registering a migaloo-terra_2.json with all the details mentioned above?

Alas, no, there's not

PR #5671, opened 2024-12-07
add dymension playground testnet

maxzonder:


dylanschultzie (approved):

dylanschultzie:
This LGTM, @JeremyParish69 can you review if you have time?

testnets/dymensionplaygroundtestnet/assetlist.json
```diff
@@ -0,0 +1,164 @@
+{
+  "$schema": "../../assetlist.schema.json",
+  "chain_name": "dymensionplaygroundtestnet",
```
JeremyParish69, line 3: this should just be 'dymensiontestnet', not have 'playground' in there.

testnets/dymensionplaygroundtestnet/assetlist.json
```diff
@@ -0,0 +1,164 @@
...
+      "type_asset": "sdk.coin",
+      "base": "adym",
+      "name": "Dymension",
+      "display": "dym",
+      "symbol": "DYM",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/dymensionplaygroundtestnet/images/dymension-logo.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/dymensionplaygroundtestnet/images/dymension-logo.svg"
+      },
+      "coingecko_id": "dymension",
```
JeremyParish69, line 26: We aren't adding CGIDs to testnet tokens.

testnets/dymensionplaygroundtestnet/assetlist.json
```diff
@@ -0,0 +1,164 @@
...
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/dymensionplaygroundtestnet/images/dymension-logo.png",
```
JeremyParish69, line 29: These images aren't identical...
And there already exists dymension logos in the registry, so we don't need duplicates.

testnets/dymensionplaygroundtestnet/assetlist.json
```diff
@@ -0,0 +1,164 @@
...
+      "type_asset": "ics20",
+      "base": "ibc/FB53D1684F155CBB86D9CE917807E42B59209EBE3AD3A92E15EF66586C073942",
+      "name": "Nim Network",
+      "display": "nim",
+      "symbol": "NIM",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "nim",
```
JeremyParish69, line 66: Usually mainnet tokens aren't being sent to a testnet. This likely comes from nimtestnet.

testnets/dymensionplaygroundtestnet/assetlist.json
```diff
@@ -0,0 +1,164 @@
...
+      "type_asset": "ics20",
+      "base": "ibc/5A26C8DC8DA66F4DD94326E67F94510188F5F7AFE2DB3933A0C823670E56EABF",
+      "name": "Mande Network",
+      "display": "mand",
+      "symbol": "MAND",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "mande",
```
JeremyParish69, line 127: Same issue here--likely not from mainnet mande

JeremyParish69:
I think we should revert this change and submit again once the issues are corrected

dylanschultzie:
thanks for the second set of eyes, @JeremyParish69

maxzonder:
thanks @JeremyParish69 

re: dymensionplaygroundtestnet - dymension runs 2 testnets, blumbus and playground, that's where the name comes from

re: nim and mande - on playground testnet, these tokens have exactly the same ibc denom and path as mainnet tokens

re: images and coingecko - i can fix it with PR

what do you think? cc @dylanschultzie

JeremyParish69:
> re: dymensionplaygroundtestnet - dymension runs 2 testnets, blumbus and playground, that's where the name comes from
> 
When there are multiple simultaneous testnets, we append an incrementing integer: 'dymensiontestnet1', dymentiontestnet2', ..., etc. The name "playground" can be retained in the pretty_name and chain_id.

> re: nim and mande - on playground testnet, these tokens have exactly the same ibc denom and path as mainnet tokens
> 
It's fine for the ibc denoms to happen to match, but the chain_names referenced here are necessarily mainnet names. Assuming they actually come from a testnet, it should instead say 'mandetestnet' and 'nimtestnet' (however they're registered)

> re: images and coingecko - i can fix it with PR
> 
This would be great

> what do you think? cc @dylanschultzie

PR #5680, opened 2024-12-10
docs: remove unnecessary commas

ChengenH:


JeremyParish69:
The commas are grammatically correct

PR #5690, opened 2024-12-12
Update AssetMantle

kombos:
Updated logo and fee params of AssetMantle chain


kombos:
will re-raise after upstream syncing

PR #5691, opened 2024-12-12
Changed wrong rpc

Quasar-hub:
Changed wrong rpc wich is api endpoint


ankurdotb (approved):
The edit is correct, although admittedly confusing since this is a REST API and not an RPC API URI. @Quasar-hub do you control this node by any chance? It will be less confusing if you get the operator of this API to stick to convention.

Functionally though, this edit it correct.

Quasar-hub:
@ankurdotb Hello,
We've identified an issue with an RPC endpoint that doesn't belong to our system. We're receiving daily failure notifications because this endpoint is not responding in the expected RPC format - instead, it's responding like a REST API endpoint.
Rather than removing the endpoint completely, we've decided to move it to the API section since that better matches its behavior. We can remove this endpoint entirely if needed.
Key points:

The endpoint is external to our system
It's causing daily run failures due to incorrect response format
Current behavior matches API rather than RPC patterns
We've relocated it to the API section as a temporary solution
We can delete the endpoint if preferred

Please let us know how you'd like to proceed.

JeremyParish69:
So now there's two ApieronNodes Archive REST endpoints, so are they functionally identical but they just have different URI? I think it might be best to remove the rpc one since the naming is confusing, and there is still an Archive REST endpoint from the same provider.

Quasar-hub:
> So now there's two ApieronNodes Archive REST endpoints, so are they functionally identical but they just have different URI? I think it might be best to remove the rpc one since the naming is confusing, and there is still an Archive REST endpoint from the same provider.

done 👌

ankurdotb (approved):

Quasar-hub:
@JeremyParish69 merge please

PR #5742, opened 2024-12-19
chore: add elys versions

cosmic-vagabond:
add new versions of Elys Network

### DIFF
--- a/elys/versions.json
+++ b/elys/versions.json
@@ -3,6 +3,106 @@
   "chain_name": "elys",
   "versions": [
     {
+      "name": "v1.4.0",
+      "recommended_version": "v1.4.0",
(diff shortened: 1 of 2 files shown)


elys/versions.json
```diff
@@ -2,6 +2,106 @@
   "$schema": "../versions.schema.json",
   "chain_name": "elys",
   "versions": [
+    {
```
JeremyParish69, line 5: Good to include this info, but it's in the wrong order. The latest version should be at the end of the array, not the start

cosmic-vagabond: @JeremyParish69 thanks for the heads up I just updated the order as suggested please review again thanks

PR #5748, opened 2024-12-22
Update Aaron Network

aaronetwork:


osmosis/assetlist.json
```diff
@@ -25575,6 +25575,56 @@
         }
       ]
     },
+    {
```
JeremyParish69, line 25578: Please relocate the asset to the bottom of the `assets` array.

PR #5752, opened 2024-12-23
add Elys logo URI and description to chain.json

ollieottersync:


elys/chain.json
```diff
@@ -73,6 +73,10 @@
       "enabled": false
     }
   },
+"logo_URIs": {
+    "png": "https://github.com/cosmos/chain-registry/blob/master/elys/images/elys.png",
```
JeremyParish69, line 77: The png url must point to the raw file.
https://raw.githubusercontent.com/cosmos/chain-registry/master/elys/images/elys.png

ollieottersync: Updated the link, thank you

JeremyParish69: Sorry, but that's still not the right URL; it must go to the raw file directly (our URLs are not quite the same)

ollieottersync:
Is the link in commit 5b41c02d862f34db4cb5f0a13c7f6a4b249cfa20 correct? To avoid confusion I removed the first commit, but I believe this one is good to go.

PR #5759, opened 2024-12-26
router usdc(ibc) added

Iamshashvat:


Iamshashvat:
@JeremyParish69, this is for noble usdc (IBC) on router chain, Can you please take a look?

routerchain/assetlist.json
```diff
@@ -38,6 +38,79 @@
...
+            "base_denom": "uusdc"
+          },
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/ethereum/images/usdc.png",
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/ethereum/images/usdc.svg",
+          "theme": {
+            "circle": true,
+            "primary_color_hex": "#2775CA"
+          }
+        },
+        {
```
JeremyParish69, line 110: This final image object is unnecessary, since it's the same as above

Iamshashvat: removed image object

PR #5767, opened 2025-01-01
feat: add vector network

Sen-Com:


vector/chain.json
```diff
@@ -0,0 +1,107 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "vector",
+  "status": "live",
+  "network_type": "mainnet",
+  "website": "https://www.playonvector.com/",
+  "pretty_name": "Vector",
+  "chain_type": "cosmos",
+  "chain_id": "vector-1",
+  "bech32_prefix": "vector",
```
JeremyParish69, line 10: It is recommended to register the bech32 prefix here: https://github.com/satoshilabs/slips/blob/master/slip-0173.md

PR #5771, opened 2025-01-02
add ashION to Osmosis

Nostradamus411:


osmosis/assetlist.json
```diff
@@ -26481,6 +26481,37 @@
...
+          "denom": "ashION",
+          "exponent": 6
+        }
+      ],
+      "type_asset": "sdk.coin",
+      "address": "osmo1svj5kd8kzj7xxtrd6ftjk0856ffpyj4egz7f9pd9dge5wr4kwansmefq07",
+      "base": "factory/osmo1svj5kd8kzj7xxtrd6ftjk0856ffpyj4egz7f9pd9dge5wr4kwansmefq07/ion.ash",
+      "name": "Burned ION",
+      "display": "ashION",
+      "symbol": "ashION",
```
JeremyParish69, line 26503: Just a heads up we're going to be more adherent to ensuring that symbols only have UPPERCASE characters unless they also have `traces` defined. We'll either need to come up with a trace type for burns, or else change all the ashASSETs to ASHASSET.

PR #5773, opened 2025-01-03
update asset lists for mantra mainnet and dukong testnet

alex-stmr:


testnets/mantrachaintestnet2/assetlist.json
```diff
@@ -62,6 +62,218 @@
...
+      "type_asset": "ics20",
+      "base": "ibc/ED07A3391A112B175915CD8FAF43A2DA8E4790EDE12566649D0C2F97716B8518",
+      "name": "Osmosis",
+      "display": "osmo",
+      "symbol": "OSMO",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "osmo-test-5",
```
JeremyParish69, line 198: The chain_name for (chain-id): `"osmo-test-5"` is `"osmosistestnet"`

testnets/mantrachaintestnet2/assetlist.json
```diff
@@ -62,6 +62,218 @@
...
+      "type_asset": "ics20",
+      "base": "ibc/F082B65C88E4B6D5EF1DB243CDA1D331D002759E938A0F5CD3FFDC5D53B3E349",
+      "name": "USDC",
+      "display": "usdc",
+      "symbol": "USDC",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "noble",
```
JeremyParish69, line 91: This is a reference to mainnet, but I doubt mainnet USDC is being sent to a testnet... I assume this is meant to be testnet USDC instead?

mantrachain/assetlist.json
```diff
@@ -58,6 +58,219 @@
...
+          "denom": "usdc",
+          "exponent": 6
+        }
+      ],
+      "type_asset": "ics20",
+      "base": "ibc/65D0BEC6DAD96C7F5043D1E54E54B6BB5D5B3AEC3FF6CEBB75B9E059F3580EA3",
+      "name": "USDC",
+      "display": "usdc",
+      "symbol": "USDC",
+      "coingecko_id": "usd-coin",
```
JeremyParish69, line 83: At the moment, we only register coingecko_id on the original asset (and not ibc-transferred tokens)

testnets/mantrachaintestnet2/assetlist.json
alex-stmr: Indeed it's intended to be a USDC from nobletestnet thanks for the catch 🙏

mantrachain/assetlist.json
alex-stmr: @JeremyParish69 Noted, it got removed so it can be merged 👍 

Ultimately we would like to leverage the `coingecko_id` for each asset so we could rely on it in some frontend Dapps

Do you know when and which id we could use for the USDC here in the future?

testnets/mantrachaintestnet2/assetlist.json
```diff
@@ -62,6 +62,218 @@
...
+      "name": "Osmosis",
+      "display": "osmo",
+      "symbol": "OSMO",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "osmosistestnet",
+            "base_denom": "uosmo",
+            "channel_id": "channel-9126"
```
JeremyParish69, line 200: This is not the correct channel ID
https://github.com/cosmos/chain-registry/blob/master/testnets/_IBC/mantrachaintestnet-osmosistestnet.json#L20

testnets/mantrachaintestnet2/assetlist.json
```diff
@@ -62,6 +62,218 @@
...
+      "name": "Ondo US Dollar Yield",
+      "display": "usdy",
+      "symbol": "USDY",
+      "traces": [
+          {
+              "type": "ibc",
+              "counterparty": {
+                  "chain_name": "nobletestnet",
+                  "base_denom": "ausdy",
+                  "channel_id": "channel-237"
```
JeremyParish69, line 147: Not the correct channel id
https://github.com/cosmos/chain-registry/blob/master/testnets/_IBC/mantrachaintestnet-nobletestnet.json

mantrachain/assetlist.json
JeremyParish69: At the moment, there isn't any approriate cgid to have here.
The coingecko ID can be derived via the data in the `traces` property, but I understand that many apps don't implement the logic to derive it.

In the future this may be expanded to allow IBC-transferred tokens to share the original coingecko ID, but at the moment we don't do this.

testnets/mantrachaintestnet2/assetlist.json
```diff
@@ -62,6 +62,218 @@
...
+      "type_asset": "ics20",
+      "base": "ibc/A4DB47A9D3CF9A068D454513891B526702455D3EF08FB9EB558C561F9DC2B701",
+      "name": "Cosmos Hub",
+      "display": "atom",
+      "symbol": "ATOM",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "cosmoshub",
```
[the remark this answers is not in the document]
alex-stmr: updated

testnets/mantrachaintestnet2/assetlist.json
alex-stmr: updated

testnets/mantrachaintestnet2/assetlist.json
alex-stmr: updated too

PR #5776, opened 2025-01-05
Add alloratestnet

kpeluso:
Adding a new testnet for the Allora Network: alloratestnet


testnets/alloratestnet/chain.json
```diff
@@ -0,0 +1,174 @@
...
+    {
+      "kind": "Moonlet",
+      "url": "https://explorer.moonlet.cloud/allora",
+      "tx_page": "https://explorer.moonlet.cloud/allora/tx/${txHash}",
+      "account_page": "https://explorer.moonlet.cloud/allora/account/${accountAddress}"
+    }
+  ],
+  "images": [
+    {
+      "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/alloratestnet/images/allora.png",
```
JeremyParish69, line 170: Note the the padding on the images will make these logos appear very small in most applications. A version without padding should be used instead if it's intended to be as big as possible.

PR #5782, opened 2025-01-07
Update Nolus testnet info

metodi96:


testnets/nolustestnet/chain.json
```diff
@@ -6,7 +6,7 @@
   "network_type": "testnet",
   "pretty_name": "Nolus Testnet",
   "chain_type": "cosmos",
-  "chain_id": "rila-1",
+  "chain_id": "rila-3",
```
JeremyParish69, line 9: If moving to a new chain, please close out the IBC connections that still exist for the rila-1 version of nolustestnet, or better yet, archive rila-1 to be nolustestnet1, and update the ibc connection files accordingly, and then rila-3 can take its place as 'nolustestnet'.

testnets/nolustestnet/chain.json
```diff
@@ -36,29 +36,29 @@
   },
   "codebase": {
     "git_repo": "https://github.com/nolus-protocol/nolus-core",
-    "recommended_version": "v0.4.0",
+    "recommended_version": "0.6.8-rc",
```
JeremyParish69, line 39: Please don't forget to update the versions.json file the same way

PR #5787, opened 2025-01-08
add arbLUNA

0xphilipp:


terra2/assetlist.json
```diff
@@ -1744,6 +1744,33 @@
...
+        },
+        {
+          "denom": "arbLUNA",
+          "exponent": 6
+        }
+      ],
+      "base": "cw20:terra1se7rvuerys4kd2snt6vqswh9wugu49vhyzls8ymc02wl37g2p2ms5yz490",
+      "name": "ERIS Arb LUNA",
+      "display": "arbLUNA",
+      "symbol": "arbLUNA",
```
JeremyParish69, line 1765: Please include `traces` when using lowercase letters in the symbols. How doese this derive from LUNA?

0xphilipp: its a vault, that is LUNA underlying but trading it to increase the underlying value. I used synthetic as the trace

PR #5791, opened 2025-01-09
Add dTIA to MilkyWay

RiccardoM:
## Description

This PR adds the support for dTIA as a fee token inside MilkyWay


milkyway/assetlist.json
```diff
@@ -286,6 +286,73 @@
...
+        },
+        {
+          "denom": "dTIA",
+          "exponent": 6
+        }
+      ],
+      "type_asset": "ics20",
+      "base": "ibc/4795C86F72EC2CB54E885A4A5B401BED9C333E008311ED9C4DE01BF546B44A21",
+      "name": "Stride Staked TIA",
+      "display": "stTIA",
```
JeremyParish69, line 308: this unit isn't defined among the denom_units

RiccardoM: Fixed it, thank you!

osmosis/assetlist.json
```diff
@@ -25759,6 +25759,61 @@
         }
       ]
     },
+    {
```
JeremyParish69, line 25762: Please only add assets to the very bottom on the existing assets array

JeremyParish69: Nvm, In this case I see how it makes sense to put the TIA straight from Celestia above the TIA the passes through Osmosis

PR #5798, opened 2025-01-09
Update assetlist.json

Quasar-hub:


dungeon/assetlist.json
```diff
@@ -35,7 +35,33 @@
           }
         }
       ],
-      "type_asset": "sdk.coin"
+      "type_asset": "sdk.coin",
+      "traces": [
```
JeremyParish69, line 39: Again, this is not how traces is used. To register an IBC-transferred version of DGN, they must be registered on those chains' assetlist.json file--NOT in this dugeon/assetlist.json file.

Quasar-hub: This is done following the example of other networks. In cosmoshub assetlist, and osmosis it is done

Quasar-hub: Help please how we can do correctly working ibc. Now it is don't work

PR #5814, opened 2025-01-14
Update ATONE asset on Osmosis with new IBC channels & denom

arlai-mk:
A new IBC connection has been setup for AtomOne <-> Osmosis, and PR #5812 has already been merged to reflect the new IBC connection.

Now updating assets on Osmosis, see below the confirmation of the IBC asset denom: 
```bash
$ echo -n transfer/channel-94814/uatone | sha256sum
bc26a7a805ecd6822719472bcb7842a48ef09df206182f8f259b2593eb5d23fb  -
```

### DIFF
--- a/osmosis/assetlist.json
+++ b/osmosis/assetlist.json
@@ -25380,7 +25380,7 @@
       "description": "The native staking and governance token of AtomOne",
       "denom_units": [
         {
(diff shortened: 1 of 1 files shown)


JeremyParish69:
This wasn't required to show on Osmosis Zone, but a change like this would have happened eventually via an automated process. There's no harm in updating it here now, so we can proceed.

PR #5815, opened 2025-01-14
KB-1230: Update Kima testnet URLs

Ben-Kima:


kimanetwork/chain.json
```diff
@@ -34,13 +34,24 @@
...
-    "compatible_versions": [
-      "v0.4.3"
-    ],
+    "recommended_version": "v0.4.5",
+    "compatible_versions": ["v0.4.5"],
     "genesis": {
       "genesis_url": "https://archive.kima.network/genesis.json"
-    }
+    },
+    "versions": [
```
JeremyParish69, line 42: `"versions"` not longer belongs under `"codebase"`, but instead goes to another file called `versions.json`.
The `"versions"` property must be removed from both chain.json files to pass validation.

Ben-Kima: Thanks Jeremy, I have quite a few changes including this locally that we will push in the coming days and then look forward to your feedback thanks.

Ben-Kima: @JeremyParish69 the changes should now be ready for your review with the checks all passing ✅ 

I'm curious as to whether you have any links or docs you can share that better describe the traces section? They would be greatly appreciated 🙏

kimanetwork/chain.json
```diff
@@ -30,16 +30,40 @@
...
-    "recommended_version": "v0.4.3",
-    "compatible_versions": [
-      "v0.4.3"
-    ],
+    "recommended_version": "v0.4.5",
+    "compatible_versions": ["v0.4.5"],
     "genesis": {
       "genesis_url": "https://archive.kima.network/genesis.json"
+    },
+    "consensus": {
```
JeremyParish69, line 45: The consensus, language, sdk, and ibc, are better stored in the "versions" file (but here is fine to include, as well).

testnets/kimanetworktestnet/assetlist.json
```diff
@@ -43,6 +43,7 @@
           "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/kimanetwork/images/kima.svg"
         }
       ],
+      "coingecko_id": "kima",
```
JeremyParish69, line 46: Please remove the coingecko ID. We don't add them to testnet assets

kimanetwork/chain.json
JeremyParish69: I left just a couple more suggestions for corrects and then it should be good to go.
Unfortunately I don't have more documentation of `traces` at the moment. I think I'll need to write some.

testnets/kimanetworktestnet/chain.json
```diff
@@ -25,61 +25,90 @@
...
+    "recommended_version": "v0.4.6",
     "compatible_versions": [
-      "v0.4.3"
+      "v0.4.6"
     ],
     "genesis": {
-      "genesis_url": "http://data-testnet.kima.finance/genesis.json"
+      "genesis_url": "http://archive.sardis.kima.network/genesis.json"
+    },
+    "consensus": {
```
JeremyParish69, line 47: Please copy or move these properties (consensus, lenguage, sdk, ibc) into the version object in the "versions.json" file.

Ben-Kima:
@JeremyParish69 thanks for all of your feedback, I've updated accordingly and trust that all should be in accordance with your standards currently.

gilkima (approved):
lgtm

testnets/kimanetworktestnet/assetlist.json
JeremyParish69: just FYI, we now allow this now that sufficient validation has been added

PR #5843, opened 2025-01-22
Add $SAM logos to Oraichain images folder

allomart1:
$SAM is the first native memecoin in Oraichain.


JeremyParish69:
No asset is being defined in the assetlist. We do not accept images without anything defining what they are. Also, the images exceed the maximum file size (250 kB)

PR #5851, opened 2025-01-23
Updated Icons for kopi chain

FlatWhiteWithSugar:
Updated Icons for kopi chain


kopi/assetlist.json
```diff
@@ -163,13 +163,13 @@
       "base": "ucusdtinj",
       "name": "cUSDT.inj",
       "display": "cUSDT.inj",
-      "symbol": "CUSDT.inj",
+      "symbol": "`CUSD  `T.inj",
```
JeremyParish69, line 166: we can't have this value for symbol. Please use either all UPPERCASE or can have some lowercase and certain symbols if traces is defined.

kopi/assetlist.json
```diff
@@ -238,7 +238,7 @@
       "base": "ibc/D8A36AE90F20FE4843A8D249B1BCF0CCDDE35C4B605C8DED57BED20C639162D0",
       "name": "USDT.inj",
       "display": "USDT.inj",
-      "symbol": "USDT.inj",
+      "symbol": "USDT.INJ",
```
JeremyParish69, line 241: for USDT.inj, it makes sense to keep the ".inj" wit the 'inj' letters being lowercase.

kopi/assetlist.json
```diff
@@ -163,13 +163,13 @@
       "base": "ucusdtinj",
       "name": "cUSDT.inj",
       "display": "cUSDT.inj",
-      "symbol": "CUSDT.inj",
+      "symbol": "CUSDT.INJ",
```
JeremyParish69, line 166: For CUSDT, there is no `traces` define, so we cannot have the "." in the symbol either.

kopi/assetlist.json
```diff
@@ -458,13 +458,13 @@
       "base": "uasusdtinj",
       "name": "asUSDT.inj",
       "display": "asusdtinj",
-      "symbol": "ASUSDT.inj",
+      "symbol": "ASUSDT.INJ",
```
JeremyParish69, line 461: Since there is no `traces` defined, so we cannot have the "." in the symbol either.

FlatWhiteWithSugar: I have updated to include the traces as the tokens are wrapped version of USDT based on either being the arb or MM token receipt

kopi/assetlist.json
FlatWhiteWithSugar: I have updated to include the traces as the tokens are wrapped version of USDT based on either being the arb or MM token receipt

kopi/assetlist.json
FlatWhiteWithSugar: I have updated the token symbol to USDT.inj

kopi/assetlist.json
```diff
@@ -165,19 +165,32 @@
...
       "keywords": [
         "dex",
         "mm"
       ],
-      "type_asset": "sdk.coin"
-     
+      "type_asset": "sdk.coin",
+     "traces": [
+        {
+          "type": "ibc",
```
JeremyParish69, line 182: There's something different going on here than just a simple IBC transfer. IBC tokens must be type `ics20`, but this is `sdk.coin`. Perhaps if this follows the same price and the convertible back into USDC, this may be a wrapped version of USDC.inj from Kopi (i.e., of  "ibc/ibc/D8A36AE90F20FE4843A8D249B1BCF0CCDDE35C4B605C8DED57BED20C639162D0").
Or perhaps it's a type: "synthetic" version of the peggy asset referenced here.

kopi/assetlist.json
```diff
@@ -460,18 +473,32 @@
...
       ],
       "keywords": [
         "dex",
         "mm"
       ],
-      "type_asset": "sdk.coin"
+      "type_asset": "sdk.coin",
+      "traces": [
+        {
+          "type": "ibc",
```
JeremyParish69, line 490: Similar issue as with cUSDT.inj
Not correct to use "ibc", but not sure the exact nuances of as-ASSETs or c-ASSETs to say exactly which trace it should be. A Price peg would likely qualify it to use a trace, and backward convertibility would likely tell us which asset to refer to.

FlatWhiteWithSugar: c assets and as assets are protocol created tokens, that are directly converted into the corresponding token based on a mint and burn mechanism users can deposit USDT into the native mm and are given c asset back which is an auto compounding asset of USDT in this example, as is the arb assets of c assets, they can all be converted back into the underlying IBC token

JeremyParish69: Thanks for explaining that. I think cASSETs currently don't have a great fit into any existing trace type just yet.
I think it's best to remove the traces definition for now

PR #5852, opened 2025-01-23
update Quicksilver json files, denoms and assets

joe-bowman:
- update github org
- update versions (add versions.json)
- update chain.json to validate against latest schema
- add latest denoms + assets
- update old assets for consistency (and make svgs actual svgs, and not just wrapped base64 encoded pngs)
- add new assets on osmosis
- add missing ibc channels


quicksilver/assetlist.json
```diff
@@ -473,6 +473,264 @@
...
+      ],
+      "base": "uqluna",
+      "name": "Quicksilver Liquid Staked LUNA",
+      "display": "qluna",
+      "symbol": "qLUNA",
+      "traces": [
+        {
+          "type": "liquid-stake",
+          "counterparty": {
+            "chain_name": "terra",
```
JeremyParish69, line 671: I suspect this is meant to refer to terra2, since there is an IBC connection being registered between quicksilver and terra2 (and not terra classic)

quicksilver/versions.json
```diff
@@ -434,7 +434,7 @@
         "v1.6.1"
       ],
       "binaries": {
-        "linux/amd64": "https://github.com/ingenuity-build/quicksilver/releases/download/v1.6.1/quicksilverd-v1.6.1-amd64"
+        "linux/amd64": "https://github.com/quicksilver-zone/quicksilver/releases/download/v1.6.1/quicksilverd-v1.6.1-amd64"
```
JeremyParish69, line 437: Note that it is recommended (and may soon be updated automatically) to combine all patch and minor versions into a single object representing the major version (in this case, v1 would include v1.6.1, v1.6.3, v1.7.5, etc. as compatible versions with each other).

quicksilver/assetlist.json
```diff
@@ -473,6 +473,264 @@
...
+          "type": "liquid-stake",
+          "counterparty": {
+            "chain_name": "archway",
+            "base_denom": "aarch"
+          },
+          "provider": "Quicksilver"
+        }
+      ],
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/quicksilver/images/qarch.png",
```
JeremyParish69, line 506: Note that it is strongly recommended (and required for sever cases) to remove all margin/padding from token images. Many of the uploaded images have transparent padding to varying degrees.

quicksilver/versions.json
joe-bowman: Quicksilver hasn't been applying Semver correctly (fixing this soon!), but these versions are not compatible.

I appreciate this is on us to resolve, but won't this screw up all historic versions (and for other chains not using semver accurately?)

quicksilver/assetlist.json
joe-bowman: fixed.

quicksilver/assetlist.json
joe-bowman: Done.

joe-bowman:
Requested changes made @JeremyParish69 - r4r

quicksilver/versions.json
JeremyParish69: Thanks for explaining. Yes, blindly assuming semver has been followed correctly for all chains, and then combining versions as if it were, could mess up version histories, and likely cannot be done safely. However, I'd at least opportunistically ask about cases where I suspect it may possible to clean up a version history. Since these patch versions are not compatible, we can keep these separated as is.

PR #5869, opened 2025-01-28
bump bitsong version upgrade history & public endpoints

hard-nett:


bitsong/versions.json
```diff
@@ -46,6 +46,163 @@
...
+        "type": "go",
+        "version": "v7.4.0"
+      },
+      "binaries": {
+        "linux/amd64": "https://github.com/bitsongofficial/go-bitsong/releases/download/v0.20.2/bitsongd-linux-amd64",
+        "linux/arm64": "https://github.com/bitsongofficial/go-bitsong/releases/download/v0.20.2/bitsongd-linux-arm64"
+      }
+    },
+    {
+      "name": "v0.20.4",
```
JeremyParish69, line 182: I don't think v0.20.4 needs it's own version object since it's compatible with other v0.20 patches

bitsong/versions.json
```diff
@@ -46,6 +46,163 @@
...
+      "name": "v0.20.0",
+      "proposal": 42,
+      "height": 20277056,
+      "recommended_version": "v0.20.2",
+      "compatible_versions": [
+        "v0.20.0",
+        "v0.20.1",
+        "v0.20.2",
+        "v0.20.3",
+        "v0.20.4"
```
JeremyParish69, line 161: is v0.20.4 compatible?

bitsong/chain.json
```diff
@@ -13,26 +13,26 @@
...
     "git_repo": "https://github.com/bitsongofficial/go-bitsong",
-    "recommended_version": "v0.15.0",
+    "recommended_version": "v0.20.4",
     "compatible_versions": [
-      "v0.15.0"
+      "v0.20.4"
     ],
     "binaries": {
-      "linux/amd64": "https://github.com/bitsongofficial/go-bitsong/releases/download/v0.15.0/bitsongd"
+      "linux/amd64": "https://github.com/bitsongofficial/go-bitsong/releases/download/v0.20.4/bitsongd"
```
JeremyParish69, line 21: If including `binaries` in the chain.json, it must match the current version, which has two binaries defined (amd and arm), whereas there is only one (just amd) defined here.

bitsong/versions.json
hard-nett: v0.20.0- v0.20.3 is not compatible with v0.20.4, as we have patched the sdk in v0.20.4.

bitsong/versions.json
hard-nett: separating v0.20.4 here is needed as v0.20.4 at first was incorrectly labeled comatible with the previous v0.20 releases

bitsong/chain.json
JeremyParish69: The binaries here still need to match

PR #5877, opened 2025-01-30
Update `ibc` neutron-nomic

clemensgg:
previous paths were not used in production


JeremyParish69:
The only channel doesn't even show up on Mintscan

PR #5895, opened 2025-02-04
Update sei testnet2 chain.json

cordt-sei:
Updates suggested+supported versions (with latest only)

Adds:
 - new peer nodes
 - new block explorer
 
 Let me know if I should also update the "versions.json" or if that is generated.


JeremyParish69:
Yes, please update the compatible and recommended version in the versions.json file to keep it in sync with the chain.json file.

JeremyParish69:
Will this be completed? Or should we close this PR?

PR #5904, opened 2025-02-06
Updated IBC Denoms and channels 

FlatWhiteWithSugar:
Updated IBC Denoms and channels


kopi/assetlist.json
```diff
@@ -213,11 +213,11 @@
           "counterparty": {
             "chain_name": "noble",
             "base_denom": "uusdc",
-            "channel_id": "channel-109"
+            "channel_id": "channel-122"
```
JeremyParish69, line 216: This channel is incorrect. The IBC connection between noble and kopi is already registered. That registration must be updated first, and that may only be changed with governance agreement or proof that there are no assets abandoned on the counterparty chains (both ways).

FlatWhiteWithSugar: https://github.com/noble-assets/noble/issues/477 we have reached out to noble to get the client updated

FlatWhiteWithSugar: I have reverted the changes for noble

JeremyParish69:
It looks like a lot of these assets are referring to IBC connections that aren't registered, causes checks to fail. Please only update for registered IBC connections

JeremyParish69:
It looks like there is a small amount of XKP transferred over to Osmosis. Small testing amounts may abandoned if the remaining evidence is solid
image

PR #5920, opened 2025-02-09
Updated Osmosis denoms and channels on kopi

FlatWhiteWithSugar:
Updated kopi to osmosis channels, in IBC folder and assetlist.json file

### DIFF
--- a/_IBC/kopi-osmosis.json
+++ b/_IBC/kopi-osmosis.json
@@ -2,22 +2,22 @@
     "$schema": "../ibc_data.schema.json",
     "chain_1": {
         "chain_name": "kopi",
-        "client_id": "07-tendermint-3",
-        "connection_id": "connection-3"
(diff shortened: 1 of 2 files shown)


JeremyParish69:
Generally, IBC channels cannot be changed once registered. Is there any evidence that the previously registered connection was unused (isn't abandoning assets on counterparty chains) and that the new one has seen more usage?

JeremyParish69:
It looks like there is a small amount of XKP transferred over to Osmosis. Small testing amounts may abandoned if the remaining evidence is solid
image

FlatWhiteWithSugar:
Yes those amounts was sent over to get the IBC denoms only on osmosis, they can be abandoned as they was for testing purposes only

JeremyParish69:
I see. Are we able to see how much OSMO may have been sent the other direction?

PR #5927, opened 2025-02-10
update manifest average_gas_price

chalabi2:
The average gas price was lower than the low gas price in both manfiesttestnet and manifestdevnet


JeremyParish69:
This should be easy to add to the validation PR check

PR #5930, opened 2025-02-12
Update coreum version

alinetskyi:


ankurdotb (changes requested):
The updated schema recommends a separate version file to track version upgrades and changes, not just replacing the current entry. Hence why the validation is failing here. Please fix.

PR #5943, opened 2025-02-17
add eden-boost and eden-boost logo

ollieottersync:


JeremyParish69:
It doesn't seem like this logo is being referenced from anywhere, so it's not adding value just sitting there.

ollieottersync:
Eco stake has added the token information. eden-boost is not a minted token but is used for accounting with staking as a multiplier

PR #5947, opened 2025-02-18
Update $SAM logo in Oraichain images

allomart1:
Updated $SAM logo with rounded shape one


JeremyParish69:
We require an image size < 250kB

allomart1:
Updated with size < 250 kB

PR #5958, opened 2025-02-20
Update XPLA Testnet v1.7.0-cube Version

sjyu9:
Update XPLA Testnet v1.7.0-cube Version


testnets/xplatestnet/versions.json
```diff
@@ -185,6 +185,60 @@
...
+      "cosmwasm": {
+        "version": "v0.53.0"
+      },
+      "ibc": {
+        "type": "go",
+        "version": "v8.4.0"
+      }
+    },
+    {
+      "name": "v1_7",
```
JeremyParish69, line 217: Please use a different names for v1.7.0-rc3 vs v1.7.0-cube if they really are incompatible with each other. If they are compatible, they should be merged into a single object.

sjyu9: Fixed version naming after checking comments.
You must update 'v1.7.0-cube' after updating 'v1.7.0-rc3' to ensure normal chain behaviour.

PR #5974, opened 2025-02-26
Rename $FX to $PUNDIAI

zakir-code:


fxcore/assetlist.json
```diff
@@ -2,6 +2,43 @@
...
+          "exponent": 18
+        }
+      ],
+      "base": "apundiai",
+      "name": "Pundi AI",
+      "display": "pundiai",
+      "symbol": "PUNDIAI",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/fxcore/images/PUNDIAI.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/fxcore/images/PUNDIAI.svg"
```
JeremyParish69, line 24: The svg and png must be visually identical, but it appears that the svg is missing the same white circular background that the png has (and it mostly transparent)

JeremyParish69: Thanks for the new upload but the new svg now has a square white background, while the png has a circle white background. Can you please provide an svg with a circular white background so they match?

zakir-code: Changed the background to a circle

JeremyParish69:
It looks like the fixed minimum is higher than the recommened low rate. This will need to be corrected
"fixed_min_gas_price": 5100000000,
          "low_gas_price": 5000000000,

PR #5975, opened 2025-02-26
Update dydx to v8.0

pswies:


dydx/chain.json
```diff
@@ -41,19 +41,24 @@
   },
   "codebase": {
     "git_repo": "https://github.com/dydxprotocol/v4-chain/",
-    "recommended_version": "protocol/v7.0.1",
+    "recommended_version": "protocol/v8.0.5",
```
JeremyParish69, line 44: Please all this version to the versions.json file--then checks should pass.

pswies: thanks for the tip! done

PR #5976, opened 2025-02-26
update ibc-channel info for manifest testnet

chalabi2:
This pr updates the ibc channel info for manifest and its connected chains.

### DIFF
--- a/testnets/_IBC/manifesttestnet-osmosistestnet.json
+++ b/testnets/_IBC/manifesttestnet-osmosistestnet.json
@@ -27,8 +27,8 @@
       "ordering": "unordered",
       "version": "ics20-1",
       "tags": {
-        "status": "killed",
-        "preferred": false
+        "status": "live",
+        "preferred": true
       }
     },
     {
@@ -47,8 +47,8 @@
       "ordering": "unordered",
       "version": "ics20-1",
       "tags": {
-        "status": "live",
-        "preferred": true
+        "status": "killed",
+        "preferred": false
       }
     }
   ]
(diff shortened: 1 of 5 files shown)


testnets/_IBC/manifesttestnet-osmosistestnet.json
```diff
@@ -2,28 +2,53 @@
...
-    "client_id": "07-tendermint-4314",
-    "connection_id": "connection-3774"
+    "client_id": "07-tendermint-4456",
+    "connection_id": "connection-3898"
   },
   "channels": [
+    {
+      "chain_1": {
+        "channel_id": "channel-5",
+        "port_id": "transfer",
```
JeremyParish69, line 17: there cannot be two channels where both ports are transfer registered to the same chain pair.

testnets/axelartestnet/assetlist.json
```diff
@@ -346,10 +346,10 @@
           "counterparty": {
             "chain_name": "manifesttestnet",
             "base_denom": "umfx",
-            "channel_id": "channel-3"
+            "channel_id": "channel-6"
           },
           "chain": {
-            "channel_id": "channel-590",
+            "channel_id": "channel-594",
```
JeremyParish69, line 352: with a new channel number, the `path` property must also be updated accordingly, and the `base` and `denom_unit` will also have a different hash now. This is the case for all assets affected by the change in channels

testnets/_IBC/manifesttestnet-osmosistestnet.json
chalabi2: But that state you are referring to exists on our chain 

https://nodes.liftedinit.tech/manifest/testnet/api/ibc/core/channel/v1/channels

PR #5992, opened 2025-03-02
update assetlist.json for Hood token on Osmosis

flarnrules:
Adds the Hood object to the assetlist.json on osmosis chain subdirectory.


JeremyParish69:
I see what you're going for. I'll drop this here: https://github.com/osmosis-labs/assetlists/blob/main/README.md#how-to-add-assets

PR #6000, opened 2025-03-05
Added utxo.int3 assets in neutron mainnet assetlist

DmitriyPyshinskiy:


neutron/assetlist.json
```diff
@@ -1772,6 +1772,258 @@
...
+            "channel_id": "channel-1"
+          },
+          "chain": {
+            "channel_id": "channel-6455",
+            "path": "transfer/channel-6455/factory/int31zlefkpe3g0vvm9a4h0jf9000lmqutlh99h7fsd/dogecoin-doge"
+          }
+        }
+      ],
+      "logo_URIs": {
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/doge.int3.svg",
```
JeremyParish69, line 1819: these images (at least the pngs) already exist in the int3face /images/ directory, so there's no need to upload them again.
I suggest that the svgs be added there as well, and perhaps update the doge png since it looks different there.

DmitriyPyshinskiy: I've updated images in /int3face folder and now use them for int3face as well as for neutron chain

PR #6059, opened 2025-03-24
add Airchains Varanasi Testnet to registry

kritarth1107:
This pull request adds the necessary configuration and asset information for the `Airchains Varanasi Testnet`.


kritarth1107:
Updated the chain name

PR #6065, opened 2025-03-25
Update IBC between Sidechain and others

thmadong:


thmadong:
The above IBC clients that have been changed have not been officially enabled before, so there is no problem of user funds. Because there has been no cross-chain transactions for a long time, the client has exceeded the trust period and cannot be updated normally, so a new client and channel have been recreated. After that, we will regularly perform client update operations to avoid the same problem from happening again.

PR #6066, opened 2025-03-25
Update juno testnet & mainnet

dimiandre:
Update juno testnet & mainnet with proper versions and chain-id

### DIFF
--- a/testnets/junotestnet/chain.json
+++ b/testnets/junotestnet/chain.json
@@ -38,21 +38,21 @@
     ],
     "consensus": {
       "type": "tendermint",
-      "version": "0.37.8"
+      "version": "v0.37.8"
     },
     "genesis": {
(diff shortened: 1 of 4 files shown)


testnets/junotestnet/chain.json
```diff
@@ -5,7 +5,7 @@
   "network_type": "testnet",
   "pretty_name": "Juno Testnet",
   "chain_type": "cosmos",
-  "chain_id": "uni-6",
+  "chain_id": "uni-7",
```
JeremyParish69, line 8: It looks like there are existing registered IBC connections which describe uni-6 with other chains, so we should create an archive chain for uni-6, (such as 'junotestnet6') and update its ibc files accordingly.

dimiandre: ok! i'll do it

PR #6097, opened 2025-04-04
Kyve gas step update

kamal-sutra:
Updating based on
https://github.com/chainapsis/keplr-chain-registry/pull/1085


JeremyParish69:
Thanks for helping to bring the registries in alignment!

PR #6099, opened 2025-04-04
Add lombard ledger testnet

hashxtree:


testnets/lombardledgertestnet/chain.json
```diff
@@ -0,0 +1,53 @@
+{
+  "$schema": "../../chain.schema.json",
+  "chain_name": "lombardledgertestnet",
```
JeremyParish69, line 3: I think chain_name `lombardtestnet` makes more sense to use. Then the mainnet will use chain_name `lombard`. Am I wrong?

testnets/lombardledgertestnet/assetlist.json
```diff
@@ -0,0 +1,56 @@
...
+      "denom_units": [
+        {
+          "denom": "ustake",
+          "exponent": 0
+        }
+      ],
+      "base": "ustake",
+      "name": "ustake",
+      "display": "ustake",
+      "symbol": "ustake",
```
JeremyParish69, line 52: Symbol should be all caps. (kind of looks like an asset kept from copy-pasting..)

testnets/lombardledgertestnet/assetlist.json
```diff
@@ -0,0 +1,56 @@
...
+        },
+        {
+          "denom": "lom",
+          "exponent": 6
+        }
+      ],
+      "base": "ulom",
+      "name": "Lom",
+      "display": "lom",
+      "symbol": "lom",
```
JeremyParish69, line 20: Symbol should be all caps

testnets/lombardledgertestnet/chain.json
JeremyParish69: note that how it's shown to use will go by pretty_name. (users will never see this chain_name). Also, changing what it is currently isn't required--just thought I'd point it out

JeremyParish69: > note that how it's shown to use will go by pretty_name (Where is `"pretty_name"`?). (users will never see this chain_name). Also, changing what it is currently isn't required--just thought I'd point it out

russanto: Checked internally, `lombardledgertestnet` is ok for us. I added the `pretty_name` field

testnets/lombardledgertestnet/assetlist.json
russanto: updated

PR #6104, opened 2025-04-05
make logo round

Nostradamus411:


JeremyParish69:
This would be nice to have, but it would make the svg and png disagree, and they are currently both defined as being the same image, so they but be visually identical

Nostradamus411:
@JeremyParish69 synced up the SVG & PNG so they're both round from the atomone-hub/assets repo

PR #6117, opened 2025-04-08
Updating Jackal Axelar IBC Channels

TheMarstonConnell:
Reverting to other channels now that clients are open again.

### DIFF
--- a/_IBC/axelar-jackal.json
+++ b/_IBC/axelar-jackal.json
@@ -13,11 +13,11 @@
     "channels": [
       {
         "chain_1": {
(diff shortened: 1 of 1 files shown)


JeremyParish69:
IBC channels cannot be changed without either: governance approval in both chains, or proof that there would be no abandoned  tokens on counterparty chains. Is there a URL endpoint that can show the balances of the ICS20 version of the main assets on each of these chains?

TheMarstonConnell:
Nothing closed, we just re-opened a channel that expired.

https://axelarscan.io/proposal/298

PR #6119, opened 2025-04-09
Add Lombard Ledger mainnet

russanto:
- Adds Lombard Ledger mainnet
- Updates Lombard Ledger testnet to latest binary


dylanschultzie:
@russanto please add Lombard to the slip-173 registry: https://github.com/satoshilabs/slips/blob/master/slip-0173.md

russanto:
Is it strictly necessary ? LOM will not be a public token.

dylanschultzie:
Good point, I think in that case it's not necessary.

dylanschultzie (approved):

PR #6140, opened 2025-04-10
Update Neutron version

toschdev:
With Proposal https://governance.neutron.org/proposals/A/65 Neutron Mercury version 6.0.0 came into effect today.


JeremyParish69:
Note that v5.0.2 is still the recommended version in the chain.json

PR #6168, opened 2025-04-16
Update chain.json

kleomedes:
Removed all expired RPCs, APIs and gRPCS. Also removed all inactive ping.pub forks from explorers section.

### DIFF
--- a/juno/chain.json
+++ b/juno/chain.json
@@ -191,57 +191,17 @@
         "provider": "Lavender.Five Nodes \ud83d\udc1d"
       },
       {
-        "address": "https://rpc-juno.pupmos.network",
-        "provider": "PUPM\u00d8S"
-      },
-      {
-        "address": "https://rpc-juno.cosmos-spaces.cloud",
-        "provider": "Cosmos Spaces"
-      },
-      {
-        "address": "https://rpc-juno-ia.cosmosia.notional.ventures/",
-        "provider": "Notional"
-      },
-      {
         "address": "https://juno-rpc.kleomedes.network",
         "provider": "Kleomedes"
-      },
-      {
-        "address": "https://juno-rpc.icycro.org",
-        "provider": "IcyCRO \ud83e\uddca"
-      },
-      {
-        "address": "https://rpc.juno.bh.rocks",
-        "provider": "BlockHunters \ud83c\udfaf"
-      },
-      {
-        "address": "https://juno-rpc.kingnodes.com",
-        "provider": "kingnodes \ud83d\udc51"
-      },
-      {
-        "address": "https://juno-rpc.reece.sh",
-        "provider": "Reecepbcups"
       },
       {
         "address": "https://juno-rpc.stakeandrelax.net",
         "provider": "Stake&Relax Validator \ud83e\udda5"
       },
       {
-        "address": "https://rpc-juno.architectnodes.com",
-        "provider": "Architect Nodes"
-      },
-      {
         "address": "https://rpc-juno-01.stakeflow.io",
         "provider": "Stakeflow"
-      },
-      {
-        "address": "https://rpc-juno.validavia.me",
-        "provider": "Validavia"
-      },
-      {
-        "address": "https://juno-rpc.w3coins.io",
-        "provider": "w3coins"
-      },
+      }, 
       {
         "address": "https://juno-rpc.publicnode.com:443",
         "provider": "Allnodes \u26a1\ufe0f Nodes & Staking"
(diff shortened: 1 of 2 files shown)


juno/chain.json
```diff
@@ -372,102 +268,38 @@
...
-        "address": "https://juno.declab.pro:9007",
-        "provider": "Decloud Nodes Lab"
-      },
-      {
-        "address": "https://juno.grpc.nodeshub.online",
-        "provider": "Nodes Hub"
-      },
-      {
-        "address": "juno-grpc.chainroot.io:443",
+        "address": "g",
```
JeremyParish69, line 302: this doesn't look right

juno/chain.json
```diff
@@ -496,38 +328,14 @@
...
-      "account_page": "https://explorer.stavr.tech/juno/accounts/${accountAddress}"
-    },
     {
       "kind": "TC Network",
       "url": "https://explorer.tcnetwork.io/juno",
       "tx_page": "https://explorer.tcnetwork.io/juno/transaction/${txHash}"
     },
     {
-      "kind": "Stakeflow",
-      "url": "https://stakeflow.io/juno",
```
JeremyParish69, line 512: This one appears active??

juno/chain.json
```diff
@@ -496,38 +328,14 @@
...
-      "account_page": "https://stakeflow.io/juno/accounts/${accountAddress}"
-    },
-    {
-      "kind": "Decloud Nodes Lab",
-      "url": "https://explorer.declab.pro/Juno",
```
JeremyParish69, line 517: This one appears active

juno/chain.json
```diff
@@ -274,76 +222,32 @@
...
         "provider": "Chainroot"
       }
     ],
     "rest": [
-      {
-        "address": "https://lcd-juno.itastakers.com",
-        "provider": "itastakers"
-      },
-      {
-        "address": "https://rest.lavenderfive.com:443/juno",
```
JeremyParish69, line 292: This appears to be working

JeremyParish69:
How did you collect this data? Some appear to be valid to remove, but others are being removed even though it's working.

juno/chain.json
kleomedes: image
Doesnt look like it?

juno/chain.json
```diff
@@ -496,38 +328,14 @@
       "tx_page": "https://atomscan.com/juno/transactions/${txHash}",
       "account_page": "https://atomscan.com/juno/accounts/${accountAddress}"
     },
-    {
-      "kind": "\ud83d\udd25STAVR\ud83d\udd25",
-      "url": "https://explorer.stavr.tech/juno",
```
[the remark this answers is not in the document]
kleomedes: We have proposed removal of all ping.pub forks that are not unique as this doesn't add value to the existing explorers.

juno/chain.json
kleomedes: This one is active, will fix

kleomedes:
> How did you collect this data? Some appear to be valid to remove, but others are being removed even though it's working.

I went thru all RPCs and APIs manually. Also checked all gRPC responsiveness via CLI. Explorers have been re-added per Dimi's request. I did a second check with the remaining active ones on ChatGPT to confirm liveness.

juno/chain.json
JeremyParish69: going to a specific endpoint appears to work, though.
https://rest.lavenderfive.com/juno/cosmos/base/tendermint/v1beta1/node_info

juno/chain.json
```diff
@@ -352,18 +256,10 @@
         "address": "https://juno-api.stake-town.com",
         "provider": "StakeTown"
       },
-      {
-        "address": "https://juno-api.stakeandrelax.net",
```
JeremyParish69, line 356: This one also works
https://juno-api.stakeandrelax.net/cosmos/base/tendermint/v1beta1/node_info

kleomedes: So that one was a duplicate, it's already in there! Stake and relax had several duplicates for various infra.
Screenshot 2025-04-17 at 1 12 01 PM

JeremyParish69: You're right. I'll add a check for duplicates eventually

juno/chain.json
JeremyParish69: Do you agree this should stay seeing that endpoints work (I'm not sure if it's expected that it goes to something like a swagger when no endpoint is specified)

JeremyParish69:
@kleomedes Please see that one last remaining unresolved comment (then we should be good to proceed)

juno/chain.json
kleomedes: Makes sense, can keep it!

PR #6170, opened 2025-04-17
Update clients, channels and assets between Cosmoshub and Kopi

dev-kopi:
This is a PR for a new client to connect to the hub. We are in touch with the Interchain team and they have agreed for us to go down this route to expedite some things we are working on.

The old client expired and we confirm there are no funds stuck on the previous route.

### DIFF
--- a/_IBC/cosmoshub-kopi.json
+++ b/_IBC/cosmoshub-kopi.json
@@ -2,22 +2,22 @@
     "$schema": "../ibc_data.schema.json",
     "chain_1": {
         "chain_name": "cosmoshub",
-        "client_id": "07-tendermint-1332",
-        "connection_id": "connection-1065"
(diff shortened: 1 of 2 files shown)


_IBC/cosmoshub-kopi.json
```diff
@@ -2,22 +2,22 @@
...
-        "client_id": "07-tendermint-4",
-        "connection_id": "connection-8"
+        "client_id": "07-tendermint-19",
+        "connection_id": "connection-43"
     },
     "channels": [
         {
             "chain_1": {
-                "channel_id": "channel-1253",
+                "channel_id": "channel-1351",
```
JeremyParish69, line 16: The old channel doesn't appear to be used much, but is there evidence that there is no abandoned supply of ATOM on Kopi?

image

PR #6186, opened 2025-04-22
add OtterSync endpoints and Kwak Elysscan explorer to elys chain.json

ollieottersync:


elys/chain.json
```diff
@@ -473,5 +485,11 @@
       "kind": "Stake-Hub by Kleomedes",
       "url": "https://www.stake-hub.xyz/elys"
     }
+        {
```
JeremyParish69, line 488: please fix the commas

ollieottersync: Done sorry about that

PR #6197, opened 2025-04-24
chain registry UI

faddat:
This is the re-envisioning of the chian registry as a consumer desktop application that lets people state sync nodes.  It's pretty good now, but needs a little spit and polish to be perfect.


pkg/chainregistry/chainregistry.go
```diff
@@ -0,0 +1,197 @@
...
+// ChainImage represents an image for the chain
+type ChainImage struct {
+	PNG   string     `json:"png,omitempty"`
+	SVG   string     `json:"svg,omitempty"`
+	Theme ImageTheme `json:"theme,omitempty"`
+}
+
+// ImageTheme represents the theme for an image
+type ImageTheme struct {
+	PrimaryColorHex string `json:"primary_color_hex"`
```
JeremyParish69, line 93: fyi, there also exists "circle" boolean and "background_color_hex"

JeremyParish69:
This is quite interesting. Is there a way to experiment with this in a separate repo (or even a branch) without making such a drastic change to the Cosmos Chain Registry as it is currently?

(Note, I'm not very fluent in Go, so it might take me a while to fully comprehend how this works)

scutuatua-crypto (approved):

PR #6210, opened 2025-04-27
Update juno testnet & mainnet

cmancrypto:
Update juno testnet & mainnet with proper versions and chain-id and archival of uni-6


JeremyParish69:
Well done! Archiving a chain is tricky for most, so glad to see you got it

cmancrypto:
Practice makes perfect! Plus Avi sent me through your instructions, so that helped a great deal. Thank you!

PR #6237, opened 2025-05-05
Register Coreum v5 release

alinetskyi:


coreum/chain.json
```diff
@@ -37,13 +37,13 @@
   },
   "codebase": {
     "git_repo": "https://github.com/CoreumFoundation/coreum",
-    "recommended_version": "v4.1.2",
+    "recommended_version": "v5.0.0",
```
JeremyParish69, line 40: The same changes must be reflected in the versions.json file

alinetskyi: Fixed

PR #6239, opened 2025-05-05
Add hINJ and Hydro Wrapped hINJ assets to assetlist.json

guardianofthesea:
Add asset info for cw20 variant and factory wrapped variant of hINJ - Hydro Staked INJ


injective/assetlist.json
```diff
@@ -1458,6 +1458,70 @@
...
+        {
+          "denom": "factory/inj14ejqjyq8um4p3xfqj74yld5waqljf88f9eneuk/inj18luqttqyckgpddndh8hvaq25d5nfwjc78m56lc",
+          "exponent": 18
+        },
+        {
+          "denom": "Hydro Wrapped hINJ",
+          "exponent": 18
+        }
+      ],
+      "base": "factory/inj14ejqjyq8um4p3xfqj74yld5waqljf88f9eneuk/inj18luqttqyckgpddndh8hvaq25d5nfwjc78m56lc",
```
JeremyParish69, line 1501: I would assume that this factory token is the replacement, and so the original would get the "legacy-mintage" trace on it instead. It's like saying "this asset is a legacy mintage (deprecated) of this other token (defined in counterparty)." If so, please relocate it to the other asset.

injective/assetlist.json
```diff
@@ -1458,6 +1458,70 @@
...
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/injective/images/hinj.png"
+        }
+      ]
+    },
+    {
+      "description": "Factory Wrapped hINJ - Hydro Staked INJ",
+      "denom_units": [
+        {
+          "denom": "factory/inj14ejqjyq8um4p3xfqj74yld5waqljf88f9eneuk/inj18luqttqyckgpddndh8hvaq25d5nfwjc78m56lc",
```
JeremyParish69, line 1493: base denom must have 0 exponent

guardianofthesea: fixed

injective/assetlist.json
```diff
@@ -1458,6 +1458,70 @@
         "twitter": "https://x.com/instantnoodlesinc"
       },
       "type_asset": "sdk.coin"
+    },
+    {
+      "description": "CW20 hINJ - Hydro Staked INJ",
+      "denom_units": [
+        {
+          "denom": "cw20:inj18luqttqyckgpddndh8hvaq25d5nfwjc78m56lc",
+          "exponent": 18
```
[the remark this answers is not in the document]
guardianofthesea: fixed

injective/assetlist.json
guardianofthesea: Both the factory variant and the cw20 variant of hINJ are in full usage currently. The CW20 version has not been depreciated, but still as your comment suggests the factory version is the successor of the CW20 version. So I have made your recommended changes.

If there is a better way to label these, let me know.

JeremyParish69: I see. In that case if would have been acceptable to either do what you just did, or to label to successor as an `additional-mintage` until the original is finally deprecated, and then come back to update. But with what you've done, it's fine; it might imply that the original has already been replaced, but we do also accepted a `deprecated` boolean property which is used for actually doing so. I think it's fine as-is.

PR #6243, opened 2025-05-07
Update bech32 prefix from 'side' to 'bc'

thmadong:


thmadong:
Our chain supports both cosmos addresses starting with side and bitcoin addresses starting with bc. We hope to mainly recommend using bc addresses.

PR #6324, opened 2025-06-03
Provenance Rebrand to Hastra

dcshock:
Provenance is in the beginning stages of rebranding as Hastra as part of our overall GTM. Until the process is complete it makes sense to list in the registry as both Provenance and Hastra so that downstream libraries that use the registry can find the appropriate endpoints based on looking up either name. 

This mostly a copy from the Provenance config with some minor updates around name and brand.


dcshock:
@JeremyParish69 - that's good info. Let me update the Provenance files with the new Hastra branding, and push that up. I'll also include a link showing the rebrand from the www.provenance.io website so you have something to approve against. If we need to take this further in the future with an actual rename of the chain, I'll circle back then.

provenance/assetlist.json
```diff
@@ -19,15 +19,15 @@
...
-        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/provenance/images/prov.svg"
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/hastra/images/hastra.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/hastra/images/hastra.svg"
       },
       "images": [
         {
-          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/provenance/images/prov.png",
-          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/provenance/images/prov.svg",
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/hastra/images/hastra.png",
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/hastra/images/hastra.svg",
```
JeremyParishTheBuilder, line 28: Sorry to nit pick, but these images have different amounts of padding, which will cause them to look different based on whether the svg vs png is used. Either these should be separated, or better yet ensure to use images with identical padding.

dcshock: Not an issue. I'll get updated images and get those submitted. Appreciate the review.

PR #6332, opened 2025-06-05
Add mBTC and derivatives to the list of Pryzm assets

madrezaz:
This PR adds new mBTC derivatives on Pryzm chain.


_non-cosmos/ethereum/assetlist.json
```diff
@@ -2520,6 +2520,29 @@
...
+        },
+        {
+          "denom": "mBTC",
+          "exponent": 18
+        }
+      ],
+      "base": "0x007115416ab6c266329a03b09a8aa39ac2ef7d9d",
+      "name": "Midas BTC Yield Token",
+      "display": "mBTC",
+      "symbol": "mBTC.midas",
```
JeremyParishTheBuilder, line 2540: why does the symbol have '.midas'? I assume that's what the prefix 'm-' stands for.

JeremyParishTheBuilder: other than this question it looks like this PR is acceptable.

madrezaz: The reason is a name collision with [another mBTC](https://github.com/cosmos/chain-registry/blob/master/_non-cosmos/ethereum/assetlist.json#L2151) which is a Babylon liquid staking token.
I followed the pattern used in [Skip Go's registry](https://github.com/skip-mev/skip-go-registry/blob/main/chains/1/assetlist.json#L292)

JeremyParishTheBuilder: I understand. Not sure that I agree with it being different from what CoinGecko has, but I see why it's there.

PR #6367, opened 2025-06-19
Update Paloma chain.json and versions.json Files 

taariq:
Update Paloma chain.json and versions.json Files to current Paloma releases, cosmos-SDK, IBC, and binaries.

### DIFF
--- a/paloma/versions.json
+++ b/paloma/versions.json
@@ -3,25 +3,25 @@
   "chain_name": "paloma",
   "versions": [
     {
-      "name": "v1.15.4",
-      "recommended_version": "v1.15.4",
+      "name": "v2.4.12",
+      "recommended_version": "v2.4.12",
       "compatible_versions": [
-        "v1.15.4"
+        "v2.4.12"
       ],
       "consensus": {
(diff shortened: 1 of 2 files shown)


paloma/versions.json
```diff
@@ -3,25 +3,25 @@
   "chain_name": "paloma",
   "versions": [
     {
-      "name": "v1.15.4",
-      "recommended_version": "v1.15.4",
+      "name": "v2.4.12",
```
JeremyParishTheBuilder, line 6: Instead of writing over v1.15.4, please add a new version object after v1.15.4

taariq: Thanks @JeremyParish69. What is the format to submit a new version object? I tried looking at other submissions, but was not sure.

JeremyParishTheBuilder: In the versions.json file, there is an array property (called "versions"). Inside that array is an object (was named "v1.15.4). Just copy and past that object below it (so there's two objects now in the 'versions' array). Then update the second object to the details for v2.4.12 (but leave the first object as v1.15.4).

PR #6369, opened 2025-06-20
remove outdated rpcs from juno

dimiandre:


juno/chain.json
```diff
@@ -186,10 +186,6 @@
         "address": "https://juno-rpc.polkachu.com",
         "provider": "Polkachu"
       },
-      {
-        "address": "https://rpc.lavenderfive.com:443/juno",
```
JeremyParishTheBuilder, line 190: This appears to still be active when visited. Why is being removed?

PR #6375, opened 2025-06-23
feat: add Buycex Infinity Chain (BCX, chainId 323)

cosvmlabs:


cosvm/chain.json
```diff
@@ -0,0 +1,674 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "evmos",
```
JeremyParishTheBuilder: This file is clearly just copy and pasted from evmos. Much needs to be corrected here

cosvm/assetlist.json
```diff
@@ -0,0 +1,34 @@
...
+        }
+      ],
+      "base": "ucvm",
+      "name": "CosVM",
+      "display": "cvm",
+      "symbol": "CVM",
+      "coingecko_id": "cosvm",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/cosvm/images/cvm.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/cosvm/images/cvm.svg"
```
JeremyParishTheBuilder: Please do not upload false SVG (a png exported as a svg)--as it is not scalable.

buycex/assetlist.json
```diff
@@ -0,0 +1,27 @@
...
+        }
+      ],
+      "base": "ubcx",
+      "name": "Buycex",
+      "display": "BCX",
+      "symbol": "BCX",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/buycex/images/bcx.png"
+      },
+      "coingecko_id": "",
```
JeremyParishTheBuilder: Please remove empty fields like this

cosvmlabs (commented):
updated

cosvm/chain.json
cosvmlabs: deleted

cosvm/assetlist.json
cosvmlabs: deleted

buycex/assetlist.json
```diff
@@ -0,0 +1,26 @@
+{
```
JeremyParishTheBuilder, line 1: all three files are missing the (required) schema property

cosvmlabs: required field are updated. thanks

buycex/chain.json
```diff
@@ -0,0 +1,57 @@
+{
```
JeremyParishTheBuilder, line 1: chain_type is required

buycex/versions.json
```diff
@@ -0,0 +1,19 @@
+{
+  "$schema": "../versions.schema.json",
+  "chain_name": "buycex",
+  "versions": [
+    {
+      "name": "v1.0.0",
+      "recommended_version": true,
```
JeremyParishTheBuilder: recommended_version refers to the specific version (can be a minor or patch version) among a set of versions that are all compatible with v1.0.0. so "recommended_version" should be something like `"v1.0.0"`, and "compatible_versions": would be
```
[
  "v1.0.0"
]   
```

buycex/versions.json
```diff
@@ -0,0 +1,19 @@
...
+      "cosmos_sdk_version": "v0.45.16",
+      "builds": [
```
JeremyParishTheBuilder: this "builds" property is not accepted, as it does not follow the schema, however we do allow urls for "binaries". (see [schema](https://github.com/cosmos/chain-registry/blob/master/versions.schema.json#L226) for details)

cosvmlabs:
required field are updated. thanks

buycex/chain.json
```diff
@@ -0,0 +1,58 @@
...
+  "chain_name": "buycex",
+  "chain_id": "bcx_323-1",
+  "pretty_name": "BuyCex Infinity Chain",
+  "status": "live",
+  "network_type": "mainnet",
+  "website": "https://infinity.buycex.com",
+  "bech32_prefix": "bcx",
+  "daemon_name": "bcxd",
+  "node_home": "$HOME/.bcxd",
+  "chain_type": "cosmos-sdk",
```
JeremyParishTheBuilder: Per the schema, cosmos-sdk is not an accepted chain type, although "cosmos" is allowed

buycex/versions.json
JeremyParishTheBuilder: this was not resolved

cosvmlabs: resolved

buycex/chain.json
cosvmlabs: "cosmos"  updated

cosvmlabs:
required data updated

buycex/chain.json
```diff
@@ -0,0 +1,58 @@
+{
```
JeremyParishTheBuilder, line 1: oh, slip_44 is also required

PR #6378, opened 2025-06-24
feat: add symphony mainnet

dblackstone1:


symphony/chain.json
```diff
@@ -0,0 +1,106 @@
+{
+  "$schema": "../chain.schema.json",
+  "chain_name": "symphonytestnet",
+  "status": "live",
+  "network_type": "testnet",
+  "pretty_name": "Symphony Testnet",
```
JeremyParishTheBuilder: Was this just copy and pasted and now reviewed? It says "testnet" and "Symphony Testnet" but being put into a directory like it's a mainnet

symphony/chain.json
```diff
@@ -0,0 +1,106 @@
...
+        "provider": "Nodes Hub"
+      },
+      {
+        "address": "https://symphony-grpc.cogwheel.zone:443",
+        "provider": "Cogwheel \u2699\ufe0f"
+      }
+    ]
+  },
+  "logo_URIs": {
+    "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/testnets/symphonytestnet/images/symphony_logo.png"
```
JeremyParishTheBuilder: Now that there is a mainnet, please relocate the images from the /testnets/symphonytestnet/ directory and just have them in the (mainnet) /symphony/ directory so that we don't have duplicate uploads of the images.

testnets/symphonytestnet3/chain.json
```diff
@@ -1,135 +0,0 @@
-{
```
JeremyParishTheBuilder, line 1: why is symphonytestnet3 being deleted?

dblackstone1: that should have never been up since we moved to 4.  don't know who put it up, but we're not running a testnet on v3 anymore.  we'll also be scaling down 4 since mainnet is going up

symphony/chain.json
dblackstone1: changed

symphony/chain.json
dblackstone1: changed

symphony/chain.json
```diff
@@ -0,0 +1,118 @@
...
+  "staking": {
+    "staking_tokens": [
+      {
+        "denom": "note"
+      }
+    ]
+  },
+  "codebase": {
+    "git_repo": "https://github.com/Orchestra-Labs/symphony",
+    "recommended_version": "v0.4.1",
```
JeremyParishTheBuilder: this version cannot be located anywhere in the versions file, but must be defined there

symphony/versions.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$schema": "../versions.schema.json",
+  "chain_name": "symphony",
+  "versions": [
+    {
+      "name": "v1.0.1",
```
JeremyParishTheBuilder: It's quite odd (misleading) to have a version name of v1.0.1, but not have that listed as a compatible version...

dblackstone1: changed to 1.0.0

symphony/chain.json
dblackstone1: changed to 1.0.0

PR #6404, opened 2025-07-04
fixed take titan chain configurations 

take-titan:
fixed take titan chain configurations as previous details was completely not aligned with running chain

### DIFF
--- a/taketitan/chain.json
+++ b/taketitan/chain.json
@@ -6,18 +6,18 @@
   "website": "https://taketitan.com/",
   "pretty_name": "TakeTitan",
   "chain_type": "cosmos",
-  "chain_id": "taketitan-12",
-  "bech32_prefix": "taketitan",
-  "daemon_name": "ttnc",
-  "node_home": "$HOME/.ssc",
+  "chain_id": "titan-1",
+  "bech32_prefix": "titan",
+  "daemon_name": "titand",
+  "node_home": "$HOME/.titand",
   "key_algos": [
     "secp256k1"
   ],
-  "slip44": 1179993421,
+  "slip44": 118,
   "staking": {
     "staking_tokens": [
       {
-        "denom": "ttnc"
+        "denom": "utitan"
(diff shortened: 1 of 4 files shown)


taketitan/chain.json
```diff
@@ -6,18 +6,18 @@
...
   "key_algos": [
     "secp256k1"
   ],
-  "slip44": 1179993421,
+  "slip44": 118,
   "staking": {
     "staking_tokens": [
       {
-        "denom": "ttnc"
+        "denom": "utitan"
```
JeremyParishTheBuilder, line 20: An asset with this 'base' must be defined (which currently it's still set to ttnc in the assetlist)

take-titan: thanks @JeremyParish69  for pointing this out. I fixed assets to match chain configuration

taketitan/chain.json
```diff
@@ -6,18 +6,18 @@
   "website": "https://taketitan.com/",
   "pretty_name": "TakeTitan",
   "chain_type": "cosmos",
-  "chain_id": "taketitan-12",
-  "bech32_prefix": "taketitan",
-  "daemon_name": "ttnc",
-  "node_home": "$HOME/.ssc",
+  "chain_id": "titan-1",
```
JeremyParishTheBuilder: Unfortunately the chain-id "titan-1" is already taken by rizon. I'd recommend upgrading to a different chain-id prefix.

JeremyParishTheBuilder: Also, a new chain should require that taketitan-12 be gracefully archived (say, using chain_name: "taketitan1", and then this new chain can use the original chain_name: "taketitan")

take-titan: Thank you. I changed the chain_id to taketitan-1 and update the submission accordingly.
I also archived the previous taketitan-12 entry under a different "chain_name": "taketitan1" and marked it as "status": "killed".

PR #6408, opened 2025-07-08
Upgrade Cosmos ICS Testnet to Gaia v25.0.0

dasanchez:


testnets/cosmosicsprovidertestnet/chain.json
```diff
@@ -19,7 +19,7 @@
         "denom": "uatom",
         "fixed_min_gas_price": 0.005,
         "low_gas_price": 0.01,
-        "average_gas_price": 0.025,
+        "average_gas_price": 0.005,
```
JeremyParishTheBuilder: Average gas price cannot be less than low gas price. Please ensure gas prices are ascending from low to avg to high

dasanchez: Thanks @JeremyParish69 ! Fixed.

PR #6423, opened 2025-07-15
Update SEDA testnet and mainnet to v1.0.0-rc.6 and v1.0.0, respectively

hacheigriega:


testnets/sedatestnet/chain.json
```diff
@@ -58,7 +58,7 @@
         "version": "v0.50.13"
       },
       "cosmwasm": {
-        "version": "0.53.2",
+        "version": "0.53.3",
```
JeremyParishTheBuilder: Looks like this has to be v0.53.3 to match the version in the "versions.json" file

`Path: cosmwasm.version, Reason: Value mismatch, In "codebase": 0.53.3, In currentVersion: v0.53.3`

hacheigriega: fixed, thanks

PR #6458, opened 2025-07-23
feat: add rji to thorchain

jpKuji:
Add "The Rujira Index" token to the THORChain asset list.


thorchain/assetlist.json
```diff
@@ -193,6 +193,39 @@
...
+      "base": "rji",
+      "type_asset": "sdk.coin",
+      "description": "The Rujira Index (RJI) is a fixed unit asset index that tracks the performance of tokens on the Rujira App Layer. RJI does not auto-rebalance, therefore by design continously overweighting the top performers and underweighting the under-performers.",
+      "denom_units": [
+        {
+          "denom": "rji",
+          "exponent": 0
+        },
+        {
+          "denom": "x/nami-index-fixed-thor1yqf5spdv8c4088zmvqsg32eq63fzepsjvntahdk0ek0yjnkt3qdqftp3lc-rcpt",
```
JeremyParishTheBuilder: I suspect the base denom (exponent 0) is x/nami-index-fixed-thor1yqf5spdv8c4088zmvqsg32eq63fzepsjvntahdk0ek0yjnkt3qdqftp3lc-rcpt and that "rji" should have exponent 8...

jpKuji: Yes you're right. Thanks for pointing it out. Will put up another MR to correct the other assets inside thorchain.

JeremyParishTheBuilder: In which case the `base` must also be corrected (not 'rji')

jpKuji: Got it. Thanks again.

PR #6463, opened 2025-07-28
fix(thorchain): add sNAMI and fix other assets denom_units

jpKuji:
Updated remaining Thorchain assets based on feedback from the previous PR. Also appended sNAMI to the end of the array.

### DIFF
--- a/thorchain/assetlist.json
+++ b/thorchain/assetlist.json
@@ -281,6 +281,16 @@
       "name": "Staked Nami",
       "display": "sNAMI",
       "symbol": "sNAMI",
+      "traces": [
+        {
+          "type": "liquid-stake",
+          "counterparty": {
+            "chain_name": "thorchain",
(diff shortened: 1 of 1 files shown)


thorchain/assetlist.json
```diff
@@ -263,6 +263,38 @@
...
+          "exponent": 0
+        },
+        {
+          "denom": "sNAMI",
+          "exponent": 8
+        }
+      ],
+      "name": "Staked Nami",
+      "display": "sNAMI",
+      "symbol": "sNAMI",
```
JeremyParishTheBuilder, line 283: For assets with lowercase letters in the symbol, please either register it's relation (via `traces`) to its origin (NAMI) (preferred), or else convert to all UPPERCASE (SNAMI).

jpKuji: Addressed. Can you check?

PR #6471, opened 2025-08-01
update new Terra2 SOLID image 

charleskimbest:
add Terra2 new CAPA image and update SOLID image


charleskimbest:
I've removed CAPA images and chaged solid.png under the 250kB size
I will update new CAPA images in the other PR, #6472

PR #6476, opened 2025-08-01
Add Sunrise Mainnet & Testnet

Senna46:
Sunrise Testnet is already live and the mainnet will be live on Aug 4th. 
We would like to merge them by then.


sunrise/versions.json
```diff
@@ -3,81 +3,25 @@
...
       "ibc": {
         "type": "go",
-        "version": "8.0.0",
-        "ics_enabled": [
-          "ics20-1"
-        ]
-      }
-    },
-    {
-      "name": "v0.0.8",
```
JeremyParishTheBuilder, line 33: Why delete the version history?

Senna46: This is the first release of sunrise's mainnet. We deem it unnecessary to provide information on binaries that have not been used in the past.

Senna46: Any future upgrades will be added here. we would like to clarify that geneisis is v1.0.0.

PR #6484, opened 2025-08-05
Add IBC channels: Bitsong ↔ Axelar and Archway ↔ Juno

Validatrium:
Add new relays

Bitsong ↔ Axelar

Archway ↔ Juno

### DIFF
--- a/_IBC/bitsong-axelar.json
+++ b/_IBC/bitsong-axelar.json
@@ -12,9 +12,14 @@
   },
   "channels": [
     {
-      "chain_1_channel_id": "channel-38",
-      "chain_2_channel_id": "channel-183",
-      "port_id": "transfer",
+      "chain_1": {
+        "channel_id": "channel-38",
+        "port_id": "transfer"
+      },
+      "chain_2": {
+        "channel_id": "channel-183",
+        "port_id": "transfer"
+      },
       "ordering": "unordered",
       "version": "ics20-1",
       "tags": {


_IBC/archway-juno.json
```diff
@@ -2,28 +2,29 @@
...
   },
   "channels": [
     {
       "chain_1": {
-        "channel_id": "channel-15", 
+        "channel_id": "channel-208",
         "port_id": "transfer"
       },
       "chain_2": {
-        "channel_id": "channel-290",
```
JeremyParishTheBuilder, line 20: It looks like this old channel was used, and so it may be that case that tokens already transferred would be abandoned if a new main IBC channel is registered to replace the current one. Are we able to show that no (or very few--testing) tokens has been abandoned on either chain?

_IBC/bitsong-axelar.json
```diff
@@ -0,0 +1,26 @@
...
+    "connection_id": "connection-57"
+  },
+  "chain_2": {
+    "chain_name": "axelar",
+    "client_id": "07-tendermint-252",
+    "connection_id": "connection-243"
+  },
+  "channels": [
+    {
+      "chain_1_channel_id": "channel-38",
```
JeremyParishTheBuilder: Not quite following the schema. chain_1 would be an object, and channel_id would be a property within

_IBC/archway-juno.json
Validatrium: Yes you can

_IBC/bitsong-axelar.json
Validatrium: My bad
I create new PR with right config file

_IBC/archway-juno.json
JeremyParishTheBuilder: Please provide this evidence

angelorc:
Hi @JeremyParish69 can you merge this pr? The old channel is expired, we checked the btsg supply on axelar and there is only few test token there, this is why we opened a new channel instead of re-active the old channel

_IBC/bitsong-axelar.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "bitsong",
```
JeremyParishTheBuilder, line 4: The chain order is opposite of alphabetical, causing checks to fail

PR #6487, opened 2025-08-06
Fix IBC channel metadata for Bitsong ↔ Axelar and Archway ↔ Juno

Validatrium:
Fixed Bitsong ↔ Axelar config file

### DIFF
--- a/_IBC/bitsong-axelar.json
+++ b/_IBC/bitsong-axelar.json
@@ -1,23 +1,23 @@
 {
   "$schema": "../ibc_data.schema.json",
   "chain_1": {
-    "chain_name": "archway",
-    "client_id": "07-tendermint-121",
-    "connection_id": "connection-119"
+    "chain_name": "bitsong",
+    "client_id": "07-tendermint-82",
+    "connection_id": "connection-57"
   },
   "chain_2": {
-    "chain_name": "juno",
-    "client_id": "07-tendermint-583",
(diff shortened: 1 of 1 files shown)


_IBC/archway-juno.json
```diff
@@ -2,28 +2,29 @@
...
-    "client_id": "07-tendermint-387",
-    "connection_id": "connection-379"
+    "client_id": "07-tendermint-583",
+    "connection_id": "connection-550"
   },
   "channels": [
     {
       "chain_1": {
-        "channel_id": "channel-15", 
+        "channel_id": "channel-208",
```
JeremyParishTheBuilder, line 16: Please show evidence token balances from the old channel (need to make sure this update isn't abandoning transferred tokens)

JeremyParishTheBuilder: It looks like there are no ARCH on Juno, but there is a small amount of JUNO on Archway (but small enough that we can proceed)
image

_IBC/bitsong-axelar.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "bitsong",
```
JeremyParishTheBuilder, line 4: chain order still needs to be corrected (alphabetical order)

PR #6496, opened 2025-08-07
Update chain.json for dymension testnet

maxzonder:


testnets/dymensiontestnet1/chain.json
```diff
@@ -2,11 +2,11 @@
   "$schema": "../../chain.schema.json",
   "chain_name": "dymensiontestnet1",
   "chain_type": "cosmos",
-  "chain_id": "dymension_1405-1",
+  "chain_id": "dymension_3405-1",
   "pretty_name": "Dymension Playground Testnet",
   "status": "live",
   "network_type": "testnet",
-  "website": "https://playground.dymension.xyz",
+  "website": https://dymension-internal-devnet.web.app/rollapps",
```
JeremyParishTheBuilder: missing the initial double quote.
Also, the URL looks a little unusual; is there any official documentation that points to this URL?

maxzonder: URL is correct, @omritoptix can cofirm

maxzonder: ok my mistake, reverted back the URL, so the changes are chain-id and rpcs @JeremyParish69

PR #6517, opened 2025-08-14
Update ampCAPA.png

charleskimbest:
Update ampCAPA.png by changing the center white color from RGBA(255, 255, 255, 100) to RGBA(254, 254, 254, 100)


charleskimbest:
SVG looks fine. The issue is with the PNG — even though alpha is 100, Keplr renders the center white as transparent, showing the background. Fix is to slightly tweak the RGB values.

PR #6523, opened 2025-08-17
Fixed CWH Project PR

iSkitzTV:
I believe I've Fixed the Issues. I added the wrong information/PR on the Last one

### DIFF
--- a/terra/assetlist.json
+++ b/terra/assetlist.json
@@ -6510,8 +6510,8 @@
           "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/noble/images/USDCoin.svg"
         }
       ]
-    }
-  ],
+    },
+    {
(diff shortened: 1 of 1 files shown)


iSkitzTV:
I think this should all now be fixed. Sorry for the Inconvenience and Issues. I wasn't sure about "You need separate asset objects (currently all the properties are mixed into a single object" So any help fixing that if its still an issue would be greatly appreciated

terra/assetlist.json
```diff
@@ -6511,5 +6511,61 @@
...
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/terra/images/ampcwh.png"
+      },
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/terra/images/ampcwh.png"
+        }
+      ]
+    }
+  ],
```
JeremyParishTheBuilder: There is a schema error here.

PR #6536, opened 2025-08-25
atomone-testnet-1 - upgrade v3.0.1

albttx:
Upgrade informations are specified in https://github.com/atomone-hub/networks/blob/master/atomone-testnet-1/upgrades/v3.0.1.md

Merge only after block https://testnet.explorer.allinbits.services/atomone-testnet-1/block/3240000 passed


JeremyParish69:
The reverenced block has passed. Is the upgrade still valid and is this ready to be merged?

PR #6543, opened 2025-08-27
Revert "Create noble-saga.json" - Older channel existed.

ertemann:
Reverts cosmos/chain-registry#6533


ertemann:
hi @JeremyParish69 they ended up making another channel canonical.

PR #6547, opened 2025-08-28
Update assetlist.json

Alex703778:
Update 
Denom LABS
Display LABS

### DIFF
--- a/osmosis/assetlist.json
+++ b/osmosis/assetlist.json
@@ -27510,14 +27510,14 @@
           "exponent": 0
         },
         {
-          "denom": "labs",
+          "denom": "LABS",
           "exponent": 6
         }
       ],
       "type_asset": "sdk.coin",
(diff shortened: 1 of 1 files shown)


osmosis/assetlist.json
```diff
@@ -27510,14 +27510,14 @@
...
-          "denom": "labs",
+          "denom": "LABS",
           "exponent": 6
         }
       ],
       "type_asset": "sdk.coin",
       "base": "factory/osmo1are7fpe5l6jzm9sjn7u4qkq6q77wwrrsxzlyw8lcegmmaxdukvuq4h46dx/LABS",
       "name": "Labrador",
-      "display": "labs",
+      "display": "LABS",
```
JeremyParishTheBuilder, line 27520: note that changing this doesn't really do anything. Apps use Symbol and Name but the "display" property is only for deriving the preferred exponent representation

Alex703778: Thank you for answer.


This change in the token metadata affects, for example, the display of the token in the DAODAO zone app. Therefore, I also suggested doing the same here in the discord as in the token metadata, so that it would be the same everywhere.

PR #6554, opened 2025-08-30
Add Intento INTO token

danieljdd:
Add Intento INTO token to the asset lists of Osmosis and Cosmos Hub.

### DIFF
--- a/cosmoshub/assetlist.json
+++ b/cosmoshub/assetlist.json
@@ -914,8 +914,7 @@
           "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/intento/images/into.png",
           "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/intento/images/into.svg"
         }
-      ],
-      "coingecko_id": "intento"
+      ]
     }
   ]
 }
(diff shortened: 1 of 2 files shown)


_IBC/osmosis-intento.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "osmosis",
+    "client_id": "07-tendermint-3549",
+    "connection_id": "connection-10824"
+  },
+  "chain_2": {
+    "chain_name": "intento",
```
JeremyParishTheBuilder: intento must be chain 1 since it comes before alphabetically (same with the file name)

danieljdd: Ah alright, I see the intento-osmosis already exists in the master branch. So its just the asset lists that are to be added.

cosmoshub/assetlist.json
```diff
@@ -866,6 +866,56 @@
...
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "intento",
+            "base_denom": "uinto",
+            "channel_id": "channel-1"
+          },
+          "chain": {
+            "channel_id": "channel-106076",
```
JeremyParishTheBuilder: This is not the correct channel

osmosis/assetlist.json
```diff
@@ -27533,6 +27533,56 @@
...
+      "name": "Intento",
+      "display": "into",
+      "symbol": "INTO",
+      "traces": [
+        {
+          "type": "ibc",
+          "counterparty": {
+            "chain_name": "intento",
+            "base_denom": "uinto",
+            "channel_id": "channel-1"
```
JeremyParishTheBuilder: This is not the correct channel_id

PR #6559, opened 2025-09-02
Adding Bitrock EVM chain

niodamian:


.github/state/coingecko.json
```diff
@@ -281,6 +281,10 @@
           "chain_name": "arbitrum",
           "base_denom": "0xaf88d065e77c8cC2239327C5EDb3A432268e5831"
         },
+        {
```
JeremyParishTheBuilder: Please don't update the coingecko cache directly

_non-cosmos/bitrock/assetlist.json
```diff
@@ -0,0 +1,156 @@
...
+      ],
+      "coingecko_id": "wrapped-bitrock",
+      "images": [
+        {
+          "image_sync": {
+            "chain_name": "bitrock",
+            "base_denom": "wei"
+          },
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/bitrock/images/wbrock.png",
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/bitrock/images/wbrock.svg"
```
JeremyParishTheBuilder, line 81: Please remove the rock.svg and wbrock.svg uploads since they are somewhat unsightly conversions of PNG to SVG, and aren't really visually identical.

.github/state/coingecko.json
niodamian: should i remove it ?

_non-cosmos/bitrock/assetlist.json
JeremyParishTheBuilder: I see rock.svg was replaced with a higher quality version, which is good. However the URL for wbrock.svg should now be removed since wbrock.svg was deleted

.github/state/coingecko.json
JeremyParishTheBuilder: Yes, remove all that changes made to this file

_non-cosmos/bitrock/assetlist.json
```diff
@@ -0,0 +1,156 @@
...
+          "exponent": 18
+        }
+      ],
+      "base": "0x826628622a8b55f4af6ac2a1dd1273bf837e651a",
+      "name": "Rockswap Token",
+      "display": "rock",
+      "symbol": "ROCK",
+      "coingecko_id": "rockswap",
+      "logo_URIs": {
+        "png": "https://assets.coingecko.com/coins/images/32444/standard/rockswap.png?1696531597"
```
JeremyParishTheBuilder: This URI is not accepted. Should repeat the first image in the images array to specify a default

_non-cosmos/ethereum/assetlist.json
```diff
@@ -2460,6 +2460,29 @@
...
+          "exponent": 0
+        },
+        {
+          "denom": "brock",
+          "exponent": 18
+        }
+      ],
+      "base": "0xde67d97b8770dc98c746a3fc0093c538666eb493",
+      "name": "Bitrock",
+      "display": "BROCK ",
```
JeremyParishTheBuilder: this sting is not defined as a unit (should be "brock")

_non-cosmos/bitrock/assetlist.json
```diff
@@ -0,0 +1,157 @@
...
+        {
+          "type": "bridge",
+          "counterparty": {
+            "chain_name": "ethereum",
+            "base_denom": "0xde67d97b8770dc98c746a3fc0093c538666eb493"
+          },
+          "provider": "Bitrock Bridge"
+        }
+      ],
+      "coingecko_id": "bitrock",
```
JeremyParishTheBuilder, line 33: Bridged assets cannot share the same coingecko_id as an original asset. If both the original asset and the bridge are from the same team and part of the same project, then the bridged version can be considered an "additional-mintage", instead of a "bridge" asset. Changing the trace "type" to "additional-mintage", and the "provider" to "Bitrock" should allow this to pass while allowing them to share the same coingecko ID

PR #6565, opened 2025-09-04
Add Intento chain fee tokens

danieljdd:
This adds fee tokens to Intento chain


jasbanza:
> This adds fee tokens to Intento chain
duplicate fees key, please delete the older invalid one

PR #6570, opened 2025-09-05
Update assetlist.json on Terra2

charleskimbest:
add dATOM

### DIFF
--- a/terra2/assetlist.json
+++ b/terra2/assetlist.json
@@ -1738,6 +1738,53 @@
           "provider": "ERIS Protocol"
         }
       ]
+    },
+    {
+      "denom_units": [
+        {
+          "denom": "ibc/223FF539430381ADAB3A66AC4822E253C3F845E9841F17FEEC207B3AA9F8D915",
+          "exponent": 6,
+          "aliases": [
+            "dATOM"
+          ]
+        },
+        {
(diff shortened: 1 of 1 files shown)


terra2/assetlist.json
```diff
@@ -1738,6 +1738,53 @@
...
         }
       ]
+    },
+    {
+      "denom_units": [
+        {
+          "denom": "ibc/223FF539430381ADAB3A66AC4822E253C3F845E9841F17FEEC207B3AA9F8D915",
+          "exponent": 6,
+          "aliases": [
+            "dATOM"
```
JeremyParishTheBuilder, line 1748: this is a poor choice of alias for the base denom. It should be used as the exponent 6 denom

terra2/assetlist.json
```diff
@@ -1738,6 +1738,53 @@
           "provider": "ERIS Protocol"
         }
       ]
+    },
+    {
+      "denom_units": [
+        {
+          "denom": "ibc/223FF539430381ADAB3A66AC4822E253C3F845E9841F17FEEC207B3AA9F8D915",
+          "exponent": 6,
```
JeremyParishTheBuilder, line 1746: This is the base denom so it must be 0. Move the 6 to the other denom unit

PR #6576, opened 2025-09-08
fix: USDrise Symbol

Senna46:
Correct the incorrect token symbol


sunrise/assetlist.json
```diff
@@ -93,7 +93,7 @@
       "base": "uusdrise",
       "name": "Sunrise USDrise",
       "display": "usdrise",
-      "symbol": "USDRISE",
+      "symbol": "USDrise",
```
JeremyParishTheBuilder, line 96: For now I'm inclined to accept, just because the exact requirements for Symbol are ill defined, but it may need to be reverted if we soon decide to enforce capital letters for symbols on assets without a relationship to its origin defined (there is currently no trace to USD)

PR #6584, opened 2025-09-09
Fix IBC channel metadata for Bitsong ↔ Axelar

Validatrium:


JeremyParish69:
the filename and chain_1 chain_2 must follow alphabetical order

PR #6586, opened 2025-09-09
Fixing Terra

niodamian:
update chain infos as well version

### DIFF
--- a/terra/versions.json
+++ b/terra/versions.json
@@ -78,14 +78,13 @@
     },
     {
       "name": "3.5.0",
-      "tag": "v3.5.0",
-      "height": 13215800,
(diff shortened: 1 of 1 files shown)


terra/versions.json
```diff
@@ -75,6 +75,43 @@
...
+    {
+      "name": "3.5.0",
+      "tag": "v3.5.0",
+      "height": 13215800,
+      "recommended_version": "v3.5.0",
+      "compatible_versions": [
+        "v3.5.0"
+      ],
+      "proposal": 12189,
+      "height": 24660000,
```
JeremyParishTheBuilder: you have height twice

PR #6587, opened 2025-09-09
feat: update version and apis

dblackstone1:


symphony/chain.json
```diff
@@ -32,8 +32,17 @@
   },
   "codebase": {
     "git_repo": "https://github.com/Orchestra-Labs/symphony",
-    "recommended_version": "v1.0.0",
-    "compatible_versions": ["v1.0.0"],
+    "recommended_version": "v1.0.7",
+    "compatible_versions": [
```
JeremyParishTheBuilder: Version updates must also be reflected in the verions.json file

dblackstone1: i've updated versions.json to include 1.0.7.  i've also removed from compatible versions so only the included tags for full releases (the ones listed) are included

symphony/versions.json
```diff
@@ -79,6 +73,30 @@
...
+      },
+      "sdk": {
+        "type": "cosmos",
+        "version": "0.50.9"
+      },
+      "cosmwasm": {
+        "version": "0.53.0",
+        "enabled": true
+      },
+      "previous_version_name": "v29",
```
JeremyParishTheBuilder, line 98: this previous version name doesn't look right...

symphony/chain.json
```diff
@@ -32,8 +32,8 @@
   },
   "codebase": {
     "git_repo": "https://github.com/Orchestra-Labs/symphony",
-    "recommended_version": "v1.0.0",
-    "compatible_versions": ["v1.0.0"],
+    "recommended_version": "v1.0.7",
+    "compatible_versions": ["v1.0.0", "v1.0.4", "v1.0.6", "v1.0.7"],
```
JeremyParishTheBuilder: The compatible versions array doesn't match the versions file. If all these versions are compatible, then the separate version objects should be combined into a single version object.

symphony/versions.json
dblackstone1: devs did this whole thing of following on from osmosis versions and it took a few tries before i finally got them to follow OUR version structure instead of theirs

symphony/chain.json
dblackstone1: i'll just remove those others.  i don't care about keeping them around so much

dblackstone1: removed.  it's now only 1.0.7

symphony/versions.json
JeremyParishTheBuilder: oh, it turns out it actually is defined for the previous version's name

symphony/versions.json
```diff
@@ -79,6 +73,30 @@
...
+      "tag": "v1.0.7",
+      "height": 3333145,
+      "recommended_version": "v1.0.7",
+      "compatible_versions": ["v1.0.7"],
+      "binaries": {
+        "linux/amd64": "https://github.com/Orchestra-Labs/symphony/releases/download/v1.0.7/symphonyd"
+      },
+      "consensus": {
+        "type": "cometbft",
+        "version": "v0.38.12"
```
JeremyParishTheBuilder, line 88: Path: consensus.version, Reason: Value mismatch, In "codebase": 0.38.11, In currentVersion: v0.38.12

symphony/versions.json
```diff
@@ -79,6 +73,30 @@
       },
       "previous_version_name": "v28",
       "next_version_name": ""
+    },
+    {
+      "name": "v1.0.7",
+      "tag": "v1.0.7",
+      "height": 3333145,
```
JeremyParishTheBuilder, line 80: This don't look to be the right block height. It's already used for previous versions.
(we'll be adding validation that the same block height isn't used on multiple version objects)

PR #6594, opened 2025-09-12
Update Symphony versions.json

nman98:


symphony/versions.json
```diff
@@ -78,6 +78,32 @@
...
+      },
+      "sdk": {
+        "type": "cosmos",
+        "version": "0.50.9"
+      },
+      "cosmwasm": {
+        "version": "0.53.0",
+        "enabled": true
+      },
+      "previous_version_name": "v29",
```
JeremyParishTheBuilder, line 106: This version name doesn't look right

nman98: Problem with the naming started mostly because when they made the chain they forked of osmosis. And while I am not sure about why but they kept the same upgrade names as some of the osmosis forks.
Example for the [first](https://ping.pub/symphony/gov/1) and [second](https://ping.pub/symphony/gov/3) where these upgrades have name v28 and v29 while the tag for the binaries were tagged from version 1.0 and are incremented from there.

The latest upgrade has the upgrade name of [v1.0.7](https://ping.pub/symphony/gov/4)

To avoid any sort of confusion I discussed with the team that this could cause to some confusion in the future and that the upgrade names should follow the naming of the tags. So hopefully from now on if the tag for the upgrade is marked for example 2.0.0 it should have the matching upgrade name instead of something like v30 even though the code used was from osmosis or any other chain.

PR #6605, opened 2025-09-17
Updated Neutron logo

luisqa-neutron:
Updated logos for NTRN and Neutron.


JeremyParish69:
There is apparent duplication of the (chain) logo (neutron and neutron-raw) are now visually identical, so we should really only have one set. Please delete one set and remove and image URL references the refer to the set being deleted.

jasbanza (approved):

luisqa-neutron:
Removed the images and added some more information regarding the maxbtc token.

neutron/assetlist.json
```diff
@@ -2066,8 +2066,39 @@
       "images": [
         {
           "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/eho.svg"
+        },
```
JeremyParishTheBuilder: the brackets aren't being used right here

JeremyParish69:
Please ensure that any images added or removed also have their references also adjusted. (neutron-raw is still being referenced in the chain.json, but doesn't exist anymore). Also, the latest set of images (ntrn) was uploaded to the root directory instead of within the /neutron/images/ directory; please relocate those.

neutron/assetlist.json
```diff
@@ -2142,12 +2143,23 @@
       "display": "maxBTC",
       "symbol": "maxBTC",
       "logo_URIs": {
-        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/maxBTC.svg"
```
JeremyParishTheBuilder, line 2165: The overall change to the rounded logo is good, but the old image (maxBTC (with the uppercase BTC)) will need to be removed, because we cannot have unused images.

neutron/assetlist.json
```diff
@@ -2142,12 +2143,23 @@
...
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/maxbtc.png",
+          "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/maxbtc.svg"
+        }
+      ],
+      "socials": {
+        "website": "https://structured.money/",
+        "twitter": "https://x.com/StructuredFi"
+        "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/maxbtc.svg"
```
JeremyParishTheBuilder: cannot have svg under 'socials'

neutron/assetlist.json
```diff
@@ -2142,12 +2143,23 @@
...
       },
       "coingecko_id": "maxbtc",
       "images": [
```
JeremyParishTheBuilder, line 2168: Have the `images` property twice?

JeremyParish69:
The schema is broken is its current state. Looks like there are duplicate object keys (socials, images) and a missing comma (really, it's just that svg is used where it shouldn't be)

PR #6606, opened 2025-09-17
chore: add new channel details between chains cosmoshub and sentinel

ironman0x7b2:
The Sentinel team has created a new channel between Cosmos Hub and Sentinel for IBC ERC-20 token transfers. Relayers will be running on both channel pairs.

### DIFF
--- a/_IBC/cosmoshub-sentinel.json
+++ b/_IBC/cosmoshub-sentinel.json
@@ -23,7 +23,7 @@
       "ordering": "unordered",
       "version": "ics20-1",
       "tags": {
-        "preferred": false
+        "preferred": false,
         "status": "ACTIVE"
       }
     },


JeremyParish69:
This would be the first instance of a mainnet ibc connection registering an IBC channel with transfer on both ports. I believe some earlier assumptions were that there would only every be one trasnfer/transfer IBC channel per chain pair. Why not use the existing transfer/transfer IBC channel?

JeremyParish69:
> @JeremyParish69 An initial version of the IBC ERC20 smart contract exists for current channels, but it has fixed metadata and doesn’t allow assigning a new contract. To address this, we deployed a new contract with customized metadata and an `IMintableAndBurnable` implementation. This required setting up a new channel between Sentinel and Cosmos Hub.
> 
> Ref: https://docs.skip.build/go/eureka/custom-erc20-integration

Ok thanks for explaining, I think I understand. I'm trying to double check with the Skip team, and also need to figure out how we're going to play this at the Registry since it's a first occurrence that was never supposed to be allowed.

JeremyParish69:
Btw, after talking with Skip, I believe this addition to be valid. This PR hasn't been forgotten--just working out the final details on how to handle it, but should be resolve soon, I believe.

_IBC/cosmoshub-sentinel.json
```diff
@@ -22,7 +22,25 @@
...
+        "port_id": "transfer"
+      },
+      "chain_2": {
+        "channel_id": "channel-97",
+        "port_id": "transfer"
+      },
+      "ordering": "unordered",
+      "version": "ics20-1",
+      "tags": {
+        "preferred": true,
```
JeremyParishTheBuilder, line 42: To allow the chain registry to keep the old channel registered, while also ensuring there is one preferred default channel, we require "preferred": true on the new 'canonical' channel.
Would you say this looks acceptable as I've edited it?

ironman0x7b2: Yes. Thank you!

PR #6610, opened 2025-09-17
Add Union mainnet

PoisonPhang:
Creates entry for Union's mainnet featuring chain, versions, and assets

### DIFF
--- a/union/chain.json
+++ b/union/chain.json
@@ -43,7 +43,7 @@
       "linux/arm64": "https://github.com/unionlabs/union/releases/download/uniond%2Fv1.2.2/uniond-release-aarch64-linux"
     },
     "consensus": {
-      "type": "cometbls",
+      "type": "cometbft",
       "version": "v1.0.1",
       "repo": "https://github.com/unionlabs/cometbls",
       "tag": "v1.0.1-cometbls"
--- a/union/versions.json
+++ b/union/versions.json
@@ -13,7 +13,7 @@
         "linux/arm64": "https://github.com/unionlabs/union/releases/download/uniond%2Fv1.0.0/uniond-release-aarch64-linux"
       },
       "consensus": {
-        "type": "cometbls",
+        "type": "cometbft",
         "version": "v1.0.1",
         "repo": "https://github.com/unionlabs/cometbls",
         "tag": "v1.0.1-cometbls"
@@ -38,7 +38,7 @@
         "linux/arm64": "https://github.com/unionlabs/union/releases/download/uniond%2Fv1.1.1/uniond-release-aarch64-linux"
       },
       "consensus":{
-        "type": "cometbls",
+        "type": "cometbft",
         "version": "v1.0.1",
         "repo": "https://github.com/unionlabs/cometbls",
         "tag": "v1.0.1-cometbls"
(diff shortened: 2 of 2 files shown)


benluelo (approved):

union/versions.json
```diff
@@ -0,0 +1,80 @@
+{
+  "$schema": "../versions.schema.json",
+  "chain_name": "union",
+  "versions": [
+    {
+      "name": "v1.0.0",
+      "recommended_version": "uniond/v1.0.0",
```
JeremyParishTheBuilder: uniond not part of semver (may be used in tag property, but not in the version property)

PoisonPhang: b0b77991f9f23c5aa232b28ed5d759dddf0b9606

union/chain.json
```diff
@@ -0,0 +1,175 @@
...
+    "compatible_versions": [
+      "v1.2.2"
+    ],
+    "binaries": {
+      "linux/amd64": "https://github.com/unionlabs/union/releases/download/uniond%2Fv1.2.2/uniond-release-x86_64-linux",
+      "linux/arm64": "https://github.com/unionlabs/union/releases/download/uniond%2Fv1.2.2/uniond-release-aarch64-linux"
+    },
+    "consensus": {
+      "type": "cometbls",
+      "version": "v1.0.1-cometbls",
```
JeremyParishTheBuilder: any version property cannot have anything other than standard semver (and the prepended 'v'), so the '-cometbls' cannot stay here

PoisonPhang: Thanks
b1f4e7f04d245eb2a073ba1a1092087e1d1695f1

union/versions.json
```diff
@@ -0,0 +1,80 @@
...
+      "recommended_version": "v1.0.0",
+      "compatible_versions": [
+        "v1.0.0"
+      ],
+      "binaries": {
+        "linux/amd64": "https://github.com/unionlabs/union/releases/download/uniond%2Fv1.0.0/uniond-release-x86_64-linux",
+        "linux/arm64": "https://github.com/unionlabs/union/releases/download/uniond%2Fv1.0.0/uniond-release-aarch64-linux"
+      },
+      "consensus": {
+        "type": "cometbls",
```
JeremyParishTheBuilder: This doesn't follow the schema. I don't think this type is allowed

PoisonPhang: e368bf7476518adce264c029648a19bcee37a3a6

benluelo: @JeremyParish69 is it possible to modify the schema? our consensus is not vanilla cometbft, but rather our fork [cometbls](https://docs.union.build/architecture/cometbls/).

benluelo: sei, for example, is listed as [sei-tendermint](https://github.com/cosmos/chain-registry/blob/master/sei/chain.json#L42), not tendermint.

cor: Yeah it's much better to list the correct consensus type, such that parties won't make incorrect assumptions about Union's consensus mechanism

JeremyParishTheBuilder: If it's a significant difference, then I think that makes sense. It can be added to the types in the schema files (chain and versions)

PoisonPhang: https://github.com/cosmos/chain-registry/pull/6627

PR #6612, opened 2025-09-17
Add LuncToken (LTK) logo

lunctoken:
- Added lunctoken.png logo for LuncToken (LTK)
- Location: terra/images/lunctoken.png
- Used for assetlist reference in Terra Classic (CW20 contract: terra1mm8tdp40r2slzwqxk8jsz66ayc4zp69muxeateq37x2xquttzsaqy7275a)


JeremyParish69:
If uploading an image, then please register the asset that it belongs to. Later, we'll be setting up automatic removal of unreferenced images (which would just delete this if LTK is not registered)

PR #6631, opened 2025-09-25
Update chain.json

raikovanton:
add gonz_chains


osmosis/chain.json
```diff
@@ -696,6 +696,10 @@
   },
   "apis": {
     "rpc": [
+      {
```
JeremyParishTheBuilder, line 699: Please add these to the bottom of their respective lists

PR #6637, opened 2025-10-02
Update MedasDigital 2.0 Infos

oxygene76:
Update MedasDigital 2.0 Infos


JeremyParish69:
A new chain id and new IBC connection means that this is a new chain, so it'll need a new directory and everything. Is medas-1 being discontinued?

oxygene76:
There was a hardfork from medasdigital-1 to medasdigital-2 in November 2024. This is the new Live Chain which is also live on osmosis and working..

Do I have to make a new directory?

JeremyParish69:
> There was a hardfork from medasdigital-1 to medasdigital-2 in November 2024. This is the new Live Chain which is also live on osmosis and working..
> 
> Do I have to make a new directory?

Yes, please.
medasdigital-1 can have all instances of chain_name value (in all files: chain, assetlist, versions, and all existing ibc files) replaced with `medasdigital1` (including the directory name and filenames(for ibc files). 
Then, medasdigital-2 can use a new directory called 'medasdigital', and use medasdigital as the chain_name is all places.

In practice, this looks like duplicating the existing directory, choosing one to represent the old (pre-fork) chain, and replacing the chain_name to the archived chain_name ('medasdigital1'), and all existing IBC files, too. Then the other directory just needs to chain-id updated to "medasdigital-2", and the IBC connection will be registered in a new IBC file.

PR #6641, opened 2025-10-05
Updated Medas Digital 2.0 Descriptions and socials

oxygene76:
@JeremyParish69 I saw you already merged the chain infos for medasdigital-2 last November.

So, could we let it be as it is?

I only updated some descriptions about the project and socials in the files.

Thank you


oxygene76:
Ok, thanks Jeremy. I deleted the file and will create correct one for next pull request..

PR #6648, opened 2025-10-08
feat(zigchaintestnet): add missing fields to chain.json

AndWa:


testnets/zigchaintestnet/chain.json
```diff
@@ -52,8 +60,76 @@
         "address": " https://public-zigchain-testnet-rpc.numia.xyz/",
```
CryptoAssassin1: There's a leading space before the URL! please remove it

AndWa: Thank you for noticing this. 🙏

It is fixed.

PR #6681, opened 2025-10-17
add ibc connections for paxi (osmosis, cosmoshub)

paxi-web3:
add ibc connections for paxi (osmosis, cosmoshub)

- Added _IBC/paxi-osmosis.json
  - channel-0 <-> channel-107354
  - status: live, preferred: true
- Added _IBC/paxi-cosmoshub.json
  - channel-1 <-> channel-1566
  - status: live
- Both verified via Hermes query and Ping.pub explorer

### DIFF
--- a/_IBC/cosmoshub-paxi.json
+++ b/_IBC/cosmoshub-paxi.json
@@ -23,8 +23,8 @@
       "ordering": "unordered",
       "version": "ics20-1",
       "tags": {
-        "status": "live",
-        "preferred": true
+        "preferred": true,
+        "status": "ACTIVE"
       }
     }
   ]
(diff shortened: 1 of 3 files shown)


_IBC/paxi-cosmoshub.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "paxi",
```
JeremyParishTheBuilder, line 9: p comes after c, so paxi must be chain2 and must come last in the file name. (it how we avoid multiple files describing the same pair). Same issue with the osmosis Ibc file

_IBC/cosmoshub-paxi.json
```diff
@@ -0,0 +1,31 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "osmosis",
```
JeremyParishTheBuilder: you have osmosis here in the cosmoshub file

PR #6690, opened 2025-10-20
update fandomchain

Cyrano71:
Compliance with slip-0173 (the bech32 prefix must be unique among all registered prefixes)
cf https://github.com/satoshilabs/slips/blob/master/slip-0173.md


_IBC/fandomchain-osmosis.json
```diff
@@ -17,7 +17,7 @@
         "port_id": "transfer"
       },
       "chain_2": {
-        "channel_id": "channel-107314",
+        "channel_id": "channel-107392",
```
JeremyParishTheBuilder, line 22: Why is the channel-id being changed? Generally, once registered, it cannot be changed.

Cyrano71: Because of the change of the bech32 prefix to be compliant with slip-0173.
I had to create a new channel

JeremyParishTheBuilder: Forgive my ignorance. But why does a new bech32 prefix necessitate a new IBC channel? I'm not aware of bech32 prefix having anything do to with ibc channel...

JeremyParishTheBuilder: nvm, I think it makes sense. Changing bech32 may require a new genesis

PR #6697, opened 2025-10-22
Add XAUt.atom on Terra2

charleskimbest:


cosmoshub/assetlist.json
```diff
@@ -1067,6 +1067,59 @@
...
+      "images": [
+        {
+          "image_sync": {
+            "chain_name": "ethereum",
+            "base_denom": "0x68749665ff8d2d112fa859aa293f07a622782f38"
+          },
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/_non-cosmos/ethereum/images/xaut.png"
+        }
+      ],
+      "coingecko-id": "tether-gold"
```
JeremyParishTheBuilder: Bridged assets cannot use the same coingecko id. If there is a Eureka-specific coingecko ID, that may be used; otherwise, please remove

charleskimbest: Thanks. I removed it.

PR #6708, opened 2025-10-25
add SEED token

MedvedCrypto:


neutron/assetlist.json
```diff
@@ -1360,6 +1360,35 @@
...
+      "base": "factory/neutron133xakkrfksq39wxy575unve2nyehg5npx75nph/seed",
+      "name": "SEED",
+      "display": "SEED",
+      "symbol": "SEED",
+      "logo_URIs": {
+        "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/seed.png"
+      },
+      "images": [
+        {
+          "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/neutron/images/seed.png"
```
JeremyParishTheBuilder: the image needs to be uploaded

MedvedCrypto: I think it`s done now:

https://github.com/chainapsis/keplr-chain-registry/pull/1314

JeremyParishTheBuilder: That's a different repo. The image needs to be uploaded to _this_ repo--at the URL mentioned in the "png" field (currently would be a broken link)

MedvedCrypto: Oh, please excuse me. Does it look normal now?

JeremyParishTheBuilder: Not quite. In addition to the URI needing to be valid, the image also has to be uploaded to this repo. If we were to allow referencing another repo, if that repo changes, the URI could break in the future

PR #6720, opened 2025-10-29
Add Mandragora's Bitway seed, peer, RPC, API and gRPC

McDaan:
We have also changed the compatible version from `v2.0.0` to `v2.0.1` as there has been an upgrade recently. For further reference: https://mainnet.itrocket.net/bitway/gov/4

### DIFF
--- a/bitway/chain.json
+++ b/bitway/chain.json
@@ -185,7 +185,7 @@
       {
         "address": "https://bitway.rpc.mandragora.io",
         "provider": "Mandragora"
-      
+      }
     ],
     "rest": [
       {


bitway/chain.json
```diff
@@ -171,7 +181,11 @@
       {
         "address": "https://bitway.rpc.m.stavr.tech",
         "provider": "🔥STAVR🔥"
-      }
+      },
+      {
+        "address": "https://bitway.rpc.mandragora.io",
+        "provider": "Mandragora"
+
```
JeremyParishTheBuilder: missing object-closing bracket }

bitway/chain.json
```diff
@@ -41,9 +41,9 @@
   },
   "codebase": {
     "git_repo": "https://github.com/bitwaylabs/bitway",
-    "recommended_version": "v2.0.0",
+    "recommended_version": "v2.0.1",
     "compatible_versions": [
-      "v2.0.0"
+      "v2.0.1"
```
JeremyParishTheBuilder, line 46: is v2.0.0 compatible with v2.0.1? I'm wondering if it should remain in the array...

bitway/chain.json
McDaan: Added, thanks!

bitway/chain.json
McDaan: It doesn't, it was a hardfork upgrade :)

PR #6752, opened 2025-11-10
  fix(gonka): correct peer IDs and update 6block endpoint paths

kotelnikova:
chain_name update gonka>> Gonka

### DIFF
--- a/gonka/chain.json
+++ b/gonka/chain.json
@@ -126,7 +126,7 @@
       },
       {
         "id": "0772f16cc65cb4d19341b192bd7eba964f11d124",
-        "address": "node2.gonka.ai:5000",
+        "address": "node1.gonka.ai:5000",
         "provider": "PS on Nebius"
       },
       {


gonka/chain.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "../chain.schema.json",
-  "chain_name": "gonka",
+  "chain_name": "Gonka",
```
CryptoAssassin1: The chain_name cannot be changed once it’s set. Users don’t see this property instead, they see the pretty_name.

gonka/chain.json
CryptoAssassin1: This is a bech32 address, not a peer ID please fix all the peer ids

kotelnikova:
Thanks @CryptoAssassin1 , updates added accordingly. PTAL

CryptoAssassin1:
> Thanks @CryptoAssassin1 , updates added accordingly. PTAL

 In both the `seeds` and `persistent_peers` sections, two entries now point to the same address `node2.gonka.ai:5000` with different node IDs:
  - ID `0772f16cc65cb4d19341b192bd7eba964f11d124` → `node2.gonka.ai:5000`
  - ID `645fbce2dedcc7166f4df7931d2c87ca5188b569` → `node2.gonka.ai:5000`

  This is incorrect you cannot have two different node identities on the same address:port combination.
  can you please explain why you are making this changes?

gonka/chain.json
```diff
@@ -120,17 +120,17 @@
...
       {
-        "id": "gonka1d7p03cu2y2yt3vytq9wlfm6tlz0lfhlgv9h82p",
+        "id": "39ebfea6d2ab91e90c26cb702345cfa2f9bc611b",
         "address": "gonka03.6block.com:8443",
         "provider": "6block"
       },
       {
         "id": "0772f16cc65cb4d19341b192bd7eba964f11d124",
-        "address": "node1.gonka.ai:5000",
+        "address": "node2.gonka.ai:5000",
```
CryptoAssassin1: Keep as node1

kotelnikova: Done. Thanks for catching this merge mistake.

PR #6754, opened 2025-11-12
Updating Neutron logo

luisqa-neutron:


CryptoAssassin1:
@luisqa-neutron Please make sure that any images added or removed have their references updated as well. Currently, testnets/neutrontestnet/chain.json still references deleted files.

PR #7425, opened 2025-11-19
Update `ibc` axelartestnet-osmosistestnet

clemensgg:
New channels requested by Osmosis team


CryptoAssassin1:
@clemensgg the old channels are still active. What is the reason for creating new ones?

clemensgg:
Hi @CryptoAssassin1 clients have been expired on both sides for a long time and it was found that for testing it's more feasible to create new clients. Let me check if it's possible to quickly recover the old ones though! (need to talk to both teams)

scutuatua-crypto (approved):

PR #7430, opened 2025-11-22
Removed fetchhub endpoints

jarnek73:


CryptoAssassin1:
@jarnek73 why you are removing the working endpoints?

jarnek73:
@CryptoAssassin1 Not active validator on the chain anymore, too much spamming comming from cosmos.directory and also leaving the chain soon.

PR #7432, opened 2025-11-26
chore(mantra): upgrade testnet v7.0.0-rc2

allthatjazzleo:


testnets/mantrachaintestnet2/chain.json
```diff
@@ -154,7 +153,7 @@
     {
       "image_sync": {
         "chain_name": "mantrachain",
-        "base_denom": "uom"
+        "base_denom": "amantra"
```
CryptoAssassin1: Image_sync will break keep "base_denom": "uom"

allthatjazzleo: @CryptoAssassin1 fixed

CryptoAssassin1: Thank you

PR #7440, opened 2025-12-04
Add moon-runners explorer

moon-runners:


injective/chain.json
```diff
@@ -402,12 +402,18 @@
       "url": "https://injective.valopers.com/",
       "tx_page": "https://injective.valopers.com/transactions/${txHash}",
       "account_page": "https://injective.valopers.com/account/${accountAddress}"
-    }
+    },
+    {
+      "kind": "moon-runners",
+      "url": "https://moon-runners.net/injective/mainnet",
```
CryptoAssassin1, line 408: Explorer URLs are not functional (all return 404).

moon-runners: Should be fixed by now..thanks

CryptoAssassin1: Thank you

PR #7445, opened 2025-12-06
Add Moon-Runnes explorer for multiples chains

moon-runners:


atomone/chain.json
```diff
@@ -632,10 +632,10 @@
       "account_page": "https://explorer.nyan-cat.net/atomone/account/${accountAddress}"
     },
     {
-      "kind": "BlockSync",
```
CryptoAssassin1, line 635: Why you are removing BlockSync explorer?

PR #7480, opened 2025-12-26
added ibc relayer connection between divinetestnet and osmosis testnet

kctoptal:


kctoptal:
Pls merge it so that we can test in testnet before rolling out for mainnet. Thanks.

PR #7498, opened 2026-01-13
Update GenesisL1

GenesisL1:
Update of info and images and script to get live peers in chain registry json format.

### DIFF
--- a/genesisl1/chain.json
+++ b/genesisl1/chain.json
@@ -48,7 +48,7 @@
     "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/genesisl1/images/l1coin.png",
     "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/genesisl1/images/l1coin.svg"
   },
-  "description": "",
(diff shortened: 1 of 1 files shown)


genesisl1/peers.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh
```
JeremyParishTheBuilder, line 1: please omit this file

genesisl1/chain.json
```diff
@@ -121,8 +174,8 @@
   ],
   "images": [
     {
-      "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/genesisl1/images/l1.png",
-      "svg": "https://raw.githubusercontent.com/cosmos/chain-registry/master/genesisl1/images/l1.svg"
+      "png": "https://raw.githubusercontent.com/cosmos/chain-registry/master/genesisl1/images/l1coin.png",
```
JeremyParishTheBuilder, line 177: changing the name of the image will have a larger impact. Any derived assets utilizing the same image will also need their refence updated to use the new name (e.g., genesisl1's assetlist, osmosis' assetlist, etc.)

PR #7517, opened 2026-01-29
Add moo chain and milkINIT on Initia

JohnnyWyles:
Adds milkINIT on Initia 
Adds moo chain for tracing purposes - there is no IBC connection here as it is a rollup
Adds milkINIT on moo for tracing

Moo is defined as an L2 on [Initia's registry ](https://github.com/initia-labs/initia-registry/tree/main/mainnets/moo)


moo/images/milkINIT.png
CryptoAssassin1, line 1: @JohnnyWyles move milkINIT image to moo images folder

PR #7522, opened 2026-02-01
add cosmoshub-dungeon-assetlist

evgen3000:


cosmoshub/assetlist.json
CryptoAssassin1, line 1: @evgen3000 The new entry uses tabs instead of spaces, replace all tab characters with 2 spaces in the new asset entry to match the existing file format.

evgen3000: done

PR #7585, opened 2026-03-17
updating bluechip chain

Bluechip23:


bluechip/chain.json
```diff
@@ -32,37 +29,51 @@
...
+        "address": "67.218.8.88:26659",
         "provider": "CryptoDungeon"
       }
     ]
   },
-  "images": [
+  "apis": {
+    "rpc": [
+      {
+        "address": "https://bluechip.rpc.bluechip.link",
```
CryptoAssassin1, line 58: The RPC/REST endpoints you've added still return chain_id: bluechip-2 and when you do, a chain_id change is treated as a hard fork , the old chain directory needs to be archived first. See the guide: https://github.com/cosmos/chain-registry/blob/master/_guides/chain.md

Bluechip23: It is treated as a hard fork even though it is an entirely new chain? I just want to make sure I am handling this correctly. 

This is an entirely new chain. no shared data. no history.

CryptoAssassin1: @Bluechip23  normally we would archive the old version (bluechip-2) into a separate directory and have the new version take its place here. However, since there are no IBC connections registered, no external asset traces on other chains, and no other registry integrations, we can proceed with an in-place update as before.
                                                                                                                                                                                      
Before this can merge, please fix the following:                                                                                                                                                                                                                                                                                                                          
  1. REST endpoint returning 502 — https://bluechip.api.bluechip.link needs to be live and healthy                                                                                     
  2. Genesis URL has a broken .git suffix                                                                                          
  3. images array was removed from both files — assetlist.json and chain.json must keep the images array alongside logo_URIs, not replace it
  4. versions.json needs updating — still references the old repo and v1, please update to reflect the new chain

Bluechip23: I believe I have updated everything correctly and the rest/RPC are now pointing at the right chain

CryptoAssassin1: 1. Genesis URL still broken — The .git suffix in codebase.genesis.genesis_url causes a 404:                                                                                        
 ❌ https://raw.githubusercontent.com/Bluechip23/bluechip-chain-official.git/main/genesis.json
 ✅ https://raw.githubusercontent.com/Bluechip23/bluechip-chain-official/main/genesis.json                                                                                          
 2. Version inconsistency — chain.json uses the commit hash main-28606ee7... but versions.json uses the tagged release v1.0.0. Since the binary is tagged at v1.0.0, please update chain.json codebase.recommended_version and compatible_versions to v1.0.0 to match.

CryptoAssassin1 (approved):

PR #7633, opened 2026-04-18
Update assetlist.json

arkliveventures-Arky:


arkliveventures-Arky:
Token is live on Coreum mainnet and transferable.

Happy to adjust metadata, image format, or structure if required.

PR #7704, opened 2026-05-29
Add files via upload

womp29792-hub:


CryptoAssassin1:
@womp29792-hub This is a Cosmostation chainlist submission (param_2.json/assets_2.json format), not a chain-registry one the images even point to cosmostation/chainlist, where this infra chain already exists (github.com/cosmostation/chainlist/tree/master/chain/infra). These files belong there, not here.

PR #7707, opened 2026-06-01
*cough cough* i think its right *cough cough*

womp29792-hub:


CryptoAssassin1:
@womp29792-hub   
1. Remove the root-level files: primary_colors.py, bight.png, chain.png look accidental and don't belong in the repo root. 
2. Network type: declared mainnet/live, but it's a single-validator node (moniker infra-pretestnet-vultr-01) and the genesis_url returns no valid genesis. Please submit as a testnet or justify mainnet status.
3. IBC placeholders: all 4 IBC files still have TODO_VERIFY_CURRENT_CLIENT_ID/CONNECTION_ID. Fill in real, on-chain-verified values.
4. IBC tags: use "status": "ACTIVE"; drop the custom source/preferred/infra_voucher_denom fields.
5. Images: commit them under infra/images/ with raw.githubusercontent.com URLs, not external infra.uquad.org links.

PR #7714, opened 2026-06-04
Update celestia <> injective IBC channel (reactivated)

santosdios:
The previous celestia <> injective channel (channel-7 / channel-152) 
expired in April 2025 after 400+ days without updates.

This PR updates the entry with the new active channel operated by Cumulo (cumulo.pro):
- channel-281 (celestia) <> channel-453 (injective-1)
- First active IBC channel between Celestia and Injective since April 2025

### DIFF
--- a/_IBC/celestia-cosmoshub.json
+++ b/_IBC/celestia-cosmoshub.json
@@ -1,0 +1,33 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "celestia",
+    "chain_id": "celestia",
+    "client_id": "07-tendermint-151",
+    "connection_id": "connection-103"
+  },
+  "chain_2": {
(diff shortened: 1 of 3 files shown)


CryptoAssassin1:
A new or replacement transfer/transfer channel is not allowed (unless governance on both chains approve it). A channel can be recovered through governance by recovering the client or replacing the client, but changing the channel id's is not allowed.
The existing channel is working fine so there's no expired client to recover and nothing to replace. The original channel is open, funded, and in active use.

santosdios:
Thanks for the clarification @CryptoAssassin1!

We checked and you're right — channel-7 / channel-152 is STATE_OPEN and 
the client is being actively updated. We misdiagnosed the issue on our end.

We'll operate as an additional relayer on the existing channel instead. 
Thanks for the quick review!

PR #7718, opened 2026-06-05
Add Snapshots for High Stakes

Jackson-DKMG:
Here's the rest of the snapshots we generate.

A potentially important detail : for Injective, we provide two snapshots (pebble and golevel). So I have the two entries in the file, as I assume it's the right way to do it -- but let me know if not.


Jackson-DKMG (commented):
sorry, messed up some of the json. Should be ok now

CryptoAssassin1 (approved):

PR #7722, opened 2026-06-08
update Hippo Protocol IBC path and denomination

0114kek:


CryptoAssassin1 (approved):

CryptoAssassin1:
@0114kek  the IBC file now points to the new channel, but osmosis/assetlist.json was left on the old channel, so the two files would contradict each other after merge. The Osmosis HP entry still uses ibc/955E5E…, which is sha256("transfer/channel-104931/ahp") — the old channel.Please update the Osmosis entry to the new channel

0114kek:
@CryptoAssassin1 Thanks for the correction. I updated it.

CryptoAssassin1 (approved):

PR #7734, opened 2026-06-16
Xion network rebrand to Verona

jburnt:
Xion Network is currently undergoing a rebrand to Verona. 
https://docs.burnt.com/verona/others/rebrand-from-xion

This PR updates only the visual aspect of the token since the base denomiation and underlying ibc connections will remaing (`uxion`) to avoid the need of a full Token Migration. 

Also related https://github.com/burnt-labs/xion/pull/574

Testnet unchanged


elys/assetlist.json
```diff
@@ -370,15 +371,19 @@
...
+            "VERONA"
+          ]
         }
       ],
       "type_asset": "ics20",
       "base": "ibc/6BFB09FE2464A7681645610F56BBEFF555A00B8AE146339FEB4609BF40FB0F4A",
-      "name": "Xion",
-      "display": "xion",
-      "symbol": "XION",
+      "name": "Verona",
```
CryptoAssassin1, line 384: Verified on-chain: uxion's bank denoms_metadata on xion-mainnet-1 still returns symbol: "XION", name: "xion", display: "XION" (denom_units: uxion, mxion, XION — no VERONA).

jburnt: @CryptoAssassin1 it will be added  in the upcoming upgrade, but the change on exchanges are happening tomorrow. Let me know if you think this would affect it somehow. The change is already live on Keplr as well.

CryptoAssassin1 (approved):

PR #7777, opened 2026-07-02
fix: remove dead Notional endpoints registry-wide (34 chains)

CryptoAssassin1:
Removes all dead **Notional** RPC/REST/gRPC endpoints across **34 chains**.

### Why
Notional's public node infrastructure is fully decommissioned — **both domains are NXDOMAIN** (verified 2026-07-02):
- `notional.ventures` → does not resolve
- `cosmosia.notional.ventures` → does not resolve

Every registered endpoint on `*.cosmosia.notional.ventures` / `*.notional.ventures` is therefore dead.

### Scope
- ~95 endpoint entries removed across 34 chains (rpc / rest / grpc).
- Where a chain's dead Notional entry was the sole element of an api array (e.g. `starname` gRPC), the empty array/key is removed.
- **`dig` is intentionally excluded**: its *only* endpoints were Notional, so removal would leave it with zero endpoints — it appears to be a dead chain and should be reviewed separately, not silently emptied here.

Notional had **no** `peers[]` entries, so this is a complete apis cleanup.

### DIFF
--- a/agoric/chain.json
+++ b/agoric/chain.json
@@ -163,10 +163,6 @@
         "provider": "StakeWithUs"
       },
       {
-        "address": "https://rpc-agoric-ia.cosmosia.notional.ventures",
-        "provider": "Notional"
-      },
-      {
         "address": "https://agoric-rpc.stakeandrelax.net",
         "provider": "Stake&Relax 🦥"
       },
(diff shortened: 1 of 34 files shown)


CryptoAssassin1:
Closing — this was based on a stale fork and came out with merge conflicts. Reopening a clean version once my fork is synced to current master. The underlying finding stands: Notional's infrastructure is fully decommissioned (both notional.ventures and cosmosia.notional.ventures are NXDOMAIN), so all its registered endpoints are dead.

PR #7782, opened 2026-07-03
Add QUAD public RPC/REST endpoints (Osmosis, Injective)

womp29792-hub:
Follow-up to #7762, which added QUAD public endpoints for Cosmos Hub, Akash, Celestia, and Stride.

This adds RPC + REST for two more chains we run public nodes for:

| Chain | Endpoint |
|---|---|
| Osmosis (`osmosis-1`) | https://osmosis.rpc.uquad.org:443 |
| Injective (`injective-1`) | https://injective.rpc.uquad.org:443 |

Both serve keyless, rate-limited RPC (`/status`, `/abci_info`) and REST (`/cosmos/base/tendermint/...`) and are currently synced to tip.

### DIFF
--- a/osmosis/chain.json
+++ b/osmosis/chain.json
@@ -769,6 +769,10 @@
       {
         "address": "https://osmosis.api.pocket.network",
         "provider": "Pocket Network"
+      },
+      {
+        "address": "https://osmosis.rpc.uquad.org:443",
+        "provider": "QUAD"
       }
     ],
     "rest": [
@@ -831,6 +835,10 @@
       {
         "address": "https://osmosis.api.pocket.network",
         "provider": "Pocket Network"
+      },
+      {
+        "address": "https://osmosis.rpc.uquad.org:443",
+        "provider": "QUAD"
       }
     ],
     "grpc": [
--- a/injective/chain.json
+++ b/injective/chain.json
@@ -190,6 +190,10 @@
       {
         "address": "https://public.stakewolle.com/cosmos/injective/rpc",
         "provider": "Stakewolle"
+      },
+      {
+        "address": "https://injective.rpc.uquad.org:443",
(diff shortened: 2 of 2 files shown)


osmosis/chain.json
```diff
@@ -831,6 +835,10 @@
       {
         "address": "https://osmosis.api.pocket.network",
         "provider": "Pocket Network"
+      },
+      {
+        "address": "https://osmosis.rpc.uquad.org:443",
```
CryptoAssassin1, line 840: @womp29792-hub  The rest entry here uses your RPC URL (osmosis.rpc.uquad.org / injective.rpc.uquad.org) same address as the rpc entry. If you run a REST/LCD (API) endpoint, please use its proper hostname (e.g. osmosis.api.uquad.org). If you don't have a separate API endpoint, please remove the rest entry and keep only rpc. Same applies to both Osmosis and Injective.

womp29792-hub: The shared hostname is deliberate — each of our gateways is a single path-routed host: Tendermint RPC at the root, LCD/REST on its standard `/cosmos/...` paths on the same host. Both REST entries are live and keyless right now:

```
curl https://osmosis.rpc.uquad.org/cosmos/base/tendermint/v1beta1/blocks/latest
curl https://injective.rpc.uquad.org/cosmos/base/tendermint/v1beta1/blocks/latest
```

This mirrors our entries merged in #7762 (cosmoshub/akash/celestia/stride each list the same host for `rpc` and `rest`) and those resolve fine on cosmos.directory. Happy to switch to dedicated `api.` hostnames if maintainers prefer, but the listed addresses are accurate and serving. Anything else needed to get this merged?

CryptoAssassin1: got it Thank you

PR #7786, opened 2026-07-04
Add QUAD public RPC/REST endpoint (dYdX)

womp29792-hub:
Follow-up to #7762 (merged) and #7782. Adds live public nodes we operate:

- `dydx.rpc.uquad.org` (dydx-mainnet-1)
- `neutron.rpc.uquad.org` (neutron-1)

Both serve keyless, rate-limited RPC + REST at current height. Each gateway is a single path-routed host: Tendermint RPC at the root, LCD/REST on its standard `/cosmos/...` paths on the same host (same pattern as our merged #7762 entries).

Verify:
```
curl https://dydx.rpc.uquad.org/status
curl https://dydx.rpc.uquad.org/cosmos/base/tendermint/v1beta1/blocks/latest
curl https://neutron.rpc.uquad.org/status
curl https://neutron.rpc.uquad.org/cosmos/base/tendermint/v1beta1/blocks/latest
```

### DIFF
--- a/dydx/chain.json
+++ b/dydx/chain.json
@@ -179,6 +179,10 @@
       {
         "address": "https://dydx-rpc.noders.services",
         "provider": "[NODERS]TEAM"
+      },
+      {
+        "address": "https://dydx.rpc.uquad.org:443",
+        "provider": "QUAD"
       }
     ],
     "rest": [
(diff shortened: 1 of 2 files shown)


womp29792-hub:
Updated: dropped the Neutron entry for now — our Neutron node is mid-resync and I don't want a stale endpoint listed. dYdX (dydx-mainnet-1) remains and is verified keyless RPC+REST at current height. I'll open a separate PR for Neutron once it's back at tip.

CryptoAssassin1 (approved):

PR #7858, opened 2026-07-28
ritbit: rename display name to Rubin, add codebase metadata

rprudnikov-itrf:
The network is publicly branded **Rubin** (https://rubin.trade). This PR aligns the display name and fills in previously missing metadata.

### Display name

- `pretty_name`: `RITBIT` → `Rubin`
- `description` reworded accordingly

`chain_name`, `chain_id` (`ritbit-mainnet`) and `bech32_prefix` (`rit`) are intentionally left unchanged — the chain id cannot be altered without a network restart, so the mismatch between the display name and the identifiers is expected and permanent.

### Added metadata

- `daemon_name`: `ritbitd`
- `node_home`: `$HOME/.ritbit`
- `key_algos`: `["secp256k1"]`
- `codebase`: `recommended_version` `v27.3`, plus consensus / sdk / ibc versions

Values are taken from the running mainnet (`/cosmos/base/tendermint/v1beta1/node_info` reports app `27.3`, Cosmos SDK `v0.54.3`) and from the node binary defaults.

`codebase.git_repo` is omitted because the source repository is not public.

### Endpoints

All endpoints already listed in the entry were verified while preparing this change:

- `https://rpc.mainnet.rubin.trade` — HTTP 200
- `https://rest.mainnet.rubin.trade` — HTTP 200
- `grpc.mainnet.rubin.trade:443` — responds to `cosmos.base.tendermint.v1beta1.Service/GetNodeInfo`

### DIFF
--- a/ritbit/chain.json
+++ b/ritbit/chain.json
@@ -4,10 +4,15 @@
   "status": "live",
   "website": "https://rubin.trade/",
   "network_type": "mainnet",
-  "pretty_name": "RITBIT",
+  "pretty_name": "Rubin",
   "chain_type": "cosmos",
   "chain_id": "ritbit-mainnet",
   "bech32_prefix": "rit",
+  "daemon_name": "ritbitd",
+  "node_home": "$HOME/.ritbit",
+  "key_algos": [
+    "secp256k1"
+  ],
   "slip44": 118,
   "fees": {
     "fee_tokens": [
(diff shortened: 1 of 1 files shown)


rprudnikov-itrf:
Pushed one more correction: the `urit` gas prices in this entry were **half** the value the network actually enforces.

The node reports:

```
$ curl -s https://rest.mainnet.rubin.trade/cosmos/base/node/v1beta1/config | jq -r .minimum_gas_price
25000000000.000000000000000000urit,0.025000000000000000uusdc
```

The entry had `fixed_min_gas_price: 12500000000`, so a wallet or relayer using the registry value would have its transactions rejected by every node on the network. Updated to `25000000000`, with `high_gas_price` scaled to keep the previous 1.6x ratio. The `uusdc` token already matched and is unchanged.

PR #7865, opened 2026-08-03
Update Unification upgrades and binary versions

Codegnosis:
Adds the two und releases missing from the registry for both Unification networks.

- 7-taryon (v1.12.0) — shipped Aug/Sep 2025 and was never filed upstream; the registry currently stops at 6-scanlan/v1.11.0 while both chains have run v1.12.0 since.
- 8-vaxildan (v1.13.0) — TestNet upgrade scheduled at height 25638900 (~11 Aug 2026), governance proposal 50, currently in its voting period. The MainNet entry will follow once that window is scheduled.

binaries now lists all six goreleaser targets (linux/darwin/windows × amd64/arm64); previous entries listed amd64 only

### DIFF
--- a/testnets/unificationtestnet/chain.json
+++ b/testnets/unificationtestnet/chain.json
@@ -37,34 +37,37 @@
   },
   "codebase": {
     "git_repo": "https://github.com/unification-com/mainchain",
-    "recommended_version": "v1.11.0",
+    "recommended_version": "v1.12.0",
     "compatible_versions": [
-      "v1.11.0"
+      "v1.12.0"
     ],
     "binaries": {
-      "linux/amd64": "https://github.com/unification-com/mainchain/releases/download/v1.11.0/und_v1.11.0_linux_x86_64.tar.gz",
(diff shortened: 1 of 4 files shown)


CryptoAssassin1:
@Codegnosis Unification `chain.json` `codebase` is still v1.11.0 on both chains so please bump it to v1.12.0 (tag/recommended_version/compatible_versions/sdk 0.53.4/cometbft 0.38.17/ibc 10.3.0/6 binaries)

Codegnosis:
> @Codegnosis Unification `chain.json` `codebase` is still v1.11.0 on both chains so please bump it to v1.12.0 (tag/recommended_version/compatible_versions/sdk 0.53.4/cometbft 0.38.17/ibc 10.3.0/6 binaries)

Thanks - good catch. I've also bumped the versions for all dependencies (IBC etc.) which were also quite a way behind.

CryptoAssassin1 (approved):

PR #7885, opened 2026-08-13
fix(cosmoshub): remove 6 peer entries whose hostnames no longer resolve

durenodes:
`certusone`, `WhisperNode`, `Stakeflow` and `MMS` each still have a seed or persistent peer entry in `cosmoshub/chain.json` pointing at a hostname that no longer resolves. A node bootstrapping from this registry spends dial attempts on addresses that cannot resolve, and starts with fewer usable seeds than the file implies.

## Removed — NXDOMAIN on two independent resolvers

Google DoH and Cloudflare DoH both return `Status: 3` with no answer section:

| Section | Address | Provider |
| --- | --- | --- |
| `seeds` | `public-seed-node.cosmoshub.certus.one:26656` | certusone |
| `seeds` | `seeds.whispernode.com:14956` | WhisperNode 🤐 |
| `seeds` | `seed-cosmos-hub-01.stakeflow.io:26656` | Stakeflow |
| `seeds` | `seed-node.mms.team:43656` | MMS |
| `persistent_peers` | `peer-cosmos-hub-01.stakeflow.io:26656` | Stakeflow |
| `persistent_peers` | `peer-cosmoshub.mms.team:26656` | MMS |

Controls from the same file, queried from the same place at the same moment, resolved normally:

```
seed.publicnode.com                  Status 0, A record present
cosmoshub-mainnet-seed.itrocket.net  Status 0, A record present
```

So this is not a resolver or network problem on my side.

## Deliberately left alone

Several other entries refuse TCP or time out from my location. I did not touch them. A refused port can be a moved service, and a timeout can be firewalling, geographic filtering, or a listener that has hit `max_incoming_connections` — none of which proves the entry is dead. Only DNS resolution is unambiguous from a single vantage point, so that is the whole scope here.

## Notes

- These six hostnames appear nowhere else in the registry, so this is a single-file change.
- Stakeflow's irisnet entries were removed for the same reason in #7871. The same hostname scheme is still listed for the Hub.
- `apis` entries for these providers were already cleaned up in the June cosmoshub pass; only `peers` remained.
- Diff is deletions only. `chain-registry validate --registryDir .` passes, 2242 tests.

### DIFF
--- a/cosmoshub/chain.json
+++ b/cosmoshub/chain.json
@@ -73,11 +73,6 @@
   "peers": {
     "seeds": [
       {
-        "id": "ba3bacc714817218562f743178228f23678b2873",
-        "address": "public-seed-node.cosmoshub.certus.one:26656",
-        "provider": "certusone"
-      },
-      {
         "id": "ade4d8bc8cbe014af6ebdf3cb7b1e9ad36f412c0",
         "address": "seeds.polkachu.com:14956",
         "provider": "Polkachu"
@@ -93,11 +88,6 @@
         "provider": "Golden Ratio Staking"
       },
       {
-        "id": "c28827cb96c14c905b127b92065a3fb4cd77d7f6",
-        "address": "seeds.whispernode.com:14956",
-        "provider": "WhisperNode 🤐"
-      },
-      {
         "id": "8542cd7e6bf9d260fef543bc49e59be5a3fa9074",
         "address": "seed.publicnode.com:26656",
         "provider": "Allnodes ⚡️ Nodes & Staking"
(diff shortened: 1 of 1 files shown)


durenodes:
Friendly ping — the workflows on this PR have not run yet. As a first-time contributor I believe a maintainer has to approve the run, so there is nothing I can do from my side to get CI green.

The change is deletions only: 6 peer entries in `cosmoshub/chain.json` whose hostnames return NXDOMAIN from both Google and Cloudflare DoH. Two controls resolved normally at the same moment, so it is not a resolver issue on my end. `chain-registry validate` passes locally (2,242 tests).

Happy to rebase or split this if that would help.

CryptoAssassin1 (approved):

PR #7897, opened 2026-08-19
Add Injective USDC (USDC.inj) to 8 chain assetlists

hyukjoon-clk:
**## Summary**
Registers Injective-originated USDC (`USDC.inj`) as a new IBC asset on 8
chains: dYdX, Cosmos Hub, Babylon, Dymension, Osmosis, Terra 2.0, Neutron,
and ZIGChain.

The underlying asset is Noble USDC bridged onto Injective's native EVM
(`erc20:0xa00C59fF5a080D2b954d0c75e46E22a0c371235a`, symbol `USDC`,
decimals `6`, confirmed on-chain via `eth_call`). It reaches these chains
either directly from Injective (dYdX) or via a two-hop path through
Cosmos Hub (all other chains).

**## Verification**
For each chain, the `base` denom hash was independently recomputed from
`traces[].chain.path` and cross-checked against:
- the chain's live IBC channel state (`STATE_OPEN`, correct counterparty
  channel) via each chain's REST endpoint
- the `_IBC/*.json` connection records in this repo

### DIFF
--- a/_IBC/dydx-injective.json
+++ b/_IBC/dydx-injective.json
@@ -1,0 +1,33 @@
+{
+  "$schema": "../ibc_data.schema.json",
+  "chain_1": {
+    "chain_name": "dydx",
+    "chain_id": "dydx-mainnet-1",
+    "client_id": "07-tendermint-26",
+    "connection_id": "connection-30"
+  },
+  "chain_2": {
+    "chain_name": "injective",
+    "chain_id": "injective-1",
+    "client_id": "07-tendermint-256",
+    "connection_id": "connection-249"
+  },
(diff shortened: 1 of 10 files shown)


JohnnyWyles:
Only DYDX and Cosmos Hub have direct-to-Injective connections here. I confirmed with Rob last week that Cosmos Lab's plan was for Osmosis to use a direct-from-Injective route. Can you double-check the intention? Most chains here seem to have direct-to-Injective connections open.

JohnnyWyles (approved):

CryptoAssassin1 (approved):

PR #7942, opened 2026-09-05
epix: refresh chain metadata, versions, endpoints, explorers, peers; add coingecko_id; mark testnet killed

MudDev:
Brings the `epix` registry entries in line with the network as it runs today. Every value below was checked against live nodes.

## Mainnet `epix/chain.json`

- **codebase**: repository is now `EpixZone/EpixChain`; recommended and compatible version `v0.7.3`; genesis URL pinned to the `v0.5.0-rc42` tag, whose `app_state` matches what the public RPC nodes serve (the previous URL returned 404, and the `main` copy gained `bank.denom_metadata` after launch); SDK `v0.54.3`, CometBFT `v0.39.3`, ibc-go `v11.1.0`, Go `1.25.9`, cosmwasm disabled. The previous block described a `v1.0.0` on SDK 0.46 / Tendermint 0.34 / ibc-go v6 that never ran on this network.
- **fees**: minimum gas price corrected to 20 gwei, set by proposal 6 (the entry still listed 1000 gwei).
- **description** added.
- **apis**: removed two gRPC endpoints that do not answer (`grpc.epix.zone:15067`, `grpc-epix.dnsarz.net:443`). The four RPC, four REST, and the EVM JSON-RPC endpoints all respond and were left unchanged.
- **explorers**: the Epix explorer is a hash-router single-chain build, so its templates are now `/#/tx/`, `/#/account/`, `/#/block/`, `/#/gov/` rather than `/epix/tx/`. Added account and block pages for the OneNov and Blockscout explorers.
- **peers**: added the Silk Nodes persistent peer that is published in the EpixChain validator onboarding docs.

## Mainnet `epix/versions.json`

Replaces the single fictional `v1.0.0` entry with the governance upgrade history: v0.5.1 through v0.7.2 with their upgrade heights and proposal ids, plus v0.7.3 as the current rolling release. The v0.7.3 entry mirrors the chain.json codebase block.

## Testnet

`testnets/epixtestnet/` is kept as on master with `"status"` set to `killed`. The 2025 network that entry described ended with the March 2026 relaunch of `epix_1917-1`, and the relaunched testnet is not maintained as a public resource.

## Asset

Adds `coingecko_id: "epix"` to the EPIX asset. The id is present in the CoinGecko `/coins/list` response that the registry's CoinGecko check validates against.

All five files validate against their schemas, and the versions entries for the recommended releases match the chain.json codebase blocks per `validate_data.mjs`.


CryptoAssassin1:
@MudDev 1. The mainnet genesis on `main` differs from what the four public RPC nodes actually serve. The `main` copy had `bank.denom_metadata` added in commit 8710869a, after launch. The node genesis matches the file at tag `v0.5.0-rc42`.

2. The testnet genesis_url points at the 2025-09-02 genesis with 247 accounts, but the live `epix_1917-1` genesis (via `rpc.testnet.epix.zone/genesis`) is from 2026-03-10 with one account and a `localtestnet` gentx. The URL does not resolve to the genesis the testnet actually runs on.
3. The testnet has already applied the v0.7.2 upgrade at height 2752415 and the node runs `v0.7.1-27-ga411b9b6`, which includes the v0.7.2 handler. A v0.7.1 binary would not get past that height.

Also noticed the mainnet feemarket min gas price has been 20 gwei since proposal 6, while fee_tokens still list 1000 gwei.

MudDev:
Thanks for the careful check, all four are addressed in 57707f62:

1. **Mainnet genesis**: `genesis_url` now pins the `v0.5.0-rc42` tag. I confirmed the `app_state` of that file hashes identically to what `rpc.epix.zone/genesis` serves; the `main` copy differs only by the `bank.denom_metadata` added in 8710869a after launch.
2. and 3. **Testnet**: removed `testnets/epixtestnet/` entirely rather than patch it. The registered genesis and versions did not describe the network that is running, and the testnet is not maintained as a public resource, so an entry would need constant correction to stay honest.
4. **Fees**: updated to the 20 gwei minimum from proposal 6 (`fixed_min` and `low` 20 gwei, `average` 25, `high` 30; current `eth_gasPrice` is about 22.5 gwei).

`chain-registry validate` and `validate_data.mjs` both pass on the branch.

CryptoAssassin1:
@MudDev thanks for the fixes, the mainnet side looks good. One thing we can't take is removing testnets/epixtestnet. Once a chain is listed in the registry it stays listed. Networks that are no longer running are marked with "status": "killed" and keep their files.

Two ways forward your call:

1. Restore the three testnet files as they are on master and just flip "status" to "killed". The 2025 testnet that was listed ended with the March relaunch, so this is accurate.

2. Treat the relaunch as a new network under the same chain id, the way dungeon1/dungeon and xiontestnet1/xiontestnet2 are handled. Rename testnets/epixtestnet to testnets/epixtestnet1 with "status": "killed", then add a new testnets/epixtestnet for the current epix_1917-1 network: publish its genesis in EpixChain and point genesis_url at it, set recommended_version to v0.7.2 or v0.7.3, and list the upgrades it actually applied in versions.json (v0.5.1 to v0.5.5 at heights 27, 52, 77, 102, 127, then v0.7.0 at 1100300, v0.7.1 at 2654619, v0.7.2 at 2752415).

MudDev:
Understood, thanks. Went with option 1 in 8244c7ae: the three testnet files are restored exactly as on master with `"status": "killed"`. The relaunched testnet is not something we maintain as a public resource, so I would rather not list it as a new network for now. Both validators pass on the branch.

CryptoAssassin1 (approved):

PR #7949, opened 2026-09-10
terpnetwork: v6.0.1 recommended, add v6.1/v6.2 sequence

hard-nett:
## Summary

Update `terpnetwork` (`morocco-1`) from the terp-core source of truth (`networks/chain-registry/terpnetwork/` on [terpnetwork/terp-core](https://github.com/terpnetwork/terp-core)).

`chain.json` `recommended_version` is **v6.0.1** until the v6.1 halt. Do not run v6.1.0 / v6.2.0 before height 23191300.

| step | version | height | proposal | artifacts |
|------|---------|--------|----------|-----------|
| now (rolling CosmWasm halt patch) | **v6.0.1** | none | none | https://s3.terp.network/releases/terp-core/v6.0.1/ |
| gov plan `v6.1` | **v6.1.0** | **23191300** | **59** | https://s3.terp.network/releases/terp-core/v6.1.0/ |
| handler arms `v6.2` at apply+2 | **v6.2.0** | **23191302** | none (not a second gov plan) | https://s3.terp.network/releases/terp-core/v6.2.0/ |

Linux amd64/arm64 Cosmovisor tarballs include `?checksum=sha256:…` matching the published `sha256sum.txt` files.

Also:
- `node_home` is `$HOME/.terpd` (matches `DefaultNodeHomeDir` and the wasm path)
- ibc-go **v11.2.0**, SDK **0.54.3** / CometBFT **0.39.3** on v6.0.x; SDK **0.55.0** / CometBFT **0.40.0** on v6.1/v6.2
- CosmWasm **0.70.3** (schema cannot encode our `-zk` guest tag)

Supersedes stale draft https://github.com/cosmos/chain-registry/pull/7552 (`Terpnetwork@v5.0.4`).

## Files

- `terpnetwork/chain.json`
- `terpnetwork/versions.json`

PR #7965, opened 2026-09-14
terpnetwork: v6.2.0 recommended after v6.1/v6.2 upgrade (supersedes #7949)

CryptoAssassin1:
## Summary

Carries the two commits from #7949 by @hard-nett unchanged, plus one commit that catches `terpnetwork` (`morocco-1`) up with the upgrade that executed on 2026-09-13.

The v6.1 plan (proposal 59) applied at height 23191300 and the v6.1.0 handler armed `v6.2` at 23191302, so nodes now run **v6.2.0** (commit `0c24074e`, which is the v6.2.0 tag). #7949 could not be updated in place because the fork is org-owned and does not accept maintainer pushes.

### On top of #7949

`chain.json`
- `recommended_version` / `compatible_versions`: v6.0.1 to **v6.2.0**
- `consensus.version`: 0.39.3 to **0.40.0**
- `sdk.version`: 0.54.3 to **0.55.0**
- `binaries`: restored, pointing at the checksummed v6.2.0 S3 tarballs (linux/amd64, linux/arm64, darwin/arm64)

`versions.json`
- `binaries` added to the `v6.0.1`, `v6.1` and `v6.2` rows (S3 tarballs with `?checksum=sha256:` matching the published `sha256sum.txt`). No other rows changed.

### Verified
- `applied_plan/v6.1` = 23191300, `applied_plan/v6.2` = 23191302, proposal 59 PASSED
- `node_info`: version 6.2.0, git_commit `0c24074e` == tag v6.2.0, cometbft v0.40.0, sdk v0.55.0, ibc-go v11.2.0, wasmd v0.70.3
- all 8 new binary URLs return 200; the v6.2.0 linux-amd64 tarball was downloaded and its sha256 matches `sha256sum.txt`; the v6.1.0 checksums equal the proposal 59 plan checksums
- `chain-registry validate` and `validate_data.mjs` pass locally

Closes #7949.

### DIFF
--- a/terpnetwork/chain.json
+++ b/terpnetwork/chain.json
@@ -9,24 +9,36 @@
   "bech32_prefix": "terp",
   "slip44": 118,
   "daemon_name": "terp",
-  "node_home": "$HOME/.terp",
+  "node_home": "$HOME/.terpd",
   "codebase": {
     "git_repo": "https://github.com/terpnetwork/terp-core.git",
     "genesis": {
       "genesis_url": "https://raw.githubusercontent.com/terpnetwork/networks/main/mainnet/morocco-1/genesis.json"
     },
-    "recommended_version": "v5.0.3",
+    "recommended_version": "v6.0.1",
     "compatible_versions": [
-      "v5.0.2",
-      "v5.0.3"
+      "v6.0.1"
     ],
     "consensus": {
       "type": "cometbft",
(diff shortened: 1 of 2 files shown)


CryptoAssassin1:
Pushed the same commit directly onto #7949 instead, closing this duplicate.

PR #7968, opened 2026-09-14
Add NodeSafe endpoints for Epix

cryptogemfunds:
add:
API
RPC
EVMRPC (You need to enable this in app.toml under the [json-rpc] section)
GRPC
Top 100 turned on (under [topholders])


CryptoAssassin1:
@cryptogemfunds 1. Your merge commits reverted two other providers' entries. Please restore the master values:
   - Vinjan.Inc RPC must stay https://rpc-epix.vinjan-inc.com (the rpc-m host in your diff returns 502, and Vinjan moved off it in #7933).
   - The scan.epix.zone explorer kind must stay "Blockscout" (renamed in #7942), not "EVM".
   Easiest is to rebase on current master and re-apply only your NodeSafe additions.

2. grpc-epix.nodesafe-app.xyz:443 is not usable. Reflection works, but every unary call (GetNodeInfo, GetLatestBlock, GetSyncing, bank Params) fails with "server closed the stream without sending trailers" through Cloudflare. Please fix the gRPC proxying or drop the gRPC entry for now.

PR #7973, opened 2026-09-16
Add Artery Network testnet

marselzaripov1988:
Adds `testnets/arterytestnet` — the public testnet of Artery Network.

Artery is a Cosmos SDK chain with its own referral and delegation model: it has no `x/staking`, validators are handled by a custom `x/noding` module. The testnet runs SDK v0.53.8, CometBFT v0.38.26 and ibc-go v10.7.0.

**Endpoints** — behind a rate-limited reverse proxy, TLS from Let's Encrypt:

| | |
|---|---|
| RPC | https://rpc.artery-testnet.online |
| REST | https://api.artery-testnet.online |
| gRPC | grpc.artery-testnet.online:443 |

Verified against the checks this registry runs daily:

- `GET /status` → 200
- `GET /cosmos/base/tendermint/v1beta1/syncing` → 200, `{"syncing":false}`

CORS headers are set on both, so browser wallets can reach them.

**Notes on the entry**

- No `staking` section. Artery has no `x/staking` module: delegated ARTR (`uartrd`) sits on the delegator's own account rather than being bonded to a validator. Advertising staking tokens would point wallets at `/cosmos/staking/...`, which the chain does not serve.
- Three assets, because ARTR exists in three states on chain: liquid (`uartr`), delegated (`uartrd`) and undelegating (`uartrr`).
- `slip44` is 546, not 118.
- No explorer yet.

Ran `.github/workflows/utility/validate_data.mjs NO_API` locally before opening: output is byte-identical with and without this entry.


CryptoAssassin1:
@marselzaripov1988 `codebase`: tag `v2.5.8` 404s

